PR #1, opened 2017-03-20
update megadd, megaesp

ausHaus:


GermanBluefox (changes requested):
Let the published inside. It is to detect "new" adapters.

mcm1957:
This issue seems to be outdated

This issue suggests to update the stable version of this adapter to ${issue.to} but the current stable version is already ${stable[adapter].version}.

So this issue should be closed.

PR #3, opened 2017-04-27
Бондарев Роман

bondrogeen:
add iobroker repo


mcm1957:
This issue seems to be outdated

This issue suggests to update the stable version of this adapter to ${issue.to} but the current stable version is already ${stable[adapter].version}.

So this issue should be closed.

PR #12, opened 2017-07-21
Master

soef:
find-my-iphone version incremented


Apollon77:
Let's wait till user agree that 0.2.9 is fixing the one issue.

Idea for stable repo is: after users that use "Latest" (so npm version) tell that new version works and do not introduce new problems then update stable repo for all users

GermanBluefox:
Do you have confirmation from users?

PR #17, opened 2017-08-24
add benq and mikrotik

instalator:


Apollon77:
Do you have test threads and results from forum?
Please activate testing in Travis-CI/Appveyor tp validate that it do not break.

Thank you!

Apollon77:
And you need to add to the other json file too for latest repository ... only stable will not work :-)

Apollon77:
Ahh was already in  latest :-)
Forum testing threads would be great in any case :-(

PR #23, opened 2017-08-27
I2C and KeContact adapters

UncleSamSwiss:
New adapters, not yet ready for "stable", but good enough to be taken into ioBroker.

### DIFF
--- a/sources-dist.json
+++ b/sources-dist.json
@@ -199,6 +199,11 @@
     "icon": "https://raw.githubusercontent.com/ioBroker/ioBroker.hue/master/admin/hue.jpeg",
     "type": "lighting"
   },
+  "i2c": {
+    "meta": "https://raw.githubusercontent.com/UncleSamSwiss/ioBroker.i2c/master/io-package.json",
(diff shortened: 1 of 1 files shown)


Apollon77:
Do we also have testing threads for those in Forum?
Please make sure testing for i2c is corrected (remove macos ...)

UncleSamSwiss:
Testing threads are:
- KeContact: http://forum.iobroker.net/viewtopic.php?f=23&t=6715
- I2C: http://forum.iobroker.net/viewtopic.php?f=23&t=7437 (unfortunately without any feedback except on this thread: http://forum.iobroker.net/viewtopic.php?f=23&t=4464#p77488)
Thanks for the feedback on the testing. Also right now I can't really test it on any system (even Linux) since it requires an NPM package that won't work on a regular Linux. What do you suggest about that? Currently I just don't do anything at start-up when the adapter is not run on a supported system.

Apollon77:
testing is great that way, when adapter only works on linus then it's ok to have testing only there ...

PR #29, opened 2017-09-04
Add broadlink2 and bmw adapters

frankjoke:
Maybe the standard npm does not check repositories with lower case iobroker only :(


Apollon77:
There is no "standard nom check" for adapters. Everything is handled by Pull Requests here and after it's in it is checked too.
Please check, JSON do not seems to be valid!!

frankjoke:
Resolved by removing/adding again bmw.

PR #31, opened 2017-09-06
Change version Mikrotik adapter

instalator:


Apollon77:
is this a special bugfix? Normally the idea is to update versions into "stable" as soon as they are prooved by multiple users in production ... 0.0.20 is 2 hours old...

PR #34, opened 2017-09-08
update landroid-s

MeisterTR:


Apollon77:
Is 1.0.1 already tested and error-free-prooved by users? Release of 1.0.1 was 15h ago

MeisterTR:
i found a litle error. i think we wait just a few days :)

PR #37, opened 2017-09-12
Updated laRenamed and added adapterstest adapters renamed

frankjoke:
Added bmw and broadlink2-Adapters and changed names where adapters have been renamed to ioBroker

### DIFF
--- a/sources-dist-stable.json
+++ b/sources-dist-stable.json
@@ -29,6 +29,12 @@
     "version": "0.0.6",
     "type": "multimedia"
   },
+  "bmw": {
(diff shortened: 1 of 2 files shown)


Apollon77:
Please also add to the second file when you add in stable. And please check that sort order is correct ... broadlink2 should come after broadlink, or ?! ;-))

frankjoke:
I updated both files and put adapters in right order!

Apollon77:
But now we have a conflict. COuld you please pull and merge and push again?! Thank you!

frankjoke:
Ok, I made the changes without the BMW-adapter in the latest because the ble-adapter was added and the change conflicts with it. If you did the merge I will add it in a next pull request.

PR #43, opened 2017-10-02
(ldittmar) Sony Bravia TV Adapter stable

ldittmar81:


GermanBluefox:
I still wait for discovery :)

PR #44, opened 2017-10-06
increased EgiGeoZone stable version

BasGo:


Apollon77:
Hey, is the version already proofed by other users? Or is it fixing a bug?
The idea is to update stable after the adapter version got used by some people in "latest" and so is "proofed" to be stable ... 2h after release this is normally uncommon ;-)
So I would wait till merge for some days

BasGo:
Hi Ingo,

it is a bugfix version: it fixes an issue where authentication is used. Actually it is not possible to use username and password because the check routine

(username !== config.username) || (password === config.password)

is wrong. For the next releases I‘ll wait some days ... like you wrote ;-)

Best regards

Bastian


Von meinem iPad gesendet

Apollon77:
When Bugfix like this we can accept update :-) Thank you!

PR #50, opened 2017-10-16
Add BLE and G-Homa to stable

AlCalzone:


GermanBluefox:
Is g-homa not the iot-system ?

Apollon77:
@GermanBluefox No not really, when I see it these are Wifi controlled Devices ... and an App to use them or you can use ioBroker :-) 
@AlCalzone Correct?

AlCalzone:
"G-Homa" in general is a system of App, Alarm sensors, WiFi plugs, etc. 
The adapter g-homa only supports the basic G-Homa WiFi plugs, so I'd label it as "hardware"

PR #52, opened 2017-10-16
(ldittmar) change type of haier

ldittmar81:
The haier adapter is an air conditioning adapter


Apollon77:
PS: I will also fix it in the latest repo!

PR #53, opened 2017-10-16
(ldittmar) change type of haier

ldittmar81:


Apollon77:
Hehe I was faster manually ... so already ok

PR #73, opened 2017-11-25
Update Google Home

angelnu:


angelnu:
@GermanBluefox - I bumped to 1.3.5 as preparation for a fix in sayit. There is no need to update chromecast before sayit but the fix in sayit will not work without both newer versions.

PR #101, opened 2018-02-25
update publish date and version in ioBroker.vr200

Eisbaeeer:
Is it realy usable to update the repo on ervery update of adapters regarding the publish date?
Eisbaeeer


Apollon77:
No!!! Published is only set on initial (first) publish. Do Not change it on Updates. 

Also Version Updates in stable should only be done after new version was used by some users in latest and so is really proofed

PR #106, opened 2018-03-21
changed telegram version, added places adapter

BasGo:
Like talked to @Apollon77 in the afternoon, here's the PR.

PR #107, opened 2018-03-22
Fixed wrong package version for telegram

BasGo:
See my comment on PR #106


Apollon77:
So 1.2.0 contains even more things ... should we really  do this in stable directly ?! 1.1.4 should be available on  npm

BasGo:
This PR was just meant for the case that there's no 1.1.4 on NPM. But there **is** a 1.1.4 on NPM, so imho we should wait a bit.

PR #114, opened 2018-04-20
Renamed ioBroker.lgtv2011 to ioBroker.lgtv11

SebastianSchultz:


SebastianSchultz:
I had to rename ioBroker.lgtv2011 to ioBroker.lgtv11 because of npm package naming restrictions.

PR #122, opened 2018-06-01
Update sources-dist.json

schmupu:
Adding lupusec to latest

### DIFF
--- a/sources-dist.json
+++ b/sources-dist.json
@@ -510,6 +510,12 @@
     "type": "iot-systems",
     "published": "2017-04-24T19:18:46.399Z"
   },
+    "lupusec": {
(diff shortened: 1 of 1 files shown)


Apollon77:
Lupusec is Alarm system, correct? Then please use (in this repo commit and also in your io-package.json) "alarm" as adapter-category please (instead of general/utility mixture)!

Do you have added Bluefox to NPM as owner?

Please add README info that it only supports Admin-v3! (because not stable till now)

It would be cool to offer io-package and Admin translations in all languages (you can use http://iobroker.net:3000/ ...)

Please remove www directory from adapter if not needed.

Thank you!

PR #124, opened 2018-06-03
Update sources-dist.json

Jey-Cee:


Apollon77:
Ich dachte das entfernen des published gilt nur für stable?! Hm ...

GermanBluefox:
"published" is the information, when this adapter was published in this repo for the first time!
It is not necessary (and even wrong) to modify it every time. 

Latest version time is "versionDate" and will be generated automatically.

Jey-Cee:
Thanks for bring light into darkness. 
Reason was that the actual version on latest shown in my admin was not actual. But now it is, so everything is ok.

PR #126, opened 2018-06-03
update version

twonky4:


BuZZy1337:
Published date must not be changed. It shows the date when the adapter was released for the first time.

Apollon77:
Published date should never be updated.

twonky4:
publish date change reverted

Apollon77:
sources-dist.json (=Latest) should not have a version at all ...

twonky4:
version removed from sources-dist.json

PR #128, opened 2018-06-05
add yeelight adapter, mihome-vacuum and landroid s in stable

MeisterTR:


Apollon77:
Testing with yeelight is red becaue odf a name conflict: yeelight-2 in package.json vs yeelight in io-package??
names must match! (beside the capital vs non-capital "B" :-)

MeisterTR:
now all names are the same...

PR #131, opened 2018-06-10
Zigbee quick fix version

kirovilya:
0.5.3 version was not open port with updated version serialport (was wrong parameter name)


GermanBluefox:
This is a big step for native support of ZigBee devices!
Please note: Published attribute is the date of the very FIRST publish event. 
It is not necessary (and even wrong) in the future to update it

kirovilya:
Ok. I understood. Thank you.

PR #152, opened 2018-08-08
added sonnen adapter

foxriver76:
-added sonnen to latest repo
-added sonnen to stable repo

### DIFF
--- a/sources-dist-stable.json
+++ b/sources-dist-stable.json
@@ -919,13 +919,6 @@
     "type": "weather",
     "published": "2016-06-01T10:34:25.177Z"
   },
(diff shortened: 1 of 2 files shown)


foxriver76:
if its to early for stable, please cherry pick latest.

Apollon77:
Please update your commit to only contain latest :-) Thank you

foxriver76:
Will discovery be able to detect my adapter when it is in latest, without modifying the discovery adapter (like https://forum.iobroker.net/viewtopic.php?f=24&t=15723). Otherwise there should be some information on it, because requirement for stable is discovery implementation.

PR #156, opened 2018-08-13
Add klf200 to repository

MiSchroe:


GermanBluefox:
Please add bluefox to owner on npm

MiSchroe:
Added bluefox as owner.

GermanBluefox:
Can it be level.blind  ?

https://github.com/MiSchroe/ioBroker.klf200/blob/master/main.js#L430


Indicator cannot be writable.
https://github.com/MiSchroe/ioBroker.klf200/blob/master/main.js#L463

It is switch.

MiSchroe:
> Can it be level.blind ?
>https://github.com/MiSchroe/ioBroker.klf200/blob/master/main.js#L430

The types which I check for can be blinds or windows. I will change the type to level.blind for all blind types, what is the best level for a window?

> Indicator cannot be writable.
> https://github.com/MiSchroe/ioBroker.klf200/blob/master/main.js#L463

I will fix that.

GermanBluefox:
>I will change the type to level.blind for all blind types, what is the best level for a window?

There is no "level.window".  But we can create it. 
Take level.blind for now.

MiSchroe:
I have changed the role of the silent state to `switch` and the level will be `level.blind`. I have moved the latter to a separate function, which makes it easier to change it to something like `level.window` when that will be available.

PR #159, opened 2018-08-16
discocery version raised to 1.2.2

unltdnetworx:
Don't know, if it's my turn to raise the version of the discorvery adapter because of the new pattern for the energymanager.


unltdnetworx:
Changed the date back to original publish date.

Apollon77:
No, this is too fast!

Normal process: new version is published on npm, will go into latest ... then users installing it and stable update only if "proofed" working ... So this needs to stay here some days till we know that it do not has side effects.
Also also adapter itself needs to be in stable repo too ... is it for yours?

unltdnetworx:
"Again what learned" as a well known german soccer player says. ;)

The adapter is in repo and works for me. No complainments or negativ feedback till now.

PR #168, opened 2018-08-29
add mihome-lamp, proxmox,kress and yeelight to stable, update landroid-s version

MeisterTR:


GermanBluefox:
Dors for some adapters the discovery possible

MeisterTR:
yeelight should be easy. mi lamp, vaccum and plug too but i have to find a way to select the difference. proxmox is a bit difficult, maybe by the port.. kress and worx only by the network name or also by a special port. i will take a look at it in the next time...

PR #173, opened 2018-09-09
 initial confirmed stable release of stiebel-isg

unltdnetworx:
initial and confirmed stable relase of the stiebel-isg adapter for stiebel-eltron/tecalor internet service gateway. (heatpumps)

forum-thread:
https://forum.iobroker.net/viewtopic.php?f=24&t=14309


unltdnetworx:
will repair the conflicts

PR #176, opened 2018-09-12
Update sources-dist.json add ioBroker.iogo

nisiode:


Apollon77:
Please also fix or remove the wrong translated texts in io-package. You can use http://iobroker.net:3000/ to get automatic translations for all languages.

Please remove docs information from io-package if no docs files are provided

Your code seems to contain other stuff from template adapter or such (checkPermission) which seems wrong ... Please check your code

GermanBluefox:
@Apollon77 Looks ok, isn't it?

Apollon77:
Yes ... lost it in my emails :-( will merge it

PR #178, opened 2018-09-14
added wiffi to stable repo and updated recent to recent version of wiffi

t4qjXH8N:
this version seems to be stable according to the feedback in the forum


Apollon77:
Please remove "version" from latest  repo file because latest version is always queried from npm automatically. And please do not change published at all. eill stay on date when first published

t4qjXH8N:
Sorry, I added wiffi to stable and reverted latest

PR #190, opened 2018-10-02
Add Xbox to latest

foxriver76:
- Add Xbox adapter
- tested in https://forum.iobroker.net/viewtopic.php?f=36&t=17044&p=178951#p178951
- bluefox added as npm owner


Apollon77:
Thank you!

PR #195, opened 2018-10-12
added stable version 2.0.0 of the wiffi-wz adapter

t4qjXH8N:
This version of the adapter supports all products from stall.biz that support sending a datagram. This adapter was tested by me and at least one other user with a different setup. I suppose it is stable.


t4qjXH8N:
Sorry for doing this wrong again.

t4qjXH8N:
Maybe "first_published" would be a better name for the field?

Apollon77:
no problem ...

GermanBluefox:
for first_published is too late. Attribute used on many places

PR #198, opened 2018-10-21

nisiode:
Add iogo in version v0.1.0


Apollon77:
Please add current data as published here

nisiode:
Changed "published" to current timestamp...
Where is this data being used for?

Apollon77:
Statistic stuff and maybe later to show „new“ tags in admin or such

GermanBluefox:
> Statistic stuff and maybe later to show „new“ tags in admin or such

Yet used. 
image

PR #199, opened 2018-10-24
Update backitup

simatec:


Apollon77:
Please do not change published! Will always point to first release!

simatec:
OK, I will pay attention to it

foxriver76:
@simatec Make a second commit with the original timestamp please.

simatec:
@foxriver76 I have fix the original timestamp

PR #206, opened 2018-11-02
Updated RWE/Innogy adapters

PArns:
- Updated latest stable version of innogy smarthome adapter (required by the INNOGY/RWE team)
- Removed RWE SmartHome adapter as the required software is no longer available


Apollon77:
Please remove versiondate and do not update it

Apollon77:
SOrry, just realized that updating also works :-)

PR #232, opened 2018-12-08
add opentherm adapter to latest

DutchmanNL:
Opentherm adapter allows to read data from opentherm gateway (heating).

currently in beta state and testing at 3 installations, in latre releases control of several values will also be possible


Apollon77:
Great to see!

Why also published date from discovery is changed? That stays as it was on initial publish. ;-) please revert this change. 

Rest of requirements for repo done? 

Will check later today.

DutchmanNL:
jup other requirements done but always good to have 4 eyes on that :)

Date discovergy changed as it was incorrect, please not discoverGy not discoverRy ;-)
that one was release by me this week but states 2017

Apollon77:
Great,

PR #250, opened 2018-12-31
added asuswrt to sources-dist.json

mcdhrts:
Thx

### DIFF
--- a/sources-dist.json
+++ b/sources-dist.json
@@ -48,6 +48,13 @@
     "published": "2018-11-10T15:30:17.077Z",
     "versionDate": "2018-11-29T18:57:58.881Z"
   },
+  "asuswrt": {
+    "meta": "https://raw.githubusercontent.com/mcdhrts/ioBroker.asuswrt/master/io-package.json",
+    "icon": "https://raw.githubusercontent.com/mcdhrts/ioBroker.asuswrt/master/admin/asuswrt.png",
+    "type": "hardware",
+    "published": "2018-12-31T13:39:00.077Z",
+    "versionDate": "2018-12-31T13:39:00.077Z"
+  },    
   "b-control-em": {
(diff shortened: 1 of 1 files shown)


Apollon77:
Hi @mcdhrts,

before we can accept the adapter please verify anything which is listed in the repository requirements!
see https://github.com/ioBroker/ioBroker.repositories#requirements-for-adapter-to-get-added-to-the-latest-repository Most is covered already

Especially testing is not complete. The adapter testing is missing. And please check the test output, there are some warnings.
For Admin3 support you missed to set "common.materialize=true" in io-package.

mcdhrts:
Hi @Apollon77,

"common.materialize=true" is set in line 5 when i´m not wrong.

The only thing i can find which is missing is testing with appveyor. I was thinking Travis-CI was enough but this part it seems i was wrong (Travis-CI AND Appveyor). 

Was that all that i have unseen? Then i will have a look at Appveyor.

Thank you.

Apollon77:
Appveyor is not really needed, Travis is enouh. And you can in the meantime also add "windows" using travis :-)

Apollon77:
If you need support, we are here!

mcdhrts:
Hi Apollon,
thank you for your help, yesterday i seen the missing testfiles and add the files and go through the tests. it seems that i use a function which is only usable with nodejs 6 or higher.
i will check if i can get around this or if i set the adapter with the requirement nodejs 6 or higher.
When im done with that i will make a new version and a new pull request here :-).

P.S. the Admin3 warning comes everytime, i realy don´t get why, the materialize=true is set in io-package.json

Thank you

Apollon77:
For admin3 warning: maybe the GULP file is missing? Also copy from template :-)

see: https://github.com/ioBroker/ioBroker.template/blob/master/test/testPackageFiles.js#L65

PR #252, opened 2019-01-02
add mihome-airpurifier

JoJ123:


Apollon77:
Hey,

I did a short check and noticed some things to fix.
* please add the valid configuration options in io-packge in native with their default values. (like ipaddres, token and these).
* please remove the docs info from io-package because no docs provided there. and then best also remove the docs directory (or better provide docs there ;-)) )

Then you should be ready :-)

Thank you

PR #255, opened 2019-01-06
update of my adapters and add statistics

foxthefox:


Apollon77:
Formal review:
* please add license field in io-package in common.license

I will accept without but please fix this!!

PR #266, opened 2019-01-17
new version for snips-adapter

unltdnetworx:


unltdnetworx:
It is an important bugfix, because of the versionnumber in npm published by the owner of the repo and an major bug, giving soundfeedback twice.
The meaning is known to me and is also considered useful. Test versions will be published in sources-dist.json.

PR #268, opened 2019-01-17
add hm ip adapter

jogibear9988:


Apollon77:
Lets wait with stable if tested by users and some more devices are in it please !

PR #274, opened 2019-01-18
update opentherm to new repro

DutchmanNL:


Apollon77:
Ääähm really??

DutchmanNL:
? Thema Adapter ist moved from my repro to Community Adapters So The entry must be changed

jogibear9988:
why is the test broken?

Apollon77:
URLs wrong

PR #278, opened 2019-01-19
pilight

arteck:


Apollon77:
Why the adapter developer itself is not sending this PR?

- Admin3 support is missing
- Bluefox is no npm owner

.. and maybe more reuirements not fullfilled.

Sorry in this status I can not accept the adapter.

Apollon77:
@arteck Do you see any chance at the moment that the topics can be handled? Do you have contact to developer? Else I would close this PR and a new one can be created once things are handled... Sorry

arteck:
ahh ok.... i close here

PR #301, opened 2019-01-28
Edited lgtv to v1.0.7

SebastianSchultz:


Apollon77:
Important bugfix or what in stable that fast after release?

Issue #302, opened 2019-01-28
Version from npm instead of sources-dist.json


Zefau:
Hi,

I updated the nello adapter recently to v2 on Github (on master branch) and kept v1 on a dedicated branch (v1). 
The `sources-dist.json` is updated with the v1 Github branch, but npm has been updated to v2 branch.

Though, ioBroker now shows v2 branch as the current branch. From my understanding, the branch from `sources-dist.json` should have been shown instead?

Cheers,
Zefau

Apollon77:
What exactly you mean? What is shown where? Screenshots?

Zefau:
ioBroker.nello has 2 branches (master and v1)
![screenshot_2019-01-28 zefau iobroker nello](https://user-images.githubusercontent.com/14368811/51863111-7b003100-2340-11e9-88d8-6f3b3b2bf82e.png)

ioBroker latest repo is told to use v1
![screenshot_2019-01-28 iobroker iobroker repositories](https://user-images.githubusercontent.com/14368811/51863112-7b003100-2340-11e9-8ef3-1adb9d08ab58.png)

though, npm has v2 as current version
![screenshot_2019-01-28 npm](https://user-images.githubusercontent.com/14368811/51863113-7b003100-2340-11e9-9ccf-fe767806c8dd.png)

ioBroker is using v2 as current version (not v1, as stated in second screenshot)
![screenshot_2019-01-28 adapters - iobroker](https://user-images.githubusercontent.com/14368811/51863203-b864be80-2340-11e9-83d9-54408250fd0e.png)

Zefau:
The other way around: I want v1 to be available in ioBroker (as stated in latest repo) and not v2, which is latest in npm.

Apollon77:
This is not possible. The urls are only used to show the icon in admin and to „the relevant io-package“ to be used for some stuff. For latest repo the real version is always got from npm

The idea is to set the latest tag on npm to the last v1 Version and use other Tag für v2 (Beta or such) when publishing. After publishing you can use npm dist-tag to change tags.

PR #306, opened 2019-02-03
nello: updating to v2

Zefau:
updating v2 in both repos has no effect since npm is already on 2.0.1 (see https://github.com/ioBroker/ioBroker.repositories/issues/302)


Apollon77:
For stable it would have an effect ... really merge it?

PR #310, opened 2019-02-06
Smappee V004

forelleblau:


Apollon77:
Is it an important bugfix? Or why that fast in stable?

forelleblau:
Thanks for the question, I know now, that i did it wrong in the past. I'll show up here later.

Apollon77:
;-) no problem. 

The idea of stable repo is really to have Version in there that had „settled“ and proofed by users. 
Exceptions are important bugfixel

PR #315, opened 2019-02-10
Add ioBroker.nuki2 to latest repository

Zefau:
I have reworked the existing ioBroker.nuki adapter and optimized the code.
Furthermore, I have added the following features:
- using both Bridge API and Web API (current adapter only uses bridge API)
- support for multiple bridges
- support for discovery within admin panel
- additional states for bridges and better separation between software / hardware bridge
  - retrieve the basic and advanced configuration from the lock
  - retrieve all users having access to the lock
- added Web Interface to view logs

### DIFF
--- a/sources-dist.json
+++ b/sources-dist.json
@@ -966,6 +966,13 @@
     "published": "2018-10-05T09:37:04.501Z",
     "versionDate": "2019-01-21T13:37:47.916Z"
   },
+  "nuki2": {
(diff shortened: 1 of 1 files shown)


Zefau:
The PR into the ioBroker.nuki adapter has been rejected by the developer (see https://github.com/smaragdschlange/ioBroker.nuki/pull/11#issuecomment-456061232) because my changes have been quite significant and the developer wants to keep the adapter simple and be able to maintain it.

PR #330, opened 2019-02-16
mieleathome

hash99:
Error corrected.

Sorry.


Apollon77:
Hi,

sorry, but there are some things not correct.
Please see: https://travis-ci.org/ioBroker/ioBroker.repositories/builds/494321217#L614
Please correct the io.package file!

Additional to this please do not append the adapter at the end of the repo-files, but sort in alphabetically. Thank you!

Please also check the infos in the adapter testing like:
https://travis-ci.com/hash99/ioBroker.mieleathome/jobs/178292997#L775

Please correct the install info in your readme ... https://github.com/hash99/ioBroker.mieleathome/tree/master#steps user should use Admin ... or at least not in node_modules ...
Please delete "widget" and "www" directories if no widget or webpage is included. Also remove the docs directories if there are no docs in there - please also remove them from io-package.

Further more:
* https://github.com/hash99/ioBroker.mieleathome/blob/master/package.json#L15 is wrong
* https://github.com/hash99/ioBroker.mieleathome/blob/master/package.json#L42 is wrong
* without a vis widget please remove https://github.com/hash99/ioBroker.mieleathome/blob/master/io-package.json#L62
* please put in https://github.com/hash99/ioBroker.mieleathome/blob/master/io-package.json#L64 native all adapter config fields with their default values
* please use setObjectNotExists instead of setObject to make sure it is not overwritten. This could remove custom specs from it which is not intended
* please only use roles defined at https://github.com/ioBroker/ioBroker/blob/master/doc/STATE_ROLES.md in your objects
* personal note: using a "hard interval" like in https://github.com/hash99/ioBroker.mieleathome/blob/master/main.js#L276 combined with external communications is a bad idea because the schedule will also run when there are network problems or such and so requests may sum up. Better is to use setTimeout to just delay till next run.
* is it really a good idea to instanciate a new Miele object on any state change? there is already one in main ...

Apollon77:
Can you please sort the entry in on correct place depending on name ... alphabetically.
What about my other topics ?

hash99:
Hopefully everything correted.

PR #339, opened 2019-02-27
Add mieleathome

hash99:


hash99:
Please check.

Apollon77:
basically ok, but what is the "iobroker-data" directory in the repo?
Also all those "MACOS" files with . at the beginning should not be in reoo (are they maybe also in npm package)

Additionally: on npm is already 1.0.0 ... Do you still want 0.0.4 to come into stable?

Apollon77:
@hash99 ping :-) any news here?

Apollon77:
@hash99 any news?

PR #341, opened 2019-03-01
Update info adapter to 1.1.3

ldittmar81:


Apollon77:
Do you have enough Tester Feedback for stable?

ldittmar81:
I guess so. There were 4 external users reporting bugs. This was fixed and then everybody said it worked fine. The current stable version does not work, because it shows on the old forum.

PR #346, opened 2019-03-07
Core Files/Testing Update and introduce adapter-core to solarlog, sma…

forelleblau:
…ppee and comfoair


GermanBluefox:
There is no need to update latest repo for time. It will be done automatically

PR #349, opened 2019-03-12
Update sources-dist.json

unltdnetworx:
new stable versions for energymanager and stiebel-isg


Apollon77:
not needed for latest. WIll be updated automatically twice a day

PR #350, opened 2019-03-12

unltdnetworx:
new stable versions for energymanager and stiebel-isg


unltdnetworx:
Yes, it is.

unltdnetworx:
It is tested with the community and then directly pushed into stable. Will change the workflow.

Apollon77:
ok, thank you

PR #352, opened 2019-03-13
Update sources-dist.json

AlGu1:
hombot adapter added


Apollon77:
Please check the following:
* remove docs, widgets, www folders if nothing changed there and for docs also in io-package or provide the needed content there

Rest looks ok

AlGu1:
Hi,
ok thanks, the folder "docs" and "www" deleted. Folder "widgets" are in use and "docs" node removed from io-package.json
AlGu

PR #356, opened 2019-03-15
Update solarlog, smappee & comfoair

forelleblau:
solarlog: previous versions were tested sucessfully, actual version is readme - update (badges etc. only)
smappee: previous versions were tested sucessfully, actual version is readme - update (badges etc. only)
comfoair: first pub in stable. Versions 0.0.7 - 0.1.1 were tested successfully (see issues). V 0.1.2 is a readme-update only.
Thanks a lot for the new badges, thats a great!


Apollon77:
>  Thanks a lot for the new badges, thats a great!

You are welcome :-)

PR #364, opened 2019-03-19
Add new radar2-adapter to stable.

frankjoke:


Apollon77:
Hi,

you could delete all the "grund" stuff including the tasks directory because no longer needed and replaced by gulp

PR #374, opened 2019-03-25
Major 'bugfix' Comfoair

forelleblau:
Please publish version as stable to publish the remark in the README that a parallel use of the adapter with the device' control panel is not recommended!


Apollon77:
Normally readme is fetched from GitHub and not local but I accept the pr ... also without an updated versioninfo field ;-)

PR #389, opened 2019-04-13
Update ASUSWRT Version to 1.0.1

mcdhrts:


Apollon77:
already 27 installations, can be considered as stable I think :-)

PR #392, opened 2019-04-17
new stable versions for DasWetter, vis-weather, ebus and sbfspot

rg-engineering:


Apollon77:
all new versions installed already 20+% ... so completely fine for stable :-)

PR #394, opened 2019-04-18
add iobroker.valloxmv adapter

hacki11:
please let me know if something is missing


Apollon77:
You should set "published" to today :-)

PR #403, opened 2019-04-28
Update BLE stable version to 0.7.2

AlCalzone:


Apollon77:
Please also Update Versiondate Field or remove it

AlCalzone:
Done. 

Where is this documented? I double-checked with `npm run update ble` which also does not change nor remove this field.

Apollon77:
Repo readme I think

AlCalzone:
Ok I was blind. The update script should be fixed then.

PR #406, opened 2019-04-30
Squeezeboxroc Adapter latest

oweitman:


GermanBluefox:
I created a couple of issues :https://github.com/oweitman/ioBroker.squeezeboxrpc/issues

oweitman:
> I created a couple of issues :https://github.com/oweitman/ioBroker.squeezeboxrpc/issues

The issues are fixed

Apollon77:
hey, I also found some things:
* you subscribe to the states twice (one time in main() and one time in the server class. better remove one :)
* you use an interval to disciver things and do external communication. M personally do not like Intervall here because the intervall fires new cals unless f the action from before has finished. When you run into network timeouts this may cause reuests piling up

Apollon77:
And please sort your adapter in on correct alphabetical position!!

oweitman:
> hey, I also found some things:
> 
> * you subscribe to the states twice (one time in main() and one time in the server class. better remove one :)
> * you use an interval to disciver things and do external communication. M personally do not like Intervall here because the intervall fires new cals unless f the action from before has finished. When you run into network timeouts this may cause reuests piling up

both points are fixed and a new release was published to github and npm

oweitman:
> And please sort your adapter in on correct alphabetical position!!

the pull request was updated with another commit to move the adapter in alphabeticaly order

GermanBluefox:
Thank you!

PR #408, opened 2019-05-04
ioBroker.roomba v1.0.0 to stable

Zefau:


Zefau:
Forgot to create a new branch on my fork for the second pull request, so it got added to this one.

Second pr is to removed Tautulli from latest repo (it is not in stable repo), because the functionality is integrated in the ioBroker.plex adapter. Tautulli itself is useless without Plex, so a dedicated adapter makes no sense.
ioBroker.tautulli repo got archived, see https://github.com/Zefau/ioBroker.tautulli.

PR #409, opened 2019-05-05
update bosesoundtouch stable version to 0.2.4

SwedishChef:


Apollon77:
ok, installed >20 times (5%) so I think should be ok

PR #410, opened 2019-05-06
add gruenbeck adapter to latest

TA2k:
https://www.forum.iobroker.net/topic/21377/aufruf-iobroker-gr%C3%BCnbeck-adapter

### DIFF
--- a/sources-dist.json
+++ b/sources-dist.json
@@ -384,6 +384,13 @@
     "published": "2016-01-15T20:18:56.071Z",
     "versionDate": "2018-03-07T21:05:22.640Z"
   },
+  "gruenbeck": {
+    "meta": "https://raw.githubusercontent.com/TA2k/ioBroker.gruenbeck/master/io-package.json",
+    "icon": "https://raw.githubusercontent.com/TA2k/ioBroker.gruenbeck/master/admin/gruenbeck.png",
+    "type": "household",
+    "published": "2019-05-06T08:44:23.717Z",
+    "versionDate": "2019-05-06T08:44:23.717Z"
+  },
   "habpanel": {
     "meta": "https://raw.githubusercontent.com/iobroker-community-adapters/ioBroker.habpanel/master/io-package.json",
     "icon": "https://raw.githubusercontent.com/iobroker-community-adapters/ioBroker.habpanel/master/admin/habpanel.png",


Apollon77:
Hi,

the adapter in general looks good, but I have a few command and one change request:

* Change-Request: in onStateCHanged you try to find out if the value has changed by checking lc and ts and such. But you completely ignore the ack-flag which is "the" field to work with here! ack== false means "user wants to control the device" and you should simply check if ack===false and then control ... also if the value may not have changed (unless device don't like that). SO you can simplify the code here very much and also bring in line with the "standard". Please change this before I add to the repo.
* I personally would never use setInterval as soon as you are doing external requests. Reason: The request can run into a network timeout or such, but the interval will fire without noticing and unless you check that in your code (which can get complex very easiely). I more like setting one setTimeout for the next call once the current communication is done, so it stays sequencial. I see that you also try to block communication and added handfling ... but with all the intervalls firing more and more commands into the queue this call can get confusing ... maybe rethink it again :-) This is not a blocker for repo.
* you also add more and more requests in the queue based on interval which can lead to "the same requests are called many times" if one got blocked once. You should only add to the queue if the same command is not already in there. This is not a blocker for repo.
* Also "Unload" should make sure that no more communication is running and make sure no new gets started (because you are using some timeouts also in the request Method ... make sure to really clean up anything to end the adapter. This else would be problematic in compact mode.

Apollon77:
>  This means every state change has an ack also if they are not changed.
Correct, all values that the adapter stores should have ack=true and thats why you can use this to ignore all those in the "on-State-changed" method. Because it is the rule that if a user wants to control a device the value needs to be set with ack=false.
So the basic definition is that "onStateChange" uses ack flag to detect if it should use an updated value to send to the device (ack=false) or not. There are only very little cases where it makes sense also to handle ack=true changes there; in 99% of the cases just ignore the value as soon as ack===true and send to device if ack===false

Please adopt to this mechanism or lets discuss about the exact reasons not to :-)

TA2k:
But the adapter calculates new states based on state changes coming from the device this means with ack.
How should I calculate states when I ignore ack states in the onstatechanged method. And the calculation should only start when the value is changed.

Apollon77:
Aahhh now I get it ... you do not use onStateChanged to control devices ... you use it to calculate values ... ok then it makes sense, but then I think you need to ignore all values with ack==false because if someone (why ever) tries to control things even if not controllable it would interfere ... The other thought is if you not should try to make trigger the calculation on the places whewre you set the values ... because in "OnStateChnged " also "ack===true" changes from a user would be used for calculations (but they never came from you as the adapter)

PR #421, opened 2019-05-27
ValloxMV going stable

hacki11:
added vallox mv release

got some positive feedback already:
https://forum.iobroker.net/topic/21897/aufruf-valloxmv-adapter


Apollon77:
1.0.0 is installed by noone according to stats ... any very new ... what was the change compared to 0.1.3 which is most installed so far?

hacki11:
Just a small fix, changes are documented here https://github.com/hacki11/ioBroker.valloxmv/blob/master/README.md

i do not want to go stable with a 0.x version, it is just a version bump and a small bugfix.

PR #423, opened 2019-05-29
new adapter pi-hole

unltdnetworx:
Forum-Thread for first testings:
https://forum.iobroker.net/topic/22715/aufruf-neuer-adapter-iobroker-pi-hole

### DIFF
--- a/sources-dist.json
+++ b/sources-dist.json
@@ -1204,6 +1204,13 @@
     "published": "2015-01-02T23:47:36.408Z",
     "versionDate": "2017-09-20T20:59:50.369Z"
   },
+  "pi-hole": {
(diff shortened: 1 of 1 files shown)


Apollon77:
Welcome to the repository.

One personal code-review note: Using intervalls is dangerous because the intervall fires regardless if the former request was finished. So on network issues or load problems you may end up in having many parallel requests running against the server ... It is better to use setTimeout and set the next timeout once the request before was finished ... so you make sure that you only have one request active at any time.
Additionally having "setObjectNotExist in each "Loop" is not ideal from a performance perspective, it would be better to only do this in the first "Loop".

PR #441, opened 2019-06-26
Update homeconnect version

TA2k:
updated readme
remove states created from previous versions of the adapter


Apollon77:
please also update or remove versionDate field

TA2k:
removed versiondate field

PR #444, opened 2019-07-01
Update ZWave stable version

AlCalzone:


Apollon77:
Had 1.3.1 also the Forbidden characters reference to js-controller? Should not a 1.4.2 go to stable?

AlCalzone:
You're right, but a 1.4.2 would have to wait a bit before stable. This one is motivated from users not being able to install 1.3.x anymore.

Apollon77:
But because js-conttoller 1.5.11 being stable an upfdate to 1.4.1 would mean many problems, or?

AlCalzone:
Hotfixed to 1.4.2 with the regex/replace change

PR #445, opened 2019-07-02
Update sources-dist.json

TA2k:
Added nightscout Adapter
https://forum.iobroker.net/topic/23575/aufruf-adapter-für-nightscout

### DIFF
--- a/sources-dist.json
+++ b/sources-dist.json
@@ -1071,6 +1071,13 @@
     "type": "climate-control",
     "versionDate": "2019-03-24T15:12:00.840Z"
   },
+    "nightscout": {
+    "meta": "https://raw.githubusercontent.com/TA2k/ioBroker.nightscout/master/io-package.json",
+    "icon": "https://raw.githubusercontent.com/TA2k/ioBroker.nightscout/master/admin/nightscout.png",
+    "published": "2019-07-02T19:15:28.166Z",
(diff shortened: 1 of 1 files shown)


GermanBluefox:
Missing description

TA2k:
I added a description in the readme. short description already in package and io-package
Nightscout (CGM in the Cloud) is an open source, DIY project that allows real time access to a CGM data via personal website, smartwatch viewers, or apps and widgets available for smartphones.

Apollon77:
Some review comments from me too:
- you should remove .vscode directory from GitHub and set an gitignore
- you can use https://translator.iobroker.in/ to translate news and such in all relevant languages, please add that for the news in future :-)

Apollon77:
Oh found one thin: you set info.connection state ... but the object is never defined. Normally it is done as instanceObject in io-package ... please check that!

TA2k:
Thanks for the advices. But it strange thate connection is working without defined in io-package.
Maybe a short docu what the best practice for translation is at the moment because with this 5 options in the gulp script it is not easy to find the correct way

ldittmar81:
I wrote a short description about Auto-Translation in German here: https://forum.iobroker.net/topic/19047/gulp-ist-kein-hexenwerk-auto-translation

PR #446, opened 2019-07-03
update hmip to v0.0.11

jogibear9988:


Apollon77:
Important fix or why so fast for stable?

PR #453, opened 2019-07-21
added countdown adapter

jack-blackson:


Apollon77:
Please see error. The adapter needs to be added to "latest" repository file too ... any normally first.

Your adapter also should support the compact mode.

Why the adapter is disabled by default?

Please also make sure to provide all languages. you can use https://translator.iobroker.in/ to get translations.

jack-blackson:
Errors are fixed, and version 1.0.2 was published on npm and on github

jack-blackson:
@Apollon77  I just saw that I also had to change the committed change to the new version - this is now also fixed.

Apollon77:
I checked again. 
* v1.0.2 is not contained in the adapter news.
* please remove "countdown" from the list of restart adapters (this list if for "other" adapters to be restarted.
* please define all values from admin_m.html also in io-package native with their default values
* @iobroker/testing should be a devDependency only
* you access states using Foreign methods to the instance 0 which will not work with multiple instances. normally simply use the "not-Foreign" methods and then without "countdown.0."
* the main method is subscribing to states and is called each minute ...
* why some code is executed after 5 or 10 seconds via timeout?
* I'm also not really sure if this "inverall-call of main really works with compact mode that well ... move it in startAdapter
* Also incompatile with compact mode is that you can not stop that Interval. So please store the Interval-reference and implement on("unload") to clearInterval on adapter stop

jack-blackson:
@Apollon77 I adjusted the requested topics

GermanBluefox:
@Apollon77 Looks better. :)

GermanBluefox:
@apollon77 we should implement what possible from your notices as adapter-check rules.

- If news are in readme and in io-package and if the order is right.
- and so on

Apollon77:
Good idea

jack-blackson:
@GermanBluefox I changed the readme and io-package in version 1.0.4 which is already published now.

PR #457, opened 2019-07-28
add nina to repo and update stable for homeconnect and nightscout

TA2k:


Apollon77:
Please add "published" for the stable entries ... nightscout is already in latest?

TA2k:
I added.
Yes it is. 
Nina is not yet in latest. But test very well:
https://forum.iobroker.net/topic/23866/aufruf-nina-gefahrenmeldung-adapter

GermanBluefox:
@TA2k I just working on integration of nightscout (as server) into ioBroker.
Your nightscout adapter is just collects a couple of values via socket.io.
My nightscout will start the nightscout server and it will be possible to collect sugar measures via iot cloud.
Can I take your name for nightscout and you will rename your adapter into nightscout-client ?

TA2k:
Of course i will rename the adapter

GermanBluefox:
> Now before merging in stable?

Of course before.

TA2k:
ok nightscout is removed it is only nina and a update for homeconnect

PR #461, opened 2019-08-05
new adapter letrika_comgw

AWhiteKnight:
new try ;-)

### DIFF
--- a/sources-dist.json
+++ b/sources-dist.json
@@ -1986,5 +1986,12 @@
     "type": "hardware",
     "published": "2015-01-03T00:02:56.370Z",
     "versionDate": "2018-07-25T09:54:52.755Z"
+  },
+  "letrika_comgw": {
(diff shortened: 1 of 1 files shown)


Apollon77:
I had a shot look, here my thoughts:
* you should remoe .vscode from GitHub
* you can use https://translator.iobroker.in/ to translate news texts and words.js for all languages
* if you do not need "on-State-Changed" or "on-Objects-changed" then do not use it. better performance
* you use intervalls, this brings a risk because you also do massive "external communications". If the communication is running into timeouts or such then the next intervall will hard fire the next requests and so on ... Please make sure that this is not "DOSing" the device or make problems in the adapter. Ideally user setTimeout and set a new timeout after the logic was executed
* you are creating many object sat the beginning. Plese have in mind that all of this is executed asyncronous. This can cause errors like "stack size exceeded". It would be better to collect all the objects to be created in an array and then process these using callbacks or promises one afte rthe other...

Apollon77:
All notes above are changeable afterwards, so welcome into the repo.

PR #466, opened 2019-08-12
added meteoalarm

jack-blackson:


Apollon77:
Hi,

Here some review comments:
* in io-package please remove your adapter from "restartAdapters" because this is for "other" adapters then yours
* also messagebox should be disabled when not needed

Else: Why you implement this as "deamon"? Such adapters that really run more in longer intervals should be more implemented as "schedule" to only use RAM when running and else have it freed.

Could you consider rebuild as schedule?

jack-blackson:
@Apollon77 I changed it to schedule and also adjusted the parameters in io-package

Apollon77:
Now your Travis-CI testing went red. This is because you need to add "11" as allowed exit code to the testing ... https://github.com/ioBroker/testing#adapter-startup-unit-test 

```
    allowedExitCodes: [11],
```

jack-blackson:
@Apollon77 Done, travis is now green

PR #473, opened 2019-08-31
Update sources-dist.json for Nuki adapter

smaragdschlange:


Apollon77:
Latest updates do not need to be updated that way ... this is done automatically.

PR #475, opened 2019-09-01
Update nina and mercedesme

TA2k:


Apollon77:
important fixes or why stable update that fast after release? :-)

TA2k:
Yes Mercedes api is changed and blocks the adapter. And Nina was added missing urls without changing code

Apollon77:
thank you!

PR #490, opened 2019-09-27
Fix ZWave2 icon url

AlCalzone:
When I added ZWave2 with #479 I started a little experiment. The adapter is the first one to use an SVG icon. While that mostly works, it breaks in the adapter list:
grafik

The reason is that `https://raw.githubusercontent.com/AlCalzone/ioBroker.zwave2/master/admin/zwave2.svg` shows the SVG source code while `https://raw.githubusercontent.com/AlCalzone/ioBroker.zwave2/master/admin/zwave2.svg?sanitize=true` has the correct MIME type to display an image.

This PR fixes that (or rather starts another experiment to see **if the query string breaks any scripts**)

/cc @GermanBluefox @Apollon77

### DIFF
--- a/sources-dist.json
+++ b/sources-dist.json
@@ -1454,7 +1454,7 @@
   },
   "zwave2": {
     "meta": "https://raw.githubusercontent.com/AlCalzone/ioBroker.zwave2/master/io-package.json",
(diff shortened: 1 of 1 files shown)


Apollon77:
Merge postponed till we start a 2.1 branch or 2.0 goes into an own branch.

PR #505, opened 2019-10-07
release vedirect & Tado to latest

DutchmanNL:
- release vedirect
- npm package ready this evening
- testing OK
- JS 2.0 compatible

### DIFF
--- a/sources-dist.json
+++ b/sources-dist.json
@@ -1247,6 +1247,11 @@
     "icon": "https://raw.githubusercontent.com/hometm/ioBroker.vcard/master/admin/vcard.png",
     "type": "misc-data"
   },
+  "vedirect": {
+    "meta": "https://raw.githubusercontent.com/iobroker-community-adapters.vedirect/master/io-package.json",
+    "icon": "https://raw.githubusercontent.com/iobroker-community-adapters.vedirect/master/admin/Vedirect_Adapterimage.png.png",
+    "type": "energy"
+  },
   "viessmann": {
     "meta": "https://raw.githubusercontent.com/misanorot/ioBroker.viessmann/master/io-package.json",
     "icon": "https://raw.githubusercontent.com/misanorot/ioBroker.viessmann/master/admin/viessmann.png",
--- a/sources-dist.json
+++ b/sources-dist.json
@@ -1172,6 +1172,11 @@
     "icon": "https://raw.githubusercontent.com/frankjoke/ioBroker.systeminfo/master/admin/systeminfo.png",
     "type": "misc-data"
   },
+  "tado": {
+    "meta": "https://raw.githubusercontent.com/iobroker-community-adapters/ioBroker.tado/master/io-package.json",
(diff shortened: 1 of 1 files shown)


sources-dist.json
```diff
@@ -1247,6 +1247,11 @@
     "icon": "https://raw.githubusercontent.com/hometm/ioBroker.vcard/master/admin/vcard.png",
     "type": "misc-data"
   },
+  "vedirect": {
+    "meta": "https://raw.githubusercontent.com/hometm/iobroker-community-adapters.vedirect/master/io-package.json",
```
ldittmar81: Wrong link

DutchmanNL: wtf happend, solved

DutchmanNL: and another typo fixed

GermanBluefox:
image

GermanBluefox:
And README must be cleared

DutchmanNL:
fixed

Apollon77:
Now we stay left with:

[E140] width and height of logo are not equal

as error

Apollon77:
Some review comments:
* Blocker: Do not use setObject because this would overwrite history settings! Please use setObjectNotExists
* All data with "role: 'state'" feels not optimal. Can you not define better roles for certain states?
* I think some serialpoert error management (listening to error events) would be helpfull, else adapter will crash when an error happens here
* setting connection state to true on EACH received data package is not ideal. Please more do ion "On(connect)" or such
* your "ID to text mappings" could be done with better performance as an ddefined object where you simply use the ID as key to lookup.
* in unload you also have code for TCP which is never initialized in ready or elsewhere
* you never set info.connection to false ...

DutchmanNL:
remarks fixed, can you have another look please :)

DutchmanNL:
added release. of Tado to this pull request

Apollon77:
Will have tonight

DutchmanNL:
All fixed for both adapters can you please have another look 🙂

Apollon77:
Tonight if 2.0.34 works out as expected ;-)

Apollon77:
Comments:
vedirect: 
* you still do not reset "info.connection" on serial disconnect, also errors are only logged, so on serial disconnect no reties are done and so not auto-fix try
* serialPort variable is wrong ... port is correct or ?! ... this can not work ... blocker?!

Tado:
* logging in warn?
* https://github.com/iobroker-community-adapters/ioBroker.tado/blob/master/main.js#L175 ?? :-) ... system.adapter. ...

Rest ok

DutchmanNL:
> * you still do not reset "info.connection" on serial disconnect, also errors are only logged, so on serial disconnect no reties are done and so not auto-fix try
> * serialPort variable is wrong ... port is correct or ?! ... this can not work ... blocker?!

Fixed second part, retry is a good suggestion will look into that today


> * logging in warn?
> * https://github.com/iobroker-community-adapters/ioBroker.tado/blob/master/main.js#L175 ?? :-) ... system.adapter. ...

- about logging currently yes, will be changed before moving to stable 
- do you have a suggestion for the second one? I disable the adapter here if credentials are not provided

PR #510, opened 2019-10-15
change nuki2 Github branch and add nuki-extended to latest

Zefau:
**nuki-extended shall replace nuki2 in near future.**
____
**FEATURES**
- Support for Nuki Smartlock and Nuki Opener
- Support for both Nuki Bridge API and Nuki Web API
- Support for hashed token on hardware bridges
- Fallback to Nuki Web API in case applied actions on Nuki Bridge API fail, e.g. due to bridge error 503
- Retry in case applied actions on Nuki Bridge API fail (when Nuki Web API is not used)
- Option to regularly synchronise instead of using Bridge API callback (which may be delayed due to Hardware Bridge)
- Refreshing all states of Nuki Web API when callback is received via Nuki Bridge API
- Retrieve authorised users for Nuki Smartlock and Nuki Opener
- Retrieve the configuration for Nuki Smartlock and Nuki Opener
- Retrieve setup Nuki Notifications
- Web Interface that shows the recent events from your Nuki Smartlock and Nuki Opener

### DIFF
--- a/sources-dist-stable.json
+++ b/sources-dist-stable.json
@@ -1119,9 +1119,9 @@
     "versionDate": "2018-10-12T16:52:47.916Z"
   },
   "nuki2": {
-    "meta": "https://raw.githubusercontent.com/Zefau/ioBroker.nuki2/master/io-package.json",
-    "icon": "https://raw.githubusercontent.com/Zefau/ioBroker.nuki2/master/admin/nuki-logo.png",
-    "version": "1.0.4",
+    "meta": "https://raw.githubusercontent.com/Zefau/ioBroker.nuki-extended/nuki2/io-package.json",
+    "icon": "https://raw.githubusercontent.com/Zefau/ioBroker.nuki-extended/nuki2/admin/nuki-logo.png",
+    "version": "1.0.5",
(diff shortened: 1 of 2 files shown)


Zefau:
@Apollon77 sorry for pinging you. Could you please have a look at this?

I have changed the Github repository of the Nuki2 adapter (because it will be replaced by nuki-extended) and not sure if this causes errors with ioBroker if the path to the repository is wrong.

This pr corrects the URL to the repo.

Thanks!

Apollon77:
1. Please check and fix the cinflicts
2. Adaptername and reponame/npmname need to match! So "nuki2" with "iobroker-.nuki.extended" will not work!

So especially because of the second topic I see two options:
1 rename back and stay with nuki2 ... you can increase major version to show big changes. I do not see any problem (I also did alexa2 event if the featureset is way higher then alexa originally was ;-) )
2 Remove nuki2 (but with this leave all current users without updates!) and only add nuki-extended

Option 2 could be backed by info pupups in info Adapter, but in fact if the adapter is compatible there should be no need to rename

Please re-thing that!

PR #515, opened 2019-10-19
Update stable-versions for energymanager and stiebel-isg

unltdnetworx:
new confirmed stable-versions for energymanager and stiebel-isg


Apollon77:
As of now 0 users have installed this energymanager version and 1 for stiebel-isg. Lets wait some days ... or is it important fixes?

unltdnetworx:
Both versions have been installed and tested more than one time. Maybe the count is that low, because I  published it to npm right before the pull-request.
But it's okay to wait a few days.

PR #521, opened 2019-10-23
Tested compact mode for Chromecast

angelnu:


angelnu:
The test error does not seem to be related to the update but a "backend" not being available. How can I trigger a re-test?

PR #522, opened 2019-10-25
add zoneminder

MeisterTR:


Jey-Cee:
Readme, License, package, io-package and admin stuff looks OK.
 
* https://github.com/iobroker-community-adapters/ioBroker.zoneminder/blob/b3658b266e3d67906772af2725db85d9e0724526/main.js#L194 --> should be 'state', value is used for read only.

* Is line 368 - 392 in main.js obsolete?

* Can not find definition for proxmox in line 406 in main.js. 

Please Check the 3 points.

MeisterTR:
should be fixed now [https://github.com/iobroker-community-adapters/ioBroker.zoneminder/commit/24d6034f1ca9a8e3d9951d6e869e3d92240acde7](url)

Jey-Cee:
check, now it is  perfect.

Apollon77:
Here some more comments:
* if you do not need objectchange or such then please comment it out.
* using intervals for "external communication" has the risk that th einterval fires hard after the given time - independently if the request before was done and finished or not. This can end up in strange situations. Better are setTimeout that sets a timeout for next run at the end.
* (Blocker!) When I understand it correctly the objects opens up a websocket connection. SO you need to make sure to close that connection in "unload"! Else compact mode will break
* also the "connected" state is only set on connect and resetted on adapter start. So it will basically stay true most of the time. This also makes not that much sense or?

MeisterTR:
fixed in latest commit..

PR #526, opened 2019-10-29
Bosch Smart Home Bridge Adapter

holomekc:


Jey-Cee:
* io-package.json: line 65, 66-79, 83, 91 & 92 obsolet?
* is style.css, custom_m.html tab_m.html used, looks like they are from template and will not be needed
* no translations in words.js

@holomekc & @Apollon77 I am not sure is src and build needed or could this structure changed to main.js and lib, like the most of adapter do this? 

Note: Use setObjectNotExist instead setObject when ever possible.

holomekc:
Hi. Thx for the feedback!

@Jey-Cee I think I adjusted everything what you mentioned.

Jey-Cee:
check, now it looks good.

Apollon77:
@holomekc there is no news entry in io-package fpr the 0.0.13 ... thi is ok for now but should be fixed with next version please, else users can not gho back to this version if needed.

But found some other stuff:
* io-package needs to contain the default values for all your setting params (those that you can edit in admin) in native
* if objectchange and message is not used please do not implement it (performance wise)

Please fix at least the first before we add it to the repository. Thank you!

holomekc:
@Apollon77 thx for the feedback.

I created a new branch so that I do not do all the changes directly on master. And I created a pull request so that it is maybe easier for you to see changes:
https://github.com/holomekc/ioBroker.bshb/pull/5/commits/19615019b5e5899880e411e7d56fff9a7ed06d47

**Regarding default values for configuration:**
**host:** I cannot really guess the value. I added one but this is much likely wrong
**mac:** Same. I added the pattern but to be honest this is more annoying than an empty field because you first need to remove it
**identifier:** I kept it empty. It would destory the purpose of a unique identifier when i add there a generated uuid
**systemPassword:** I kept it empty
**certsPath:** I added the directory I use. But the user needs to create such a directory and adjust the access rights
**pairingDelay:** I think this is fine

To be hones I think the default values only make sense for the last two settings. Regaring empty values. I would remove them when you do not have a better idea for a default value here.

**Regarding news:**
I added the history. I hope this is ok. I actually do not get the connection between this field and the use-case that a user wants to downgrade the version? Is this linked?

I read this guide here:
https://github.com/ioBroker/ioBroker.docs/blob/master/docs/en/dev/adapterdev.md#structure-of-io-packagejson

I cannot find "news" listed there. If this is really a required field it should be mentioned there and should be part of the SCHEMA.md description.

Thank you so much again.

holomekc:
Hi. Should I adjust something else?

Apollon77:
Now that controller 2.0 is out I will do check tomorrow

Apollon77:
Great! Welcome in the repo

holomekc:
Thank you!

PR #545, opened 2019-11-17
Stockmarket Adapter

waoler:


Apollon77:
* please add Bluefox to npm owners
* please add GitHub repo description
* please add checks for response error handling, else the adapter will crash on an not succesdfull return

waoler:
Thanks for your answer. Bluefox is already added to the npm owners. Also the GitHub repo got already a description. 
What do you mean by the third todo? Isn´t it enough to return out of the onReady() function? Like this:
`this.log.error("No API Key set. Please edit your adapter settings and restart this adapter!");
return;`

Apollon77:
What I mean is that thre is no error handling on your http request. You collect data the the request, you do not verify the response code, you directly assume that the return is json and parse it without any try/catch or check or such ... If the communication would not be successfull your adapter would crash with an exception

waoler:
Yeah. youre right. I will fix that :)

//DONE. travis test is running

Apollon77:
Then please doone last thing: please sort in the entry in the repo file on the correct alphabetical place :-)

waoler:
Its ready to publish finally!

PR #547, opened 2019-11-19
add weishaupt wem adapter

TA2k:


Apollon77:
* Please check the results/comments from https://adapter-check.iobroker.in/
* the package.json seems weired ... can you please check that only relevant content is i. there?
* also io-package seems copied from an installation? "installedFrom" need to be removed at minimum. please check other fields
* On your request you also should verify the responsecode (200) and that body is set before starting to parse it.

Additional note:
* I personally would never use "setInterval" because it has the risk that if there are network issues or a request is hanging/times out or such the intervall will fire next request and next request and such. Better practive is to use setTimeout and set the new after one request has been finished. Else you may stack up requests
* if you do not need objectCHange then please also do not register the method :-)


Thanjk you for checking and adjusting

TA2k:
fixed

Apollon77:
welcome in the repo

PR #549, opened 2019-11-25
Add tesla-motors to repositories

dbweb-ch:
Test-Thread:
https://forum.iobroker.net/topic/26428/test-adapter-tesla-motors-v0-1-x
Released current version is v0.2.0, also on NPM.
I still get 2 errors on https://adapter-check.iobroker.in/, but I've added Bluefox on NPM and included travis testing, so this should be fine.

### DIFF
--- a/sources-dist.json
+++ b/sources-dist.json
@@ -1222,6 +1222,11 @@
     "icon": "https://raw.githubusercontent.com/ioBroker/ioBroker.terminal/master/admin/terminal.png",
     "type": "utility"
   },
+  "tesla-motors": {
+    "meta": "https://raw.githubusercontent.com/dbweb-ch/ioBroker.tesla-motors/master/io-package.json",
+    "icon": "https://raw.githubusercontent.com/dbweb-ch/ioBroker.tesla-motors/master/admin/tesla-motors.png",
+    "type": "hardware"
+  },
   "text2command": {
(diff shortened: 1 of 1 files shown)


Apollon77:
Here some review comments:
* please use https://translator.iobroker.in/ to translate all news and texts to all languages
* when you check the travis you see some unhandled Promise rejections that you might want to check and fix (e.g. https://travis-ci.com/dbweb-ch/ioBroker.tesla-motors/jobs/260147804#L771)
* please remove tab_m.html if you do not provide a tab :-)
* you set connection to false twice at the beginning. of the adapter code in onReady
* I personally do not recomment usage of "setInterval" for external communications because of there are timeouts or delays you might end up in "stacking" requests ... setTimeout at the end of an action is better (but as said, my personal experience)
* (needs fix!) your adapter supports compact mode! Plese also test with this - and you need to make sure that all Intervals/Timeouts/connections are cleanly closed on "onUnload" ... currently all setInterval are not ended!
* (needs fix!) Do not user setObject ! Always use setObjectNotExist or (to update existing objects) extendObject! When you really overwrite an object you may kill custom settings on the object 8e.g. from History adapters) which is not ideal for the users.
* (needs fix) native is an object and not an array
* (needs fix) The role "value.speed" is for wind speed ... not car speed :-) also value.dostance is more for weather usage then other usage. You should not use those roles for car meanings ;-)

dbweb-ch:
Hi,

Thanks for your review. This should all be fixed now.

PR #553, opened 2019-11-27
add velux kix 300 adapter

TA2k:


Apollon77:
Here some review comments:
* you may want to delete .vscode directory
* I personally would not recommend to use intervals ... better are timeouts that are set after action was finished
* (needs fixing) from whatever you copied/generated the adapter you have "allowCustoms" and custom_m.html in it, but do not use it. Please remove or use!

GermanBluefox:
@TA2k please close the found issues

TA2k:
issues are fixed

PR #579, opened 2020-01-09
Added swiss-weather-api Adapter

baerengraben:


baerengraben:
https://forum.iobroker.net/topic/28772/test-adapter-swiss-weather-api-v0-1-x

Apollon77:
Hi, here some review comments:
* Note: You may want to remove .vscode from GitHub
* Note: There seems to be an eror in your github actions: https://github.com/baerengraben/ioBroker.swiss-weather-api/runs/384741918
* Note: You might want to address: (node:6638) [DEP0005] DeprecationWarning: Buffer() is deprecated due to security and usability issues. Please use the Buffer.alloc(), Buffer.allocUnsafe(), or Buffer.from() methods instead. (see testing travis) --> https://github.com/baerengraben/ioBroker.swiss-weather-api/blob/master/main.js#L65 Buffer.from(data) should work instead
* Note: is schedule like https://github.com/baerengraben/ioBroker.swiss-weather-api/blob/master/io-package.json#L73 really ok? should it not be */30 ? ... ok maybe both works :-)
* Note: You should add a try-catch around https://github.com/baerengraben/ioBroker.swiss-weather-api/blob/master/main.js#L60 and other file operations, else whole adapter crashes on error
* Note: you might want to use path.join instead of "linux style paths" ;-) (they should work on windows too ... but)
* Note: you should add try-catch areound https://github.com/baerengraben/ioBroker.swiss-weather-api/blob/master/main.js#L97 to make sure to handle error/network cases! WHole error handling for the external communication is missing at the moment in all places!.
* Note: Please use setState when not also using await or such (e.g. https://github.com/baerengraben/ioBroker.swiss-weather-api/blob/master/main.js#L153)
* Note: All requets are fired at the same time ... hopefully this is ok
* Note: All setObjectNotExists - setState combination are probematic because both are asyncronous, so on first case when object is not existing a setState might be faster then the object create, meaning state is not set! Better is to do setState in callback of setObjectNotExists

And to your comment on https://github.com/baerengraben/ioBroker.swiss-weather-api/blob/master/main.js#L2227: Once you are dne with your work you need to stop the adapter. This is no "hack", because there are processes in the adapter core logic that reports values like memory usage, cpu usage, connectio/alive status and such and these data are reported by default every 15s ... so without calling "stop" in general this reporting is running as expected and so process still runs.

The "hack" is that you set it as 10s timeout ... normally you need to consider all your async (http requests, object creations and state set) parts and once they are all done and executed you do one call to stop! This should it be.
Currently with the 10s timeout it means that if communication needs longer because of reasons you might have added only parts of the data before you end the adapter by yourself. Also depending on the hardware the first adapte rrun can take time to create all objects and such ...
So please (also see above) consider the async behaviour of Javascript when coding!

In fact ther eis no "show stopper" from my point of view ... but you might want to revise things before I add it to the repo (especially stable) Please tell me

baerengraben:
Hi Apollon77
Many thanks for this detailed review report! As you can see this is my first JS-Adapter and I appreciate good feedbacks. :) 
I will try to improve the Adapter and will create a new release.

baerengraben:
Hi Apollon77

I did some corrections on the code:

> Hi, here some review comments:
> 
>     * Note: You may want to remove .vscode from GitHub
=> done
> 
>     * Note: There seems to be an eror in your github actions: https://github.com/baerengraben/ioBroker.swiss-weather-api/runs/384741918
=> Yes. I removed the workflow.
> 
>     * Note: You might want to address: (node:6638) [DEP0005] DeprecationWarning: Buffer() is deprecated due to security and usability issues. Please use the Buffer.alloc(), Buffer.allocUnsafe(), or Buffer.from() methods instead. (see testing travis) --> https://github.com/baerengraben/ioBroker.swiss-weather-api/blob/master/main.js#L65 Buffer.from(data) should work instead
=> done
> 
>     * Note: is schedule like https://github.com/baerengraben/ioBroker.swiss-weather-api/blob/master/io-package.json#L73 really ok? should it not be */30 ? ... ok maybe both works :-)
=> done
> 
>     * Note: You should add a try-catch around https://github.com/baerengraben/ioBroker.swiss-weather-api/blob/master/main.js#L60 and other file operations, else whole adapter crashes on error
=> done
> 
>     * Note: you might want to use path.join instead of "linux style paths" ;-) (they should work on windows too ... but)
=> done
> 
>     * Note: you should add try-catch areound https://github.com/baerengraben/ioBroker.swiss-weather-api/blob/master/main.js#L97 to make sure to handle error/network cases! WHole error handling for the external communication is missing at the moment in all places!.
=> Hm... I do "req.on('error', function(e) { ..." because I read that http.get() cannot be handled with a usual try-catch. is this not ok like this?
> 
>     * Note: Please use setState when not also using await or such (e.g. https://github.com/baerengraben/ioBroker.swiss-weather-api/blob/master/main.js#L153)
=> done
> 
>     * Note: All requets are fired at the same time ... hopefully this is ok
=> The requests are independent from each other. so this is ok.
> 
>     * Note: All setObjectNotExists - setState combination are probematic because both are asyncronous, so on first case when object is not existing a setState might be faster then the object create, meaning state is not set! Better is to do setState in callback of setObjectNotExists
=> done
> 
> And to your comment on https://github.com/baerengraben/ioBroker.swiss-weather-api/blob/master/main.js#L2227: Once you are dne with your work you need to stop the adapter. This is no "hack", because there are processes in the adapter core logic that reports values like memory usage, cpu usage, connectio/alive status and such and these data are reported by default every 15s ... so without calling "stop" in general this reporting is running as expected and so process still runs.
> 
> The "hack" is that you set it as 10s timeout ... normally you need to consider all your async (http requests, object creations and state set) parts and once they are all done and executed you do one call to stop! This should it be.
> Currently with the 10s timeout it means that if communication needs longer because of reasons you might have added only parts of the data before you end the adapter by yourself. Also depending on the hardware the first adapte rrun can take time to create all objects and such ...
> So please (also see above) consider the async behaviour of Javascript when coding!

=> Thanks for this explanations! :) I tried to correct this by using "await getWeatherInfos()". But I'm not sure if this is realy ok like this. So the http-requests are awaited before the adapter is stopping..? My tests seems ok...But somehow I do have a bad feeling, that now it could be possible that the adapter is stopping before the http-requests have been ended. ?
> 
> In fact ther eis no "show stopper" from my point of view ... but you might want to revise things before I add it to the repo (especially stable) Please tell me

=> Can you please recheck the code? If it's ok I will do a new release. Thank you very much!

Apollon77:
Thank you very much!
Yea ... the async nature of Javascript is sometimes hard to understand in the beginning ... but If you did once it gets easier ;-)

Apollon77:
> * Note: you should add try-catch areound https://github.com/baerengraben/ioBroker.swiss-weather-api/blob/master/main.js#L97 to make sure to handle error/network cases! WHole error handling for the external communication is missing at the moment in all places!.

Yes for network stuff your "On(error) works fine. But if the JSON response is e.g. n html error page and you do "JSON parse" on it it will throw an exception ... my idea was to add try catch there tio handle JSON parse errors more gracefully

Apollon77:
Welcome in the repo :-)

PR #584, opened 2020-01-17
- added plenticore adapter

StrathCole:
ioBroker.plenticore Adapter for KOSTAL Plenticore Plus

### DIFF
--- a/sources-dist.json
+++ b/sources-dist.json
@@ -957,6 +957,11 @@
     "icon": "https://raw.githubusercontent.com/iobroker-community-adapters/ioBroker.places/master/admin/places.png",
     "type": "geoposition"
   },
+  "plenticore": {
+    "meta": "https://raw.githubusercontent.com/StrathCole/ioBroker.plenticore/master/io-package.json",
+    "icon": "https://raw.githubusercontent.com/StrathCole/ioBroker.plenticore/master/admin/plenticore.png",
+    "type": "energy"
+  },
   "plex": {
     "meta": "https://raw.githubusercontent.com/Zefau/ioBroker.plex/master/io-package.json",
     "icon": "https://raw.githubusercontent.com/Zefau/ioBroker.plex/master/admin/plex.jpg",


GermanBluefox:
Can you fix this?
image

StrathCole:
Done in 1.1.0 just pushed to git and npm

Apollon77:
Here are my review comments:
* Needs fix please: add all editable fields to io-package native with their default values
* Note: Do you really want to write in the adapters node_module directory? This is deleted on every update of the adapter ... Better use iobroker-data/instance directory? (https://github.com/StrathCole/ioBroker.plenticore/blob/master/main.js#L10)
* Note: How long the code in "Unload" normally needs? 500ms is the default time limit before adapter is killed hard. If it needs more time set io-package.common-stopTimeout with a higher value
* Note: if you do not need on-message please do not implement it 
* Note: The code https://github.com/StrathCole/ioBroker.plenticore/blob/master/main.js#L1149 is executed arallel to the other code because async, so it could be that the variable is set to false later ...
* Note: Do I see it correct that in many places createOrSetState is used also during runtime and not initial only? so for each "setState" you read the object too? This costs performance! Please re-think it. Alternatively you could use the oject cache from adapter class ...
* Note: The two api calls are done kind of in parallel at https://github.com/StrathCole/ioBroker.plenticore/blob/master/main.js#L1632 and other places. If you rely on that the first was finished before doing the second you should chain them in the callback!
* Needs fix: you do not store the timoeut here and do not end it in unload: https://github.com/StrathCole/ioBroker.plenticore/blob/master/main.js#L1666
* Needs fix: please use adapter.terminate if available instead of process.exit, else compact mode gets killed. (https://github.com/StrathCole/ioBroker.plenticore/blob/master/main.js#L1818)
* Note: The code in https://github.com/StrathCole/ioBroker.plenticore/blob/master/main.js#L1664-L1667 culd mean that you start polling twice (e.g. when I have set 6000) ;-)
* Note: Me personally do not like setInterval when doing external communications. Reason: When a timeout or such happens the next intervall fires as requested also when the run before is still running ... So requests can pile up and strange things happen. I think better is to set a timout for next run at the end of last run.
* Needs fixing: Please add intgration testing that starts adapter - or is there a reason not to have it?

StrathCole:
Hopefully everything was addressed now.

PR #585, opened 2020-01-19
Added somfy tahoma adapter

StrathCole:
Adapter for controlling Somfy Tahoma

### DIFF
--- a/sources-dist.json
+++ b/sources-dist.json
@@ -1222,6 +1222,11 @@
     "icon": "https://raw.githubusercontent.com/iobroker-community-adapters/ioBroker.tado/master/admin/tado.png",
     "type": "climate-control"
   },
+  "tahoma": {
+    "meta": "https://raw.githubusercontent.com/StrathCole/ioBroker.tahoma/master/io-package.json",
+    "icon": "https://raw.githubusercontent.com/StrathCole/ioBroker.tahoma/master/admin/tahoma.png",
+    "type": "iot-systems"
+  },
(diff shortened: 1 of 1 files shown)


GermanBluefox:
Can you fix this?
image

StrathCole:
Oh of course. I was told those files should only remain local and not be included in the git repo.

StrathCole:
Done that and pushed to git and npm 0.2.6

Apollon77:
Ok, then also my review comments:
* Needs Fix please: username field missing in io.package - native
* Note: pollInterval is default value 20 ... really? In code you re set it to 5s ... maybe better to adjust default value too
* Needs Fix please: testing only checks packages and install ... you should use the integration test too please - or is there a reason against it?
* Note:you do not need tio implement on-message if you do not use it.
* Note: processStateChangeAck should never be needed because ack=true should only come from adapter and best practive it to ignore these changes
* Note: in timeout cases the variable response in https://github.com/StrathCole/ioBroker.tahoma/blob/master/lib/tahoma.js#L170 might be no object ... mabe check this too (i had crashed´s because of this)
* Needs Fix please: you use timeouts in your "Lib" that are not stoppable on unload and can lead to problems in compact mode: https://github.com/StrathCole/ioBroker.tahoma/blob/master/lib/tahoma.js#L189 Please make sure to handle that  (also applyTimer)

Rest looks good

StrathCole:
All of the mentioned review commends should have been addressed now. Current is 0.3.0.

PR #586, opened 2020-01-20
ioBroker.valloxmv 1.0.3

hacki11:
Bugfix


GermanBluefox:
Can you fix this?
image

hacki11:
fixed all three issues

Apollon77:
Hi,

I have only one review comment :-)
* Note: You might want to check that config is set before having the adapter crash with Exceptions :-) https://travis-ci.com/hacki11/ioBroker.valloxmv/jobs/277729462#L528

hacki11:
Thanks for the review @Apollon77,

currently the adapter is not crashing, just logging the error.
on connection errors (invalid url or unreachable device) the adapters health is changed and retries connecting until the device is back online or the url is configured correctly.

what could be added is a validation, if the hostname/ip and port is configured at all.
if it is not set the adapter could be shut down immediately

PR #591, opened 2020-01-22
add mydlink adadpter to latest

Garfonso:
There is a thread on the forum, but it does not have any posts for a long time now...
Maybe if it is in latest, some people will find it and test.

For me it works fine since over a month now. 👍

### DIFF
--- a/sources-dist.json
+++ b/sources-dist.json
@@ -807,6 +807,11 @@
     "icon": "https://raw.githubusercontent.com/foxthefox/ioBroker.musiccast/master/admin/musiccast.png",
     "type": "multimedia"
   },
+  "mydlink": {
+    "meta": "https://raw.githubusercontent.com/iobroker-community-adapters/ioBroker.mydlink/master/io-package.json",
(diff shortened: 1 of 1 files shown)


GermanBluefox:
@Apollon77 LGTM

Apollon77:
Hi,

here my review comments, just two tiny things to be fixed. Thank you!

* Please also enable travis-ci testing additional to GitHub actions
* Please adjust the "connectionType" field to the newst format (sorry we changed afterwards :-( ) --> https://github.com/ioBroker/ioBroker.repositories#connection-types there are now two fields
* When I understand the code correctly it should be no problem that you use setObjectAsync . Please make sure to really check that objects do not exists before, else it will remove custom details in js-controller < 2.2.x ( https://github.com/iobroker-community-adapters/ioBroker.mydlink/blob/master/main.js#L67) setObjectNotExists should be used more safe

But else looking really great.

Garfonso:
ok, thanks for the review. Fixed the issues and also changed to setObjectNotExistsAsync. In one case I had to use extendObject which will be  called during startup for every motion detector (i.e. not very often). 
I did some testing here and all CI tests did run through.

Apollon77:
ExtendObject is also fine ;-) I Check later.

PR #598, opened 2020-01-31
new adapter for openligadb soccer sport data

oweitman:
https://forum.iobroker.net/topic/29506/test-adapter-openligadb-v0-0-x

### DIFF
--- a/sources-dist.json
+++ b/sources-dist.json
@@ -892,6 +892,11 @@
     "icon": "https://raw.githubusercontent.com/iobroker-community-adapters/ioBroker.openhab/master/admin/openhab.png",
     "type": "iot-systems"
   },
+  "openligadb": {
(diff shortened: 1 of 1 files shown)


Apollon77:
Hi,

looks godd, but I have some review comments:
* Needs fix: You need to make sure to e able to clear any timeout and interval set by the adapter and do that on-unload. (https://github.com/oweitman/ioBroker.openligadb/blob/master/openligadb.js#L40 does not exists or I'm blind ;-) ... and it should clear all set timeouts) Else compact mode will break

Rest looks fine!

oweitman:
new version v0.0.7 included now the removal of the timeout. this is now reported as debug message in log.

Apollon77:
Timeouts and intervals are just some of the topics that needs to be done in unload. You need to clean up all resources - so also servers and serialports needs to be closed and such.

oweitman:
thanks

PR #600, opened 2020-02-03
Added MieleCloudService

Grizzelbee:
Added MieleCloudService (formerly known as MieleAtHome) to get it into the repo.

### DIFF
--- a/sources-dist.json
+++ b/sources-dist.json
@@ -730,7 +730,7 @@
   "mielecloudservice": {
     "meta": "https://raw.githubusercontent.com/Grizzelbee/ioBroker.mielecloudservice/master/io-package.json",
     "icon": "https://raw.githubusercontent.com/Grizzelbee/ioBroker.mielecloudservice/master/admin/mielecloudservice.png",
-    "type": "hardware"
+    "type": "household"
   },
   "mihome": {
     "meta": "https://raw.githubusercontent.com/ioBroker/ioBroker.mihome/master/io-package.json",
--- a/sources-dist.json
+++ b/sources-dist.json
@@ -727,6 +727,11 @@
     "icon": "https://raw.githubusercontent.com/soef/ioBroker.miele/master/admin/xmiele.png",
     "type": "household"
   },
+  "mielecloudservice": {
+    "meta": "https://raw.githubusercontent.com/Grizzelbee/ioBroker.mielecloudservice/master/io-package.json",
+    "icon": "https://raw.githubusercontent.com/Grizzelbee/ioBroker.mielecloudservice/master/admin/mielecloudservice.png",
+    "type": "hardware"
+  },
   "mihome": {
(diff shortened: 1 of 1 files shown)


Apollon77:
Please check your adapter at https://adapter-check.iobroker.in/ and fix the things mentioned there. Also please read over https://github.com/ioBroker/ioBroker.repositories#development-and-coding-best-practices and checks what may be relevant (e.g. storing and clearing timeouts).

Additionally what is obj.isNotNull ?? https://github.com/Grizzelbee/ioBroker.mielecloudservice/blob/master/main.js#L45 ?!

Grizzelbee:
All checked and fixed (as far as I can see). But the AdapterChecker says that travis-ci is still failing - what is not the case. It passes. Maybe it's a travis-ci.com/.org thing.

obj.isNotNull was a leftover from a former contributor, I have overseen. Removed it.

Grizzelbee:
After some Code impovements have been done: May I ASK for another revirew?

Apollon77:
Looks ok now. Welcome in the repo.

One Personal side note: Using "node-schedule" just to do something where setInterval or better setTimeout is already capable of feeld unneeded (also I have no idea what happens if I put 120s as Interval ... will node_scheule handle a*/120 correctly? ;-)
In my works it should be enough to use setInterval or much better setTimeout to trigger the next execution and set a new timeout when this one is finished.

The downside of schedule/intervals is that they fire a new call regardles of if the old one has finished. This can lead to DOS like behaviour especially with external/cloud communications because if there are network errors or such that add latency or such could mean that requests pile up and then all fire at the same timepoint ...
Maybe rethink this approach

Apollon77:
PS: one additional downside of "scheduler": This about you developed an successfull adapter :-) And most people will most likely not change the default polling interval.

Using the scheduler means that all adapters our there always fire at the same second "*/interval" ... this already could become kind of a DOS behaviour with the external cloud service. I would like see you change it soon :-)

Apollon77:
PPS: Makes it really sense that your repo is fork of mieleathome? You should contact GitHub to decouple it ... also because one that has forked mieleathome can not create a fork of your repo

Grizzelbee:
Hallo Ingo,

vielen dank für deine Anmerkungen und Anregungen. Ich weiß das wirklich zu schätzen. 
Das Repo habe ich vom ursprünglichen Fork trennen lassen.  ;)
Das mit dem node-Schedule war mir auch schon aufgefallen, ich hatte aber keine gute Alternative am Start. Ich schwenke in der Version 1.0.5 gerade auf setTimeout - danke für den Tip in diese Richtung.

Apollon77:
Sag bescheid wenn done dann reviewe ich nochmal. Repo warten bis dahin ist ok für dich?

Grizzelbee:
Ist erledigt. V1.0.5 ist online und ich denke sogar ein Kandidat um in ein, zwei Wochen ins Stable zu kommen.

PR #611, opened 2020-02-12
release web-speedy

DutchmanNL:


Apollon77:
* Needs fix: Please also exscute integration tests on travis
* index_m is current? feels like dummy files?!
* makes it really sens to always do require('speedtest-net')()... ?! This always creates a new function instance, or ?! That feels very strange
* https://github.com/iobroker-community-adapters/ioBroker.web-speedy/blob/master/main.js#L177 ... admin will most like not catch up new state Lists without an object reload
* timeouts not cleared ;-) gggrrrrrrrr ...
* When such a test is running )considering it needs some time to measure). Is there any way to cancel this? If yes please also do in unload. Else is there any maximum time because elsoe you might ned to consider it as stopTimeout in io-package (normally the controller expect the proecess to be finished after 500ms ... is this really always the case here?
* Please make sure to check compact mode usage. Especially all relevant stop/retsart cases

DutchmanNL:
> * Needs fix: Please also exscute integration tests on travis

added,  please note its not mandatory in adapter creation templates anymore

> * index_m is current? feels like dummy files?!

removed html_m as I don't use a. config page

> * makes it really sens to always do require('speedtest-net')()... ?! This always creates a new function instance, or ?! That feels very strange

No, it does not fixed :)

> * https://github.com/iobroker-community-adapters/ioBroker.web-speedy/blob/master/main.js#L177 ... admin will most like not catch up new state Lists without an object reload

Hmm, I don't see another way handling this creating states will result in same behavior.
If you have any other suggestions than accepting the reload I am open for it :)

> * timeouts not cleared ;-) gggrrrrrrrr ...

Shame one me my friend, fixed !

> * When such a test is running )considering it needs some time to measure). Is there any way to cancel this? If yes please also do in unload. Else is there any maximum time because elsoe you might ned to consider it as stopTimeout in io-package (normally the controller expect the process to be finished after 500ms ... is this really always the case here?

Impossible to detect so hard kill is the only option.

> * Please make sure to check compact mode usage. Especially all relevant stop/retsart cases

compact mode disabled, this adapter us much to much memory intensive to run in compact mode on low hardware so should be avoided.....

PR #659, opened 2020-03-18
updates for my repositories

foxthefox:


Apollon77:
RE-CHECK!

foxthefox:
I am sorry, but I have to ask. I see the error on Travis but I have no idea what should be changed, to not cause the failure.
`1) Test Repository

       Test all Packages in latest are loadable via http and name is equal to io-package.json are :

     Error: the string "Error occurred, see console output" was thrown, throw an Error :)

      at runMicrotasks (<anonymous>)

      at processTicksAndRejections (internal/process/task_queues.js:97:5)`

Any help would be appreciated, thanks

Apollon77:
RE-CHECK!

PR #673, opened 2020-03-24
Update nuki-extended in stable

Zefau:


Apollon77:
RE-CHECK!

GermanBluefox:
## Automated adapter checker

### [ioBroker.nuki-extended](https://github.com/Zefau/ioBroker.nuki-extended)
[Downloads](https://www.npmjs.com/package/iobroker.nuki-extended) Number of Installations (latest) Number of Installations (stable)
[NPM](https://nodei.co/npm/iobroker.nuki-extended/)

- [ ] :heavy_exclamation_mark: [E144] common.installedFrom field found in [io-package.json](https://github.com/Zefau/ioBroker.nuki-extended/blob/master/io-package.json). Must be removed.


Apollon77:
Please remove the mentioned field and publish a new version

Zefau:
Does it make any difference? The version is already published to npm like that. I can remove it, but it does not change anything, does it?

Apollon77:
It should be overridden by installation in js.controller 2.2+. But I'm not really sure. If not then it would break the "remember installation localtion" feature and pot. install wrong versions for users.

So yes it is published that way - which does not make it better :-)

Zefau:
If I have some spare time, I will develop a Github Action to remove this line on a commit or pull request. I'm releasing to npm via Github Actions as well, so this should fix it sustainably.

For now, I removed the line manually. Please merge.

Apollon77:
... only because last version was online that way too ...

PR #680, opened 2020-03-29
- added darksky adapter to repository

StrathCole:


Apollon77:
RE-CHECK!

GermanBluefox:
## Automated adapter checker

### [ioBroker.darksky](https://github.com/StrathCole/ioBroker.darksky)
[Downloads](https://www.npmjs.com/package/iobroker.darksky) Number of Installations (latest) 
[NPM](https://nodei.co/npm/iobroker.darksky/)

:thumbsup: No errors found
- [ ] :eyes: [W400] Cannot find "darksky" in latest repository
- [ ] :eyes: [W505] setTimeout found in "main.js", but no clearTimeout detected


StrathCole:
The mentioned setTimeout is the following:
```
setTimeout(function() {
			adapter.stop();
		}, 5000);
```
Does this really need a clearTimeout for a schedule adapter?

Apollon77:
Some review comments:
* Please check your roles, e.g. value.probability is not defined as a role - check https://github.com/ioBroker/ioBroker.docs/blob/master/docs/en/dev/stateroles.md
* to your question regarding the "killswitch timeout": In fact this timeout is only used if the code is written as you did and then I find 5s very low for the number of tates created by the adapter. Ideally you would make sure that you respect all the state creations and execute on "stop" command after all were executed. This could by done by Promises or also by counting state creations and number of executed callbacks. With this static 5s you are at a very high risk that you kill the adapter before all states are created on lower CPU systems ... But in fact yes: you can ignore the clearTimeout warning in your case.

StrathCole:
> Please check your roles, e.g. value.probability is not defined as a role

Yes, I know. The problem is that I do not find matching roles for many states. E. g. there are states for: `value.precipitation.day.forecast.0` (today) and `value.precipitation.day.forecast.1` (tomorrow) but what about the current precipitation chance and the hourly forecasts?

Apollon77:
I know. Please stick to the existing role definitions and you're welcome to open an issue at js-controller repo to expose your ideas for additional new roles.

And also please understand the meaning of roles: Roles are there that visualizations and editor "understand" a bit better about the meaning of a state value to display it correctly, e.g. to automatically choose an icon or such. So it needs to be a balance what is needed ... but if it is too many then visus can not really adjust to all of am ;-(

StrathCole:
Sadly DarkSky has been integrated into Apple and the sign-up has been canceled. The api will stop working end of 2021 so the adapter no longer makes sense for the repo.

PR #681, opened 2020-03-29
suggest jarvis for latest repository

Zefau:


Apollon77:
RE-CHECK!

GermanBluefox:
## Automated adapter checker

### [ioBroker.jarvis](https://github.com/Zefau/ioBroker.jarvis)
[Downloads](https://www.npmjs.com/package/iobroker.jarvis) Number of Installations (latest) 
[NPM](https://nodei.co/npm/iobroker.jarvis/)

- [ ] :heavy_exclamation_mark: [E300] Not found on [travis](https://travis-ci.com/). Please setup [travis](https://travis-ci.com/)
- [ ] :eyes: [W400] Cannot find "jarvis" in latest repository


Zefau:
E300: available on travis, see https://travis-ci.com/github/Zefau/ioBroker.jarvis
W400: is this pr

Apollon77:
Hi,

some review comments:
* please add your nodejs limitation also to package.json as "engine" property then npm and also js-controller will better take care to not allow installation at all, so you would not need to take care in your code.
* basically constructs like https://github.com/Zefau/ioBroker.jarvis/blob/master/lib/library.js#L122 might break compact mode because these timeouts can/will not be cleared and if unload is called in between and your callback is then coming later this might cause big trouble (e.g. because database connection is already done ... In this case it seems to not be used anywhere ... ?
* we are currently in the process to upgrade socket.io to 2.3.0 and new web version is alreayd in latest. Please update your included socket.io too (https://github.com/Zefau/ioBroker.jarvis/blob/master/www/vendor/socket.io.js) and make a dependency to web 3.0. I think it makes no sense to now add a new adapter with that old socketio.

GermanBluefox:
@Zefau  any progress here? Socket.io is now 2.3.0 everywhere.

Apollon77:
@Zefau ?

PR #754, opened 2020-05-10
Add time-switch to latest

walli545:
I created a new adapter that allows the user to create time scheduled switching actions using a vis widget.
Adapter repo: https://github.com/walli545/ioBroker.time-switch
Forum thread: https://forum.iobroker.net/topic/31479/test-adapter-time-switch-v1-0-x-github-latest/
Adapter request: https://github.com/ioBroker/AdapterRequests/issues/219

### DIFF
--- a/sources-dist.json
+++ b/sources-dist.json
@@ -1377,6 +1377,11 @@
     "icon": "https://raw.githubusercontent.com/ioBroker/ioBroker.tileboard/master/admin/tileboard.png",
     "type": "visualization"
   },
+  "time-switch": {
+    "meta": "https://raw.githubusercontent.com/walli545/ioBroker.time-switch/master/io-package.json",
+    "icon": "https://raw.githubusercontent.com/walli545/ioBroker.time-switch/master/admin/time-switch.png",
+    "type": "time-and-date"
+  },
   "tinker": {
     "meta": "https://raw.githubusercontent.com/simatec/ioBroker.tinker/master/io-package.json",
     "icon": "https://raw.githubusercontent.com/simatec/ioBroker.tinker/master/admin/tinker.png",


Apollon77:
RE-CHECK!

GermanBluefox:
## Automated adapter checker

### [ioBroker.time-switch](https://github.com/walli545/ioBroker.time-switch)
[Downloads](https://www.npmjs.com/package/iobroker.time-switch) Number of Installations (latest) 
[NPM](https://nodei.co/npm/iobroker.time-switch/)

:thumbsup: No errors found
- [ ] :eyes: [W400] Cannot find "time-switch" in latest repository


Apollon77:
Here some review comments:
* Please do not subscribe to obejcts changes if you do not need it
* on State Change most stuff is not needed when "ack" concept is used. normally you ignore all state changes with ack=true (these should come from the adater itself) and only react on ack= false) currently this ack handling is missing ... is this intended?

walli545:
> Here some review comments:
> 
> * Please do not subscribe to obejcts changes if you do not need it
> * on State Change most stuff is not needed when "ack" concept is used. normally you ignore all state changes with ack=true (these should come from the adater itself) and only react on ack= false) currently this ack handling is missing ... is this intended?

I will look into it and implement the suggested changes 👍

walli545:
@Apollon77 I implemented the suggested changes with release 2.1.0

GermanBluefox:
There is a subscription on object: https://github.com/walli545/ioBroker.time-switch/blob/master/src/main.ts#L97
But I could not find any handlers for that. What is the logic behind it?

DutchmanNL:
@walli545 any update?

walli545:
@GermanBluefox @DutchmanNL 
Sorry for my late reply but I didn't have much time to work on the adapter due to personal reasons. I already removed that subscription in dev branch and the change will come with the next release.

DutchmanNL:
@walli545 don't worry private life is always more Important!

Please give a sign when we can handle another review.

Until then all best for whatever is currently in your mind and hopefully a positive outcome

Apollon77:
@walli545  any news here?

walli545:
> @walli545 any news here?

I gave out a pre release of version 2.2.0 for testing today and will comment here if i make the release including the changes.

walli545:
@Apollon77 @DutchmanNL I released version 2.2.0 today so I think you can merge this now.

Apollon77:
Hello,
sorry for the long waiting time, but I noticed some small things to check

* please adjust the js.controller dependency in io-package to >=2.0.0 because you use adapter-core 2.4.0
* please add all configurable entries with default values in io-package native key.

AFter that we should be ready!
Thank you for your patience

walli545:
@Apollon77 I just release 2.2.1 which changes the dependency to js-controller >= 2.0.0

As to the entries in the native key, there is only one setting which is `shedules.onOff` which is an array that includes the created schedules. It is filled by using the admin UI and no other default values exist.

PR #761, opened 2020-05-13
add minuvis

SVallant:
added ioBroker.minuvis Adapter to latest


SVallant:
did work with account svallant

PR #783, opened 2020-05-24
E-Control-at-fuel Adapter

xXBJXx:
[Forumbeitrag](https://forum.iobroker.net/topic/33033/e-control-at-fuel-kraftstoffpreise-f%C3%BCr-%C3%B6sterreich)

### DIFF
--- a/sources-dist.json
+++ b/sources-dist.json
@@ -1686,6 +1686,11 @@
     "icon": "https://raw.githubusercontent.com/iobroker-community-adapters/ioBroker.wled/master/admin/wled.png",
     "type": "lighting"
   },
+   "wallpanel": {
+    "meta": "https://raw.githubusercontent.com/xXBJXx/ioBroker.wallpanel/master/io-package.json",
+    "icon": "https://raw.githubusercontent.com/xXBJXx/ioBroker.wallpanel/master/admin/wallpanel.png",
+    "type": "general"
+  },
   "wm-bus": {
     "meta": "https://raw.githubusercontent.com/soef/ioBroker.wm-bus/master/io-package.json",
     "icon": "https://raw.githubusercontent.com/soef/ioBroker.wm-bus/master/admin/wm-bus.png",
--- a/sources-dist.json
+++ b/sources-dist.json
@@ -1686,6 +1686,11 @@
     "icon": "https://raw.githubusercontent.com/iobroker-community-adapters/ioBroker.wled/master/admin/wled.png",
     "type": "lighting"
   },
+  "wallpanel": {	
+    "meta": "https://raw.githubusercontent.com/xXBJXx/ioBroker.wallpanel/master/io-package.json",	
+    "icon": "https://raw.githubusercontent.com/xXBJXx/ioBroker.wallpanel/master/admin/wallpanel.png",	
+    "type": "general"
+  },
   "wm-bus": {
     "meta": "https://raw.githubusercontent.com/soef/ioBroker.wm-bus/master/io-package.json",
     "icon": "https://raw.githubusercontent.com/soef/ioBroker.wm-bus/master/admin/wm-bus.png",
--- a/sources-dist.json
+++ b/sources-dist.json
@@ -239,6 +239,11 @@
     "icon": "https://raw.githubusercontent.com/mrbungle64/ioBroker.ecovacs-deebot/master/admin/ecovacs-deebot.png",
     "type": "household"
   },
+  "e-control-at-fuel": {
+    "meta": "https://raw.githubusercontent.com/xXBJXx/ioBroker.e-control-at-fuel/master/io-package.json",
+    "icon": "https://raw.githubusercontent.com/xXBJXx/ioBroker.e-control-at-fuel/master/admin/e-control-at-fuel.png",
+    "type": "misc-data"
+  },
   "egigeozone": {
     "meta": "https://raw.githubusercontent.com/BasGo/ioBroker.egigeozone/master/io-package.json",
     "icon": "https://raw.githubusercontent.com/BasGo/ioBroker.egigeozone/master/admin/egigeozone.png",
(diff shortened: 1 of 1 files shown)


Apollon77:
RE-CHECK!

GermanBluefox:
## Automated adapter checker

### [ioBroker.e-control-at-fuel](https://github.com/xXBJXx/ioBroker.e-control-at-fuel)
[Downloads](https://www.npmjs.com/package/iobroker.e-control-at-fuel) Number of Installations (latest) 
[NPM](https://nodei.co/npm/iobroker.e-control-at-fuel/)

- [ ] :heavy_exclamation_mark: [E201] Bluefox was not found in the collaborators on [NPM](https://www.npmjs.com/package/iobroker.e-control-at-fuel)!. Please execute in adapter directory: `npm owner add bluefox iobroker.e-control-at-fuel`
- [ ] :eyes: [W400] Cannot find "e-control-at-fuel" in latest repository


Apollon77:
Ok, the npm error seems to be a false positive ... bluefox is added

Apollon77:
@xXBJXx  Please adjust the following:
* You use Sentry but this is nowhere informed about in the Readme. Please add information about it. YOu can do like the other adapters (e.g. check Tuya adapter)
* Please remove the onObjectChange handler if not needed, same for on StateChanged
* What all these "awaits" in https://github.com/xXBJXx/ioBroker.e-control-at-fuel/blob/master/main.js#L630 and following lines are doing? I thought the value is an object?

xXBJXx:
@Apollon77  
I added the information about Sentry to the README and
removed the unnecessary await and onObjectChange / StateChanged.

Apollon77:
Hey, thank you, fine in general,  ut one note:
All those https://github.com/xXBJXx/ioBroker.e-control-at-fuel/blob/master/main.js#L611 without await in front of them will run async and so the "try/catch" around it will have no effect ... is this intended? Any error in a setStateAsync would crash the adapter because of an unhandled Promise rejection ... Maybe you want to add the await in from of all relevant ...Async calls?

Apollon77:
Please tell me if you want to adjust it (I would say you should, else one use with a strange response can spam your sentry account to death :-)

xXBJXx:
oh yes i will change it tomorrow

xXBJXx:
I changed the code a little so that it works better and swapped the setStateAsync for setState because in my opinion it is not necessary.

I first tried await this.setStateAsync because there are quite a few await then these messages appear in the log at start

`admin.0 | 2020-07-03 00:57:30.239 | info | (24244) Unsubscribe from all states, except system's, because over 3  seconds the number of events is over 200 (in last second 0)
`

`admin.0 | 2020-07-03 00:58:31.169 | info | (24244) Subscribe on all states again`

GermanBluefox:
How many variables will be written normally with one read?
Brief check shows that for every station 20-50 states. If the user have 5 stations so up to 250 variables will be written all together!

xXBJXx:
I looked at the code again and found a few bugs that caused the problem with the many events and fixed them.
A maximum of 10 stations are created per region and each station has 23 states that are written but not all at once.
Now it's going as it should.

xXBJXx:
I have another adapter that could now be included in the latest [WallPanel](https://forum.iobroker.net/topic/36438/test-adapter-wallpanel)

DutchmanNL:
@Apollon77 e-control : all feedback has been taken care of, positive advice

@xXBJXx 
About Wallpaper : 
- sentry I used but try/catch elements which indicate code errors are not send to sentry
- looks like the create state function is called at each api call causing an extend object (write)
is it needed to changes these objects all the time ?
https://github.com/xXBJXx/ioBroker.wallpanel/blob/master/main.js#L583-L631
- there is no check if state is writeable which could cause un-executable api calls.
Or are all states writable ? : https://github.com/xXBJXx/ioBroker.wallpanel/blob/master/main.js#L698-L731

xXBJXx:
> Über Wallpaper:
> 
> * Sentry Ich habe verwendet, aber Try / Catch-Elemente, die auf Codefehler hinweisen, werden nicht an Sentry gesendet
> * Es sieht so aus, als würde die Funktion zum Erstellen eines Status bei jedem API-Aufruf aufgerufen, wodurch ein erweitertes Objekt (Schreiben) verursacht wird. Ist 
>   es erforderlich, diese Objekte ständig zu ändern? 
>   https://github.com/xXBJXx/ioBroker.wallpanel/blob/master/main.js#L583-L631
> * Es wird nicht überprüft, ob der Status beschreibbar ist, was zu nicht ausführbaren API-Aufrufen führen kann. 
>   Oder sind alle Staaten beschreibbar? : https://github.com/xXBJXx/ioBroker.wallpanel/blob/master/main.js#L698-L731

xXBJXx:
@DutchmanNL 
> About Wallpaper :
> 
> * sentry I used but try/catch elements which indicate code errors are not send to sentry

Before I send messages to sentry I have to take a look at how it should be structured.

> * looks like the create state function is called at each api call causing an extend object (write)
>   is it needed to changes these objects all the time ?
>   https://github.com/xXBJXx/ioBroker.wallpanel/blob/master/main.js#L583-L631

I changed the extendObjectAsync to setObjectNotExistsAsync.

> * there is no check if state is writeable which could cause un-executable api calls.
>   Or are all states writable ? : https://github.com/xXBJXx/ioBroker.wallpanel/blob/master/main.js#L698-L731

All of these states are writable - they are all command states.

DutchmanNL:
thank you, looks good now @Apollon77 positive advice for both

xXBJXx:
I withdraw the pull request and close it since nothing has happened here for months.

Apollon77:
I'm sorry, but also we need to find time to review. Are you still interested?

Apollon77:
@xXBJXx Review comments (if you still want or not your thing):
* please do not set all the states async as you do in https://github.com/xXBJXx/ioBroker.e-control-at-fuel/blob/master/main.js#L208 and following. Please use await and Async methods or other options to make sure that states are set one after the other to not overload the system ... also the use of setState inside an asny c method runs all of that in parallel too
* you do not need to add sentry as a dependency ... the js-controller has that included

PR #790, opened 2020-05-28
Add iobroker.melcloud to latest

Black-Thunder:
Original adapter request: https://github.com/ioBroker/AdapterRequests/issues/339
Forum thread: https://forum.iobroker.net/topic/33855/test-adapter-melcloud-v1-0-x-github

### DIFF
--- a/sources-dist.json
+++ b/sources-dist.json
@@ -750,6 +750,11 @@
     "icon": "https://raw.githubusercontent.com/ausHaus/ioBroker.megaesp/master/admin/megad.png",
     "type": "hardware"
   },
+  "melcloud": {
+    "meta": "https://raw.githubusercontent.com/Black-Thunder/ioBroker.melcloud/master/io-package.json",
+    "icon": "https://raw.githubusercontent.com/Black-Thunder/ioBroker.melcloud/master/admin/melcloud.png",
+    "type": "climate-control"
+  },
   "mercedesme": {
     "meta": "https://raw.githubusercontent.com/TA2k/ioBroker.mercedesme/master/io-package.json",
     "icon": "https://raw.githubusercontent.com/TA2k/ioBroker.mercedesme/master/admin/mercedesme.png",


Apollon77:
RE-CHECK!

GermanBluefox:
## Automated adapter checker

### [ioBroker.melcloud](https://github.com/Black-Thunder/ioBroker.melcloud)
[Downloads](https://www.npmjs.com/package/iobroker.melcloud) Number of Installations (latest) 
[NPM](https://nodei.co/npm/iobroker.melcloud/)

:thumbsup: No errors found
- [ ] :eyes: [W400] Cannot find "melcloud" in latest repository


Apollon77:
Hi,

please check the following:
* why you subscribe for *devices? https://github.com/Black-Thunder/ioBroker.melcloud/blob/master/main.js#L157 what does that mean?
* Additionally most of the time it is better to subscribe for cstate changes once the initi is done (else you might get onStateChange calls but some data are nit initialized ?!
* You might want to check that response is not null/undefined - there are some cases like timeouts where "err" is not defined but response is also empty it is also good to check the HTTP response code

I would like to understand the first topic before accepting it here

Black-Thunder:
Hi, thanks for taking time to review the code.

> * why you subscribe for *devices? https://github.com/Black-Thunder/ioBroker.melcloud/blob/master/main.js#L157 what does that mean?

That's a typo, should be just `"devices.*.control.*"`. I'll correct that.
Generally, only state changes under this specific node(s) are interesting for the adapter because these will trigger a corresponding request to the cloud in order to apply the settings/commands to the device. 

> * Additionally most of the time it is better to subscribe for cstate changes once the initi is done (else you might get onStateChange calls but some data are nit initialized ?!

I think in this specific case it's very unlikely to happen. But you're right, better safe than sorry. :wink: I'll also change that.

> * You might want to check that response is not null/undefined - there are some cases like timeouts where "err" is not defined but response is also empty it is also good to check the HTTP response code

OK, didn't know that. I'll extend the error handling and checks there.

Apollon77:
Thank you! Just give a ping here once done

Black-Thunder:
Should be fine now with v1.0.5. At least I hope so. :wink:

Black-Thunder:
Just wanted to know if you have already had time to check my changes. Or is there any further action from my side needed?

Apollon77:
Everything fine, welcome in the repo

PR #836, opened 2020-06-30
Updated latest innogy-smarthome lib to 0.3.6

PArns:
Some code needs still to be reworked, but we need to bring the adapter back functional first!

I've already fixed some basic stuff, but ofc. more needs to be done


Apollon77:
Good luck

PR #863, opened 2020-07-23
Update stable version for klf200.

MiSchroe:
I constantly get an error in the adapter checker, that bluefox is missing in the owner list of the npm package though I have added the user:
image

### DIFF
--- a/sources-dist-stable.json
+++ b/sources-dist-stable.json
@@ -626,7 +626,7 @@
     "meta": "https://raw.githubusercontent.com/MiSchroe/ioBroker.klf200/master/io-package.json",
     "icon": "https://raw.githubusercontent.com/MiSchroe/ioBroker.klf200/master/admin/klf200.png",
     "type": "hardware",
-    "version": "0.9.5"
+    "version": "1.0.1"
   },
   "km200": {
     "meta": "https://raw.githubusercontent.com/frankjoke/ioBroker.km200/master/io-package.json",


Apollon77:
RE-CHECK!

GermanBluefox:
## Automated adapter checker

### [ioBroker.klf200](https://github.com/MiSchroe/ioBroker.klf200)
[Downloads](https://www.npmjs.com/package/iobroker.klf200) Number of Installations (latest) Number of Installations (stable)
[NPM](https://nodei.co/npm/iobroker.klf200/)

- [ ] :heavy_exclamation_mark: [E201] Bluefox was not found in the collaborators on [NPM](https://www.npmjs.com/package/iobroker.klf200)!. Please execute in adapter directory: `npm owner add bluefox iobroker.klf200`
- [ ] :eyes: [W505] setTimeout found in "build/main.js", but no clearTimeout detected


Apollon77:
For the NPM error: Then it seems to be a falso positive somehow ...

Apollon77:
There seems to be a timeout used which is not clearable in unload?!

MiSchroe:
> For the NPM error: Then it seems to be a falso positive somehow ...

I have analyzed this a bit more: npm view didn't show bluefox as maintainer, either. Thus I tried to fix that and it worked after pressing the button "Update package settings" in npm settings of the package on npmjs.org. The next problem is, that the maintainer in your adapter checker is fetched from npms.io which has outdated data in my case (it shows and older version). It's probably just a matter of time until that one will get fixed automatically.

> There seems to be a timeout used which is not clearable in unload?!

Yes, it is used to wait for a second before retrying the connect and therefore I don't have logic to clear the timer.
https://github.com/MiSchroe/ioBroker.klf200/blob/37615aad0f9bf4ebdb2125aa3e16d2f28125f2b8/src/main.ts#L268
`await new Promise((resolve) => setTimeout(resolve, 1000));`

Apollon77:
> Yes, it is used to wait for a second before retrying the connect and therefore I don't have logic to clear the timer.
https://github.com/MiSchroe/ioBroker.klf200/blob/37615aad0f9bf4ebdb2125aa3e16d2f28125f2b8/src/main.ts#L268
await new Promise((resolve) => setTimeout(resolve, 1000));

Please find a way for the future to make this timeout clearable. This is a big problem else for compact mode: Imagine: adapter is started/stopped in very fast mannor and so "Unload" is called exactly when thiis timeout is running. AFter unload callback is called the adapter DB and such are gone but your code do not know that and tries to acess it ... this will produce big problems up to a crash of the compact mode group.
Alternative to clear such kinds of timeouts is to remember that unload was called and check this in the "resulting" medthod after such a "delay".

PR #909, opened 2020-09-07
Changed icon for vis-icontwo and vis-inventwo

jkvarel:


DutchmanNL:
@Apollon77 pre-check : only logo change content OK

PR #911, opened 2020-09-08

arteck:
fullybrowser
hs100

### DIFF
--- a/sources-dist-stable.json
+++ b/sources-dist-stable.json
@@ -369,7 +369,7 @@
     "meta": "https://raw.githubusercontent.com/arteck/ioBroker.fullybrowser/master/io-package.json",
     "icon": "https://raw.githubusercontent.com/arteck/ioBroker.fullybrowser/master/admin/fully.png",
     "type": "utility",
-    "version": "1.0.3"
+    "version": "2.0.0"
   },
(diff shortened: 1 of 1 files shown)


DutchmanNL:
@arteck should we really release this one or remove from repo and replace with tablet control ?

there are now 2 adapter doing exactly the same which is quite inefficient

arteck:
we have 2 hue adapter and 2 fritzbox and and and..

Apollon77:
@DutchmanNL This adapter is NOT controling the tablet! It is controlling the FullyBrowserAPI ... so it can do much more and is another app and a valid approach, so YES we SHOULD have two different adapters unless Tablet Control also can control all aspects of the FulyBrowser App itself

DutchmanNL:
@Apollon77 it does and more, has been developments by Issi as replacement for the fully Browser Adapter and also discussed with the current owner of fully kiosk Adapter who indicated to stop developing this Adapter as new functionality would not be included anymore that's why I was surprised and mentioned it...

Apollon77:
@arteck ??

arteck:
the other Adapter have not a Stable Status .. where is the Problem ?? the Adater is smaller as the other one..

PR #916, opened 2020-09-12
 Add kostal-piko-ba to stable

hombach:
Add kostal-piko-ba to stable


Apollon77:
RE-CHECK!

hombach:
RE-CHECK!

Apollon77:
Hi,

some review comments:
* please check if you can assign better matching roles to the objects (see https://github.com/ioBroker/ioBroker.docs/blob/master/docs/en/dev/stateroles.md)
* you use Sentry, please add at least an info to the readme about this!
* this.stop is a method or ?! https://github.com/hombach/ioBroker.kostal-piko-ba/blob/master/main.js#L175
* yes cleaning an interval before setting a new one is a good idea to prevent double calls, e.g. (https://github.com/hombach/ioBroker.kostal-piko-ba/blob/master/main.js#L202)

Please check and adjust and inform back here

hombach:
Open points fixed and version bumbed to1.0.2

PR #929, opened 2020-09-18
add Todoist2 Version 0.7.1 to latest

rde-master:
next version of todoist2

### DIFF
--- a/sources-dist.json
+++ b/sources-dist.json
@@ -704,7 +704,7 @@
     "meta": "https://raw.githubusercontent.com/Scrounger/ioBroker.linux-control/master/io-package.json",
     "icon": "https://raw.githubusercontent.com/Scrounger/ioBroker.linux-control/master/admin/linux-control.png",
     "type": "hardware"
-  },  
+  },
   "logparser": {
     "meta": "https://raw.githubusercontent.com/Mic-M/ioBroker.logparser/master/io-package.json",
     "icon": "https://raw.githubusercontent.com/Mic-M/ioBroker.logparser/master/admin/logparser.png",
@@ -1465,6 +1465,11 @@
     "type": "hardware",
     "published": "2019-08-09T19:16:19.154Z",
     "versionDate": "2019-08-09T19:16:19.154Z"
+  },
+  "todoist2": {
+    "meta": "https://raw.githubusercontent.com/rde-master/ioBroker.todoist2/master/io-package.json",
+    "icon": "https://raw.githubusercontent.com/rde-master/ioBroker.todoist2/master/admin/todoist.png",
+    "type": "messaging"
   },
   "tr-064": {
     "meta": "https://raw.githubusercontent.com/iobroker-community-adapters/ioBroker.tr-064/master/io-package.json",


Apollon77:
RE-CHECK!

GermanBluefox:
## Automated adapter checker

### [ioBroker.linux-control](https://github.com/Scrounger/ioBroker.linux-control)
[Downloads](https://www.npmjs.com/package/iobroker.linux-control) Number of Installations (latest) Number of Installations (stable)
[NPM](https://nodei.co/npm/iobroker.linux-control/)

:thumbsup: No errors found

### [ioBroker.todoist2](https://github.com/rde-master/ioBroker.todoist2)
[Downloads](https://www.npmjs.com/package/iobroker.todoist2) Number of Installations (latest) 
[NPM](https://nodei.co/npm/iobroker.todoist2/)

:thumbsup: No errors found
- [ ] :eyes: [W400] Cannot find "todoist2" in latest repository


Apollon77:
hello, some review comments. Please adress them and inform here please:
* please add translations for all languages too io-package (you can use the translation tool)
* is     "stopBeforeUpdate": true, really needed?
* with     "@iobroker/adapter-core": "^2.3.1", the js.controller dependency needs to be set as >=1.5.8 in io-package
* this should not be there because it will be fired alsoe when "ready" event was pot. not yet fired.
* timeout needs to be cleared before newly set else it could run twice (https://github.com/rde-master/ioBroker.todoist2/blob/master/main.js#L2654)

rde-master:
Hello, here are my comments:

_please add translations for all languages too io-package (you can use the translation tool)_
--> i have adden translation to desc. 
But i do not do that for the news section. That ist not needed.

_is "stopBeforeUpdate": true, really needed?_
--> yes it is! sometimes i get an error without it.
_with "@iobroker/adapter-core": "^2.3.1", the js.controller dependency needs to be set as >=1.5.8 in io-package_
--> its done
_this should not be there because it will be fired alsoe when "ready" event was pot. not yet fired._
--> what ist this??
_timeout needs to be cleared before newly set else it could run twice_ 
--> now cleared is in

Version 0.7.2 is in Github and npm

Apollon77:
>  But i do not do that for the news section. That ist not needed.

Why? ioBroker offers translation tools to automate this ... but in fact without localized news users from other languages will fall back to english ...

>  this should not be there because it will be fired alsoe when "ready" event was pot. not yet fired.


ohh ... missing link: https://github.com/rde-master/ioBroker.todoist2/blob/master/main.js#L2725

rde-master:
Then why are there no automatic translations when there are automatic tools?
If the respective language does not exist, the ioBroker can do it itself.
I can't confirm the current translation anyway because I can't speak the languages.

i have now translated the news section manually with the tool.



This code is in the "main" function. This is only carried out when the adapter is "ready".
It is there to ensure that the "main" function is executed again after the set time.
I can think of no other way to do this otherwise.

Apollon77:
> Then why are there no automatic translations when there are automatic tools?
If the respective language does not exist, the ioBroker can do it itself.

This would be an interesting feature request for admin, but I'm not sure if this will work performance wise "on the fly".

> This code is in the "main" function. This is only carried out when the adapter is "ready".
It is there to ensure that the "main" function is executed again after the set time.
I can think of no other way to do this otherwise. 

No this setTimeout is triggered on the "main level". Without the "ready" event you can not be sure that your DB connection to ioBroker is estabished or such, so if the initialization needs more then "poll" time (because of whatever reason" triggering the "main" will not help at all because you can not interact with iobroker and will most likely throw a lot of errors and proces gets killed. Your "Main" is triggered by your code in "ready" handler so this is completely enough. It makesno sense to work around the ready event - it is required, else nothing will work!

rde-master:
I have now completely deleted the "setTimeout" functions. And instead inserted a "setinterval" in the "Ready" area.

I hope it's better now
If so, please give me a little feedback. then I upload the new version to NPM.

Apollon77:
you can do it that way. Personally I can only say that setInterval with "externel endpoints" are always dangerous and so the setTimeout in general is mich better - and you setTimeout was completely ok - yonly the "extra one I told you should had been removed). Why setInterval is bad? Easy: When there are Timeouts or such in external communication then it can be that requests pile up because the interval fires independently if the former run was completes ...

rde-master:
Yes, i can see the Problem, so i have some code to prevent that:
https://github.com/rde-master/ioBroker.todoist2/blob/master/main.js#L2667

if the loop fails 10 times, then it is ended and the adapter has to be restarted.


this can also be changed, e.g. then a new run is made e.g. done in an hour.

I would like to decide how it works in practice.

The other code in the main function will not be executed if there is no connection to the API.

Apollon77:
Hey, I did check again,

above you said

>  with "@iobroker/adapter-core": "^2.3.1", the js.controller dependency needs to be set as >=1.5.8 in io-package

https://github.com/rde-master/ioBroker.todoist2/blob/master/io-package.json#L4 ... needs to be like that https://github.com/ioBroker/ioBroker.digitalstrom/blob/master/io-package.json#L174-L178

Apollon77:
Sorry if I was not clear enough before

rde-master:
sorry for the late answer.
I have Changed it in Version 0.8.3.
Also i have changed more since 7.1. There is now axois in it.
There is much better error handeling now in it.

I hope now we can release the version ?

Apollon77:
But the js-controller dep is still wrong ... need to point to 1.5.8 if you use adapter-code 2.3.1!

Apollon77:
please also remover https://github.com/rde-master/ioBroker.todoist2/blob/master/main.js#L191 if you do not need it

Apollon77:
Rest is Ok :-)

rde-master:
I have removed it in Version 0.8.4  (it is on GitHub and NPM)

Apollon77:
Then still the js-controller dep is missing to be fixed :-))
https://github.com/rde-master/ioBroker.todoist2/blob/master/io-package.json#L255 ... need to be 1.5.8

rde-master:
oh sorry,... now it is right... :-) i hope :-D

PR #997, opened 2020-11-24
Update sources-dist.json

Steiger04:
image folder has changed.


Apollon77:
thank you I also fix it in stable

PR #1004, opened 2020-11-28
#1003 Added legrand-ecocompteur to stable

raintonr:


raintonr:
You will notice the build fails, but this is nothing to do with me:

```
latest: [377/378] Check zwave2
    1) Test all Packages in latest are loadable via http and name is equal to io-package.json are
```

Apollon77:
i know

PR #1017, opened 2020-12-03
Add Growatt Adapter 0.0.14 to stable

PLCHome:
The tester thread can be found here:
[https://forum.iobroker.net/topic/36474/test-adapter-growatt-v0-0-14](https://forum.iobroker.net/topic/36474/test-adapter-growatt-v0-0-14)

### DIFF
--- a/sources-dist-stable.json
+++ b/sources-dist-stable.json
@@ -389,6 +389,12 @@
     "type": "geoposition",
     "version": "0.3.2"
   },
+  "growatt": {
+    "meta": "https://raw.githubusercontent.com/PLCHome/ioBroker.growatt/master/io-package.json",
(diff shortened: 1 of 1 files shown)


PLCHome:
From unstable merge request:
>Hello,
>sorry that it took that long. One personal note: an Interval is risky combined with external communication because interval fires independent if the run before was finished ... better is Timeouts that are set anew at the end of an action.
>else is ok, I added the adapter to the repo

Interval to Timeouts is fixed in 0.0.14

PLCHome:
image

PLCHome:
RE-CHECK!

PLCHome:
RE-CHECK!

PR #1019, opened 2020-12-05
Removed nello, Updated LaMetric and OctoPrint

klein0r:


Apollon77:
Nello is from Zefau ... why you remove it? :-) Thank you for info

klein0r:
@Apollon77 See adapter readme "nello API has been shut down. This adapter will thus not work."

PR #1023, opened 2020-12-06
Add MyStrom to latest

TA2k:


Apollon77:
RE-CHECK!

GermanBluefox:
## Automated adapter checker

### [ioBroker.mystrom](https://github.com/TA2k/ioBroker.mystrom)
[Downloads](https://www.npmjs.com/package/iobroker.mystrom) Number of Installations (latest) 
[NPM](https://nodei.co/npm/iobroker.mystrom/)

:thumbsup: No errors found
- [ ] :eyes: [W400] Cannot find "mystrom" in latest repository


Apollon77:
Hi,some review comments:
* When you have a password and want to protect it event better you can also set that key in "protectedNative" (additional to encryptedNative) to prevent other adapters to access that object property
* best practice for external communication would be not to use Intervals ... https://github.com/TA2k/ioBroker.mystrom/blob/master/main.js#L61 ... on timeout or other cases requests can be added over and over
* you should reset connection status here https://github.com/TA2k/ioBroker.mystrom/blob/master/main.js#L66 , or ?! else noone will see that there is a problem
* this log in https://github.com/TA2k/ioBroker.mystrom/blob/master/main.js#L89 before checking that it is set can kill the adapter :-) ... in other places it is in a try/catch, but that a debug log prevent other code from ecxecuting migth be not that good ;-)

In fact nothing that prevent the adapter to be added, just "notes" from me

PR #1053, opened 2021-01-05
add fuelpricemonitor to latest

HGlab01:


HGlab01:
RE-CHECK!

GermanBluefox:
## Automated adapter checker

### [ioBroker.fuelpricemonitor](https://github.com/HGlab01/ioBroker.fuelpricemonitor)
[Downloads](https://www.npmjs.com/package/iobroker.fuelpricemonitor) Number of Installations (latest) 
[NPM](https://nodei.co/npm/iobroker.fuelpricemonitor/)

:thumbsup: No errors found
- [ ] :eyes: [W400] Cannot find "fuelpricemonitor" in latest repository


Apollon77:
hey, some review comments:
* when using adapter-core 2.4.0 as dep you need to requre js-controll2r >=2.0.0 in io-package
* would this not better a schedule adapter?
* e.g. https://github.com/HGlab01/ioBroker.fuelpricemonitor/blob/master/main.js#L119 ... always check that response and response.body is provided! and I would also add a try/catch around the JSON.parse  (what if an html error page was returned??) so please add more error handling
* weird construct https://github.com/HGlab01/ioBroker.fuelpricemonitor/blob/master/main.js#L148
* basically you could also remember that an unload happened because if in that second the requests are executed then a new timeout could be set

Please check and adjust the cases. Thank you

HGlab01:
Hi @Apollon77!
Thanks for your feedback!

> when using adapter-core 2.4.0 as dep you need to requre js-controller >=2.0.0 in io-package

not sure what you mean. I guess this dependency is managed by adapter-core and therefore I did not define a dependency to js-controller. Is this explicit dependency needed?
> would this not better a schedule adapter?

agree :-)
> e.g. https://github.com/HGlab01/ioBroker.fuelpricemonitor/blob/master/main.js#L119 ... always check that response and response.body is provided! and I would also add a try/catch around the JSON.parse (what if an html error page was returned??) so please add more error handling

Response and response.body understood! JSON.parse is within a try/catch isn't?
```
try {
    resolve(JSON.parse(response.body));
} catch (error) {

```
or shall I do it in two steps?
```
try {
    if (!response || !response.body) {
        error = `Response or response.body empty in getData()`;
    } else {
        let result = JSON.parse(response.body);
        resolve(result);
    }
} catch (error) {
```
Thanks for your teaching ;-)

Apollon77:
For js-controller dep: adjust here https://github.com/HGlab01/ioBroker.fuelpricemonitor/blob/master/io-package.json#L165

For try catch it is a matter of code style ... but you areright in fact it was catched ... the other code is more clean ... up to you :-)

Apollon77:
Do you want to change type to schedule? Easier to be done before it gets installs while in repo because updating type "in production " is very meeehhh

HGlab01:
> Do you want to change type to schedule? Easier to be done before it gets installs while in repo because updating type "in production " is very meeehhh

yes, sure, I will do and as soon as all changes are done I give you a hint!

> For js-controller dep: adjust here https://github.com/HGlab01/ioBroker.fuelpricemonitor/blob/master/io-package.json#L165

ah, sorry, was not aware! will change

HGlab01:
sorry, one additional supporting question, @Apollon77:
is there a way to configure a schedule adapter, to run after install and/or after pressing image
Currently fist run will be executed on cron-job-trigger

Apollon77:
sure https://github.com/iobroker-community-adapters/ioBroker.weatherunderground/blob/master/io-package.json#L288

HGlab01:
Hi @Apollon77 again!
Your feedback is incorporated

PR #1056, opened 2021-01-07
statistics -> 1.0.1

foxthefox:


Apollon77:
Can I get a 1.0.2 with fix for that please asap too?

https://github.com/foxthefox/ioBroker.statistics/issues/54

Apollon77:
Or: is the 1.0.1 more possibly hitting thzat error or was it the same risk in 0.2.2?

foxthefox:
> Or: is the 1.0.1 more possibly hitting thzat error or was it the same risk in 0.2.2?

I think it it is in both, the code is more or less similar and in both cases without try catch, I make nom today and a new PR for 1.0.2

PR #1068, opened 2021-01-13
Please update sources-dist-stable.json for milight-smart-light adapter to version 1.1.1

Steiger04:


Steiger04:
I found another bug. Please do not take over yet.

PR #1092, opened 2021-01-23
Pls add new Adapter for LG ESS Home Inverter

Morluktom:
This is a new pull request. On the old one was no activity over one month

### DIFF
--- a/sources-dist.json
+++ b/sources-dist.json
@@ -720,6 +720,11 @@
     "icon": "https://raw.githubusercontent.com/AWhiteKnight/ioBroker.letrika_comgw/master/admin/letrika_comgw.png",
     "type": "energy"
   },
+  "lg-ess-home": {
+    "meta": "https://raw.githubusercontent.com/Morluktom/ioBroker.lg-ess-home/master/io-package.json",
+    "icon": "https://raw.githubusercontent.com/Morluktom/ioBroker.lg-ess-home/master/admin/lg-ess-home.png",
+    "type": "energy"
+  },  
   "lgtv": {
     "meta": "https://raw.githubusercontent.com/SebastianSchultz/ioBroker.lgtv/master/io-package.json",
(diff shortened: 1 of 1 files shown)


Apollon77:
Ok, I see it :-))

Apollon77:
RE-CHECK!

GermanBluefox:
## Automated adapter checker

### [ioBroker.lg-ess-home](https://github.com/Morluktom/ioBroker.lg-ess-home)
[Downloads](https://www.npmjs.com/package/iobroker.lg-ess-home) Number of Installations (latest) 
[NPM](https://nodei.co/npm/iobroker.lg-ess-home/)

:thumbsup: No errors found
- [ ] :eyes: [W400] Cannot find "lg-ess-home" in latest repository


Apollon77:
Hey, 
some review comments:
* for mass object creation and stteset please "await" the object creation (e.g. https://github.com/Morluktom/ioBroker.lg-ess-home/blob/master/lib/LgEss.js#L235 and more)
* you should also await setState to lower parallel executed stuff - also check with js-controller 3.2 that no warning appear!
* how many writeable fields exist in your adapter? subscribing each single one works ok up to a certain number, if igher subscring to * and handling in adapter has better performance ... (https://github.com/Morluktom/ioBroker.lg-ess-home/blob/master/lib/LgEss.js#L324)
* please also add a safeguard for the case that "unload" was called (and so all timer were stopped) but in fact a request was still running. this would then be processed and a new timer would be set ... without DB and in compact mode ends in an error flood :-)

Thank you for checking my comments ...

Morluktom:
Hi Apollon, 
rework done.
I have only 4 items you can write, so i think a single subscrib should be ok, or?
thanks
Tom

Apollon77:
You should also await all calls to createDataPoint :-)) else the problem is still there.

Morluktom:
sorry did not understand. ;)
done

PR #1165, opened 2021-03-08
add anel hut to latest

dan1-de:
Hi, 
please add my adapter for anelhut to stable. 
This adapter was tested with the developer of anel. 
All devices are working.

Br
Daniel

### DIFF
--- a/sources-dist.json
+++ b/sources-dist.json
@@ -39,6 +39,11 @@
     "icon": "https://raw.githubusercontent.com/PArns/ioBroker.amazon-dash/master/admin/amazon-dash.png",
     "type": "hardware"
   },
+  "anelhut": {
+    "meta": "https://raw.githubusercontent.com/dan1-de/ioBroker.anelhut/master/io-package.json",
+    "icon": "https://raw.githubusercontent.com/dan1-de/ioBroker.anelhut/master/admin/anelhut.png",
+    "type": "iot-systems"
+  },
   "artnet": {
     "meta": "https://raw.githubusercontent.com/ioBroker/ioBroker.artnet/master/io-package.json",
     "icon": "https://raw.githubusercontent.com/ioBroker/ioBroker.artnet/master/admin/artnet.png",


Apollon77:
RE-CHECK!

GermanBluefox:
## Automated adapter checker

### [ioBroker.anelhut](https://github.com/dan1-de/ioBroker.anelhut)
[Downloads](https://www.npmjs.com/package/iobroker.anelhut) Number of Installations (latest) 
[NPM](https://nodei.co/npm/iobroker.anelhut/)

- [ ] :heavy_exclamation_mark: [E140] width and height of logo are not equal


dan1-de:
Sorry, but I have no logo with same width and height.

Apollon77:
Hello, here are my review comments:
* please add  the default values for all configation options to io-package native https://github.com/dan1-de/ioBroker.anelhut/blob/main/io-package.json#L75-L78
* I know that it will be manual work but would it not be better to encrxpt the passwords stored here? I'm sorry but the auto-encrypt feature from cointroller can not be used in "arrays" :-(
* it seems that info.connection is missing as object? https://github.com/dan1-de/ioBroker.anelhut/runs/2058555301#step:6:233 ... ideally add it to instanceObjects in io-package
* is the subscribe here https://github.com/dan1-de/ioBroker.anelhut/blob/75a6aee8e3dadb5f18d3a2a70da5b4cb70b9b2b9/src/main.ts#L118 really to be executed on every update? SHould not be a bigger issue, but still strange
* I think it needs unneeded calls that you check the object existence on every write ... I think this is enough once (you can use a map /Objects-keys to remember which object was created/checked already for instance. Same for all the devices/channels in update method
* I also think that the template code https://github.com/dan1-de/ioBroker.anelhut/blob/75a6aee8e3dadb5f18d3a2a70da5b4cb70b9b2b9/src/main.ts#L236-L282 can easiely be removed, or ?! ;-)
* * you need to clear all timeout and connections in https://github.com/dan1-de/ioBroker.anelhut/blob/75a6aee8e3dadb5f18d3a2a70da5b4cb70b9b2b9/src/main.ts#L288-L299 (e.g. the socket/dgram you open in your communication class
* I would also see it as best practice to add a try catch around the socket bin call because it can fail if the port is already used or such.

Please check the comments and adjust accordingly. Please poke here again when I should have another look

Apollon77:
for 1.) best conact use in Telegram or Discord ... there are developer groups

for 2.) I would simply make a object "createdObjects={};" and as soon as you create one set the ID there as true and check if it is set 

I will re-reiew the next days

dan1-de:
HI @Apollon77, 

can we move this to latest ?

Apollon77:
Can you please check the testing of your adapter. The last tests failed ... And yes afte rthis we can go ahead

dan1-de:
RE-CHECK!

GermanBluefox:
## Automated adapter checker

### [ioBroker.anelhut](https://github.com/dan1-de/ioBroker.anelhut)
[Downloads](https://www.npmjs.com/package/iobroker.anelhut) Number of Installations (latest) 
[NPM](https://nodei.co/npm/iobroker.anelhut/)

- [ ] :heavy_exclamation_mark: [E140] width and height of logo are not equal


dan1-de:
@Apollon77 
I can not run the unit tests. Is this an error with the CI toolchain ?
See: https://github.com/dan1-de/ioBroker.anelhut/pull/17/checks?check_run_id=3154532029

Apollon77:
@dan1-de just remove testing for nodejs 10 and add node.js 16 ... nodjes 10 is eol

Apollon77:
Please just poke when I should have another look

dan1-de:
Hi @Apollon77, 

I removed the node v10 and the tests are working now with node v12 & v14.
image

Apollon77:
Great, then welcome. Would be awesome if you can check the topics listed above, but no showstoppers

PR #1177, opened 2021-03-14
Add eufy-security adapter

bropat:


Apollon77:
RE-CHECK!

GermanBluefox:
## Automated adapter checker

### [ioBroker.eufy-security](https://github.com/bropat/ioBroker.eufy-security)
[Downloads](https://www.npmjs.com/package/iobroker.eufy-security) Number of Installations (latest) 
[NPM](https://nodei.co/npm/iobroker.eufy-security/)

:thumbsup: No errors found
- [ ] :eyes: [W400] Cannot find "eufy-security" in latest repository


Apollon77:
Hey,

here some review comments
* best practice is to use path.join instead of https://github.com/bropat/ioBroker.eufy-security/blob/master/src/main.ts#L85 ... but just nitpick :-)
* (needed) you should remove that part here because uneeded with the auto decrypt feature https://github.com/bropat/ioBroker.eufy-security/blob/master/src/main.ts#L96-L106
* really hard get object from instance 0? https://github.com/bropat/ioBroker.eufy-security/blob/master/src/main.ts#L259
* better add try/catch around all filesystem calls ... they can crash the adapter ... https://github.com/bropat/ioBroker.eufy-security/blob/master/src/main.ts#L324
* storing the timeout directly in adapter instance (https://github.com/bropat/ioBroker.eufy-security/blob/master/src/main.ts#L332) can lead to issues if we add such a variable in adapter. so choose a more "non-potential-conflicting" name please :-)
* please clear all timeouts before doing other actions that can generate an exception to make sure they are cleared in any case https://github.com/bropat/ioBroker.eufy-security/blob/master/src/main.ts#L365-L384
* maybe use value.temperature as role here https://github.com/bropat/ioBroker.eufy-security/blob/master/src/main.ts#L975 ... and indicator.lowbat in https://github.com/bropat/ioBroker.eufy-security/blob/master/src/main.ts#L1347 ... please also check the others maybe again (see https://github.com/ioBroker/ioBroker.docs/blob/master/docs/en/dev/stateroles.md)
Thank you, most are advices - only one change is required. Please check and adjust and poke here again when done

bropat:
Hi @Apollon77,

I have corrected everything. Let me know if it fits now. ;-)

Apollon77:
Hi @bropat,

thank you very much.
* Still question: why you request hard coded instance 0 here https://github.com/bropat/ioBroker.eufy-security/blob/master/src/main.ts#L247 ?
* all the stuff in https://github.com/bropat/ioBroker.eufy-security/blob/master/src/main.ts#L365-L380 are allowed to take max 500ms ... is that for sure? If not you can increase that 500ms timeout in io-package by setting common.stopTimeout to an higher value (just asking because I do not know what all happens there)

The rest looks fine. Thank you for info on those topics and for your patience

bropat:
Hi @Apollon77,

1. I actually need the version of the current instance. That "0" is hardcoded here is not correct, you are absolutely right.
What is the best way to do it?
2. Yes, here only the motion states are set to false, the persistent info of the adapter is written to the filesystem (few things) and the connection with the station/device is disconnected via UDP (P2P) (this is asynchronous; UDP packet is only sent).

Thank you. No problem ;) You're welcome.

Apollon77:
1.) this.namespace returns "adaptername.instance" ... I think also an "this.instance" exists. Use that :-)

Apollon77:
@bropat ready for another review round?

bropat:
@Apollon77 

Sorry for the late reply. I am ready :)

ldittmar81:
@Apollon77 something is wrong here. I cannot found eufy-security on the list!

PR #1182, opened 2021-03-17
update stable version of Discovergy adapter 

DutchmanNL:


Apollon77:
MERGE BLOCKED RIGHT NOW because also contains ESPHome where review is pending

PR #1186, opened 2021-03-18
Speedport Adapter

hacki11:
**Checklist**  
[x] The adapter is published to NPM?
[x] Developer Best practices are known and considered (see https://github.com/ioBroker/ioBroker.repositories#development-and-coding-best-practices)
[x] All requirements to bring a new adapter into Latest repository are fulfilled (see https://github.com/ioBroker/ioBroker.repositories#requirements-for-adapter-to-get-added-to-the-latest-repository)
[ ] The adapter was checked by https://adapter-check.iobroker.in/ and no warnings or errors are shown? (If travis is mentioned but testing is ok then ignore this please)
Two errors are shown:
* some git url issue => what should i change?
* bluefox added in npm as maintainer but checker complains 

[x] Please add a link to Forum testing thread, if existing.
https://forum.iobroker.net/topic/43356/test-adapter-speedport-v0-0-x

PR #1198, opened 2021-03-23
Add speedport to latest

hacki11:
The old PR was deleted automatically because i had to reset my master branch.
https://github.com/hacki11/ioBroker.speedport/

### DIFF
--- a/sources-dist.json
+++ b/sources-dist.json
@@ -1512,6 +1512,11 @@
     "icon": "https://raw.githubusercontent.com/iobroker-community-adapters/ioBroker.sourceanalytix/master/admin/sourceanalytix.png",
     "type": "energy"
   },
+  "speedport": {
+    "meta": "https://raw.githubusercontent.com/hacki11/ioBroker.speedport/master/io-package.json",
+    "icon": "https://raw.githubusercontent.com/hacki11/ioBroker.speedport/master/admin/speedport.png",
+    "type": "infrastructure"
+  },
   "spotify-premium": {
     "meta": "https://raw.githubusercontent.com/iobroker-community-adapters/ioBroker.spotify-premium/master/io-package.json",
     "icon": "https://raw.githubusercontent.com/iobroker-community-adapters/ioBroker.spotify-premium/master/admin/spotify-premium.png",


Apollon77:
old PR was https://github.com/ioBroker/ioBroker.repositories/pull/1186 ... for review reference

Apollon77:
RE-CHECK!

GermanBluefox:
## Automated adapter checker

### [ioBroker.speedport](https://github.com/hacki11/ioBroker.speedport)
[Downloads](https://www.npmjs.com/package/iobroker.speedport) Number of Installations (latest) 
[NPM](https://nodei.co/npm/iobroker.speedport/)

:thumbsup: No errors found
- [ ] :eyes: [W400] Cannot find "speedport" in latest repository


hacki11:
Review findings had been fixed in the last pr, any tasks for me left here?

Apollon77:
Stay tned for review update

Apollon77:
@hacki11  For the encrypt feature the admin dependency needs to be in "globalDependencies" :-) Please adjust that last thing then we are ready to go

hacki11:
what do you mean with globalDependencies? Is this a package.json or a io-package.json entry? Or is it just dependencies in package.json?
I took https://github.com/Apollon77/ioBroker.meross as a template for the encryption topic, i couldn't find a admin dependency there.

Happy to do, but a bit stuck where

Apollon77:
Meros is bad example because meros has "backward compatibility" for encryption and so do not need this :-)

But this is meansed: https://github.com/Vertumnus/ioBroker.googleauth/blob/master/io-package.json#L102-L106

Apollon77:
@hacki11 ready for next review round?

hacki11:
Currently the adapter is in development. Is it ok to have such versions in latest? Than yes, the current finding you mentioned is fixed. Of course new code was coming in again. It is more a continuous process than a one time quality gate in the current stage

Apollon77:
You need to decide that. In Latest Repo people officially see the adapter and will more likely install it ... if you think it is ok for that then go ... if not we can close here and you open again when you are ready for it

hacki11:
Then hit me with your findings! More people more testing, it's a good thing

Apollon77:
Ok, in general honestly I  have not that much to note. The main thing is: How long does it take to complete one "data fetch"? In fact the scenario that could happen is that the timer triggers and you "just start to get the data" and then unload is called. Then two things will happen (especially when in compact mode): Your code will run and will try to set all data (but DB is already gone because unload is kind of directly triggering the callback) that leads to a lot of logs ... but then you will set the timer anew and so anything will continue to run but it should not.
SO better remember when unload was called and add checks to the relevant places to "stop" if unload was called.

Rest is completely fine

hacki11:
Best case would be to cancel all promises on unload and only reset the timeout when not already unloaded, correct? Or ist it sufficient to await the promises and then quit without setting objects? Thx for the review

Apollon77:
No idea if you really can cancel promises ... If you await anything then you need to maybe adjust the time for unload (default is 500ms). So best is set a variable "stopped" to true and check after calls are done and just do not continue if stopped is set

Apollon77:
@hacki11 SHould we bring the adapter to latest ? What do you think?

Apollon77:
@hacki11 ?

hacki11:
Go for it, let's merge👍

PR #1245, opened 2021-04-25
Add ioBroker.pixelit to latest

o0shojo0o:


o0shojo0o:
RE-CHECK!

GermanBluefox:
## Automated adapter checker

### [ioBroker.pixelit](https://github.com/o0shojo0o/ioBroker.pixelit)
[Downloads](https://www.npmjs.com/package/iobroker.pixelit) Number of Installations (latest) 
[NPM](https://nodei.co/npm/iobroker.pixelit/)

:thumbsup: No errors found
- [ ] :eyes: [W400] Cannot find "pixelit" in latest repository


Apollon77:
Hi, some review comments:
* https://github.com/o0shojo0o/ioBroker.pixelit/blob/master/io-package.json#L144 ... there is no js.controller 2.4 :-) I think you mean >= 2.0.0
* normally you check for ack=true to decide if the adapter needs to react on a change (https://github.com/o0shojo0o/ioBroker.pixelit/blob/master/main.js#L98)

Rest is fine, please check these two things.

o0shojo0o:
Thanks for your reviewe.

> Hi, some review comments:
> 
> * https://github.com/o0shojo0o/ioBroker.pixelit/blob/master/io-package.json#L144 ... there is no js.controller 2.4 :-) I think you mean >= 2.0.0

Oh yeah, that's what I mean ;) (fixed)

> * normally you check for ack=true to decide if the adapter needs to react on a change (https://github.com/o0shojo0o/ioBroker.pixelit/blob/master/main.js#L98)
> 
> Rest is fine, please check these two things.

In this case the ack=true has been set by the adapter itself after it has successfully sent the input to the API.
Therefore the event is not to be processed further by the adapter.

Apollon77:
> In this case the ack=true has been set by the adapter itself after it has successfully sent the input to the API.

Yes but what you do if a user sets ack=true? Then this should also not go out, but would because from then is not from your adapter

o0shojo0o:
> Yes but what you do if a user sets ack=true? Then this should also not go out, but would because from then is not from your adapter

I think I have understood it now... ;)

Apollon77:
@o0shojo0o ready for another check?

o0shojo0o:
@Apollon77 yes we can ;o)

Apollon77:
Then welcome

PR #1252, opened 2021-05-04
added ikettle2

Jey-Cee:


Apollon77:
RE-CHECK!

Apollon77:
Yes please add .title, else there are issues with Admin, we reverted in adapter-creator, sorry

Some more review comments:
* you may want to cleanup https://github.com/Jey-Cee/ioBroker.ikettle2/blob/master/admin/words.js
* best call callback in unload in the callback from socket.end() to make sure it is really finished before callinf unload-callback
* please check for ack flag and not from in https://github.com/Jey-Cee/ioBroker.ikettle2/blob/master/main.js#L122
* you set the "ack" flag before even executing the action? https://github.com/Jey-Cee/ioBroker.ikettle2/blob/master/main.js#L126-L153 

Please check the comments

Jey-Cee:
> * you set the "ack" flag before even executing the action? https://github.com/Jey-Cee/ioBroker.ikettle2/blob/master/main.js#L126-L153

Yes, because the value will be changed inside the function. So if i want to see if the adapter has accepted the command this is needed. 

Other things are done.

Apollon77:
Sorry for the delay ... and welcome in the repo

PR #1302, opened 2021-06-10
(simatec) solax added

simatec:


Apollon77:
RE-CHECK!

simatec:
RE-CHECK!

GermanBluefox:
## Automated adapter checker

### [ioBroker.solax](https://github.com/simatec/ioBroker.solax)
[Downloads](https://www.npmjs.com/package/iobroker.solax) Number of Installations (latest) 
[NPM](https://nodei.co/npm/iobroker.solax/)

:thumbsup: No errors found
- [ ] :eyes: [W400] Cannot find "solax" in latest repository


simatec:
@Apollon77 ping ;-)

Apollon77:
Just a small topics:
* Please do admin deps (https://github.com/simatec/ioBroker.solax/blob/1e1ba7d8c70ed74f9c03c4aad8d801cc5f138390/io-package.json#L120) in "globalDependencies" unless you really need that admin version "on the same host as the adapter"
* some of your rtoles in io-package and code do not make sense or are invalid  ... e.g. indicators are booleans by definition (see https://github.com/ioBroker/ioBroker.docs/blob/master/docs/en/dev/stateroles.md#indicators-boolean-read-only) ... maybe value would fit better
* When async7await then please noc allbacks ... that gives chaos!! https://github.com/simatec/ioBroker.solax/blob/master/main.js#L268 (example) - also here long/lat is initialized asny later probably (because no promise is returned when callback is there I think) ... also here and other places https://github.com/simatec/ioBroker.solax/blob/master/main.js#L653
* is the data request really needed x:x:20 every minute exactly then? Becaus ethe effect here is that ALL users that use this adapter will reach the solax cloud system at the exact same time (if time is ntp synced as usual today it is even close in ms) ... depeneding on how many this becomes a DDOS against the cloud system! Rethink that because it can give big issues with the solax guys. Same for the second schedule.
Please check and adjust

simatec:
@Apollon77 Thanks for your feedback. I tried to fix everything in version 0.3.2

simatec:
@Apollon77 
In version 0.3.3 I have adjusted all roles and revised the await functions. Furthermore, the interval is now executed as discussed using setInterval. Now everything should fit for the latest ;-)

Apollon77:
* why you removed the admin dep?? It is still relevant :-) but please do in "common.globalDependencies"
* async unneeded here https://github.com/simatec/ioBroker.solax/blob/master/main.js#L269 because you use no await
* you still have await/callback mixes in e.g. https://github.com/simatec/ioBroker.solax/blob/master/main.js#L643
The latter point could be rewitten to

```
                        const state = await adapter.getStateAsync(`data.${resultID}`);
                        if (state && state.val >= 0) {
                            json[`${resultID}`] = state.val;
                        }
```
and formally a try/catch around it for error cases
* I think you missed that line to remove? https://github.com/simatec/ioBroker.solax/blob/master/main.js#L718

simatec:
@Apollon77 
As discussed in Telegram, I have fixed all async / await problems in version 0.3.4 and removed the remaining bugs as well

Apollon77:
welcome

simatec:
@Apollon77 many thx 🙏 😊

PR #1341, opened 2021-07-18
Add luxtronik2 to latest

UncleSamSwiss:
Just to be clear: this is not an alternative to luxtronik1. These are two different hardwares which require different ways to talk to them.

### DIFF
--- a/sources-dist.json
+++ b/sources-dist.json
@@ -885,6 +885,11 @@
     "icon": "https://raw.githubusercontent.com/iobroker-community-adapters/ioBroker.luxtronik1/master/admin/luxtronik1.png",
     "type": "climate-control"
   },
+  "luxtronik2": {
+    "meta": "https://raw.githubusercontent.com/UncleSamSwiss/ioBroker.luxtronik2/master/io-package.json",
+    "icon": "https://raw.githubusercontent.com/UncleSamSwiss/ioBroker.luxtronik2/master/admin/luxtronik2.png",
+    "type": "climate-control"
+  },
   "material": {
     "meta": "https://raw.githubusercontent.com/ioBroker/ioBroker.material/master/io-package.json",
     "icon": "https://raw.githubusercontent.com/ioBroker/ioBroker.material/master/admin/material.png",


Apollon77:
RE-CHECK!

GermanBluefox:
## Automated adapter checker

### [ioBroker.luxtronik2](https://github.com/UncleSamSwiss/ioBroker.luxtronik2)
[Downloads](https://www.npmjs.com/package/iobroker.luxtronik2) Number of Installations (latest) 
[NPM](https://nodei.co/npm/iobroker.luxtronik2/)

:thumbsup: No errors found
- [ ] :eyes: [W400] Cannot find "luxtronik2" in latest repository


Apollon77:
Hey, some small stuff:
* Please all admin>=4.0.9 dep as common.globalDependencies to use the passwoerd encryption
* formally "close websocket" is async, callback should be called when closed ... callback or event or such? Can give issues in compact mode when restart is too fast
* do you really do extendObject in each data response process? Please not ... only if needed. This will cost a lot performance asnd produces I/O because objects are lareays updated (timestamp)

Please check and adjust

Apollon77:
@UncleSamSwiss any update? :-)

UncleSamSwiss:
> * Please all admin>=4.0.9 dep as common.globalDependencies to use the passwoerd encryption

Fixed.

> * formally "close websocket" is async, callback should be called when closed ... callback or event or such? Can give issues in compact mode when restart is too fast

What do you mean by this? `WebSocket.close()` doesn't return a Promise. Also, all handlers (e.g. `this.webSocket.on('close', ...)` make sure the adapter is not closing, and if it is, the handler does nothing. I assume, in compact mode a new instance of the adapter class is created for every "restart", so there shouldn't be any conflict.

> * do you really do extendObject in each data response process? Please not ... only if needed. This will cost a lot performance asnd produces I/O because objects are lareays updated (timestamp)

No, every call to `extendObjectAsync()` is depending on a condition that is only true at start-up:
- the method `createLuxTreeAsync()` is only called from `onReady()`
- the method `handleWsMessageAsync()` is only called with a `NavigationMessage` when the `REFRESH` message was sent to the heat pump which only happens from `this.webSocket.on('open', ...)`, thus only at start-up or when reconnecting.
- the handlers' `extendObjectAsync()` is only called after creating a handler that didn't exist yet


> * it sseems that the luxtronic class used also created a network connection ... but this is not closed in unload or?!

You are right, there is an undocumented socket being open for a few milliseconds, this is now fixed.

@Apollon77 please check my comments and provide me feedback. If everything is OK, I will create a new release and update this PR afterwards.

Apollon77:
For close: aah yes it's a Websocket, then ok - we can not do anything more.

So rest ok for me. New version released? Then we could merge it

PR #1355, opened 2021-07-29
Go-e from Version 1.0.13 to 1.0.15 on stable

MK-2001:
Go-e from Version 1.0.13 to 1.0.15 on stable


MK-2001:
Closes https://github.com/MK-2001/ioBroker.go-e/issues/72

PR #1420, opened 2021-09-05
Bosch Smart Home Bridge Adapter stable 0.1.14

holomekc:
Bosch Smart Home Bridge Adapter stable 0.1.14


holomekc:
Changes from 28.08.2021:
https://github.com/holomekc/ioBroker.bshb/commits

PR #1433, opened 2021-09-16
Update CI testing

Smart-Gang:
Update CI testing from PR and update of dependecy version numbers


Apollon77:
Ok, for the future: just for testing and CI or dev dependency adjustments no new version is needed at all

PR #1477, opened 2021-10-13
Add xiaomi-gateway3 adapter

lasthead0:


lasthead0:
RE-CHECK!

GermanBluefox:
## Automated adapter checker

### [ioBroker.xiaomi-gateway3](https://github.com/lasthead0/ioBroker.xiaomi-gateway3)
[Downloads](https://www.npmjs.com/package/iobroker.xiaomi-gateway3) Number of Installations (latest) 
[NPM](https://nodei.co/npm/iobroker.xiaomi-gateway3/)

- [ ] :heavy_exclamation_mark: [E201] Bluefox was not found in the collaborators on [NPM](https://www.npmjs.com/package/iobroker.xiaomi-gateway3)!. Please execute in adapter directory: `npm owner add bluefox iobroker.xiaomi-gateway3`
- [ ] :eyes: [W400] Cannot find "xiaomi-gateway3" in latest repository


lasthead0:
RE-CHECK!

GermanBluefox:
## Automated adapter checker

### [ioBroker.xiaomi-gateway3](https://github.com/lasthead0/ioBroker.xiaomi-gateway3)
[Downloads](https://www.npmjs.com/package/iobroker.xiaomi-gateway3) Number of Installations (latest) 
[NPM](https://nodei.co/npm/iobroker.xiaomi-gateway3/)

- [ ] :heavy_exclamation_mark: [E201] Bluefox was not found in the collaborators on [NPM](https://www.npmjs.com/package/iobroker.xiaomi-gateway3)!. Please execute in adapter directory: `npm owner add bluefox iobroker.xiaomi-gateway3`
- [ ] :eyes: [W400] Cannot find "xiaomi-gateway3" in latest repository


lasthead0:
RE-CHECK!

lasthead0:
RE-CHECK!

GermanBluefox:
## Automated adapter checker

### [ioBroker.xiaomi-gateway3](https://github.com/lasthead0/ioBroker.xiaomi-gateway3)
[Downloads](https://www.npmjs.com/package/iobroker.xiaomi-gateway3) Number of Installations (latest) 
[NPM](https://nodei.co/npm/iobroker.xiaomi-gateway3/)

- [ ] :heavy_exclamation_mark: [E201] Bluefox was not found in the collaborators on [NPM](https://www.npmjs.com/package/iobroker.xiaomi-gateway3)!. Please execute in adapter directory: `npm owner add bluefox iobroker.xiaomi-gateway3`
- [ ] :eyes: [W400] Cannot find "xiaomi-gateway3" in latest repository


lasthead0:
what i do wrong? I ran command `npm owner add bluefox iobroker.xiaomi-gateway3`, i see bluefox  as collaborator in iobroker.xiaomi-gateway3 but checke still not. what i should to do?

Apollon77:
looks all correct ... the check sometimes seems to have a false positive

lasthead0:
I got it. thx!

lasthead0:
Hi. Should i do something else or you just have no enough time to review and merge?

Apollon77:
it's on me ... mega sorry ... hopefully in the course of next week

lasthead0:
that's ok. i understand. i just wanted to find out that i don't miss out anything. thx!

lasthead0:
I still hope to see my adapter in repository)) thx.

Apollon77:
yes, me too :-)) sorry again, will try soon

Apollon77:
Hi,
sorry that it took that long. here are the review comments :-)
* please add a common.globalDependencies to the io-package and add admin => 4.0.9 for it, else the encrypted native might not be supported
* please also add the info channel to https://github.com/lasthead0/ioBroker.xiaomi-gateway3/blob/master/io-package.json#L158 or is it created somewhere else?
* Using a sleep that way like https://github.com/lasthead0/ioBroker.xiaomi-gateway3/blob/master/lib/gateway3.js#L227 is problematic because if your adapter gets unloaded this timeout can not be killed. normally you only have 500ms (or you call the callback from unload) and short time after this the iobroker db is "gone"
* I think it could be a good idea to add an error handler on https://github.com/lasthead0/ioBroker.xiaomi-gateway3/blob/master/main.js#L119 too :-) else you might get crashes on error
* You also need to cleanup all connections and other stuff in unload. like your mqtt connection, pot telnet connection (if I got the code right) ... 

Thank you for checking and adjusting. Please ping back

lasthead0:
Hi!

I allmost done but have question about point 3.
yes, i understand that i can't manually kill timeout invoked in Promise, but i have to do sleep at some places, and i didn't find more suitable way then Promise. And honestly i don't clearly understand point of problem. Won't the timeout run even if the adapter shell get unloaded?  
Even if in some cases adapter will get unloaded until those timeouts running they are too short to takes a lot of resources.

Thx, and Happy New Year!

Apollon77:
> And honestly i don't clearly understand point of problem. Won't the timeout run even if the adapter shell get unloaded?

And exactly this is the issue. The timeout will stay running and then execute the next of your code ... but in this case the "databases" are long destroyed and each setState (example) will log an error ... 

So you could also set a "unloaded=true" and then simply add a "if (unloaded) return;" after the sleep ... You need t make sure your code stops after you called unload callback (or 500ms after unload was called even without callback)

lasthead0:
yes, more clear. so, if i clear timeout this should prevent execution code right after await sleep? this is a goal?
Is this a good idea to store timeouts ids in global var (array for example) and clear they on unload?

Apollon77:
Yes, clearing timeouts is one idea and several adapters do that with a global var. I think in your case the "challenge" is that you use "sleeop" several times so there might be more sleeps "running" at the same time, or?! But yes collecting all of them and clearing on unload is a good idea

Apollon77:
PS: maybe having an "unloaded=true" variable and using this in the "sleep" to not resolve is also an idea because then it would be "just" in one central place ...

lasthead0:
I guess i fixed all issues. I didn't publish package on npm yet and will do as soon as you say thats all ok. Issues fixes is in last 4 commits in master.

1. done
2. done
3. I added timeouts collecting to global var and do clear them on unload.
4. From README of mqtt library which i used i got that connections errors handled in mqtt library itself. Connect function don't throw error, so try...catch doesn't effect. But i added check for mqtt client on undefined in onUnload to prevent crash.
5. i added mqtt connection closing to onUnload. i think for telnet the same is unnecessary because opening and closing connection invoke for each separate "action" over telnet.

Apollon77:
Thank you very much. Please now also release the nnew version to npm and I will be happy to accept this PR :-)

lasthead0:
Happy to hear that!:)I just published new version 0.3.1 on npm.

PR #1487, opened 2021-10-19
Update sources-dist-stable.json: swiss-weather-api

baerengraben:
And another "bugfix" in the swiss-weather-api adapter, which caused the change in the SRF API. This version has been tested by several people and seems to be stable now. 
Now the adapter is worthy of version 1.0.0 *drum roll* :)


Apollon77:
wwwooohhooooo :-)

PR #1506, opened 2021-11-02
panasonic-comfort-cloud adapter added.

marc2016:


Apollon77:
RE-CHECK!

GermanBluefox:
## Automated adapter checker

### [ioBroker.panasonic-comfort-cloud](https://github.com/marc2016/ioBroker.panasonic-comfort-cloud)
[Downloads](https://www.npmjs.com/package/iobroker.panasonic-comfort-cloud) Number of Installations (latest) 
[NPM](https://nodei.co/npm/iobroker.panasonic-comfort-cloud/)

- [ ] :heavy_exclamation_mark: [E300] Not found on [travis](https://travis-ci.com/). Please setup [travis](https://travis-ci.com/)
- [ ] :eyes: [W400] Cannot find "panasonic-comfort-cloud" in latest repository


Apollon77:
Hi,

sorry that it took that long, here are review comments:
* Please update testing lib https://github.com/marc2016/ioBroker.panasonic-comfort-cloud/blob/master/package.json#L28 to prevent issues
* Please add "encryptedNative"seciond in io-package also for "password" and for this to work update js.controller dep to >= 2.0 and add a common.globalDependencies in same format with "admin >=4.0.9". Then the password will be encrypted and decrpted automatically - But this will be a breakting change for all current users!
* Please add refreshInterval with it's default value to native section in io-package
* Please rethink using a schedule to refresh the data, because this ends up in the fact that ALL users of this adapter will request (when staying with default interval which is a common case) pdates at the exact same timepoint from the panasonic servers - this can lead (with enough users) to high peaks and should be avoided! it is better to use at least an Interval which is then more spread - or even better to use a timeout which is set initially and then next is set after request has finished.
* Please only use Async methods with await, else use the callback style (also with ignoring callback) methods without Async
* Aree all these states with type "array" really correct? https://github.com/marc2016/ioBroker.panasonic-comfort-cloud/blob/master/src/main.ts#L286 is there really a stringified array as value in there? Additionally there might be better roles for some that have "state" right now--> https://github.com/ioBroker/ioBroker/blob/master/doc/STATE_ROLES.md (like value.temperature" for "read only temperatures"
* As optimization you could remember "readonly" state ids yourself because you created them before instead of always querying (https://github.com/marc2016/ioBroker.panasonic-comfort-cloud/blob/master/src/main.ts#L414) ... same for guid ... but yes works that way.

ldittmar81:
@marc2016 Have the suggestions been implemented and can we look at the project again or do you need some help? :-)

marc2016:
First of all, thank you very much for the review :-) Unfortunately, I haven't had the time to implement it, but I will start now!

marc2016:
- [ ] Please update testing lib https://github.com/marc2016/ioBroker.panasonic-comfort-cloud/blob/master/package.json#L28 to prevent issues

- [ ] Please add "encryptedNative"seciond in io-package also for "password" and for this to work update js.controller dep to >= 2.0 and add a common.globalDependencies in same format with "admin >=4.0.9". Then the password will be encrypted and decrpted automatically - But this will be a breakting change for all current users!

- [ ] Please add refreshInterval with it's default value to native section in io-package

- [ ] Please rethink using a schedule to refresh the data, because this ends up in the fact that ALL users of this adapter will request (when staying with default interval which is a common case) pdates at the exact same timepoint from the panasonic servers - this can lead (with enough users) to high peaks and should be avoided! it is better to use at least an Interval which is then more spread - or even better to use a timeout which is set initially and then next is set after request has finished.

- [ ] Please only use Async methods with await, else use the callback style (also with ignoring callback) methods without Async

- [ ] Aree all these states with type "array" really correct? https://github.com/marc2016/ioBroker.panasonic-comfort-cloud/blob/master/src/main.ts#L286 is there really a stringified array as value in there? Additionally there might be better roles for some that have "state" right now--> https://github.com/ioBroker/ioBroker/blob/master/doc/STATE_ROLES.md (like value.temperature" for "read only temperatures"

- [ ] As optimization you could remember "readonly" state ids yourself because you created them before instead of always querying (https://github.com/marc2016/ioBroker.panasonic-comfort-cloud/blob/master/src/main.ts#L414) ... same for guid ... but yes works that way.

Apollon77:
@marc2016 ready for re-review?

marc2016:
Sorry, the last point is still missing. I'll try to get it done this weekend.

Apollon77:
all fine, just poke here when you are done ...

marc2016:
@Apollon77 ready for re-review! :-)

Apollon77:
Hi @marc2016 just some small stuff left :-)

* https://github.com/marc2016/ioBroker.panasonic-comfort-cloud/blob/master/src/main.ts#L313-L322 ... type needs to be number and not string for such cases like this ...please also check the others

Else it looks fine

marc2016:
Hi @Apollon77 ,

I have also corrected this point. You can look over it again :-)

Many thanks again for the review.

PR #1553, opened 2021-12-20
add openknx

TA2k:
add free open knx alternative

### DIFF
--- a/sources-dist.json
+++ b/sources-dist.json
@@ -1236,6 +1236,11 @@
     "meta": "https://raw.githubusercontent.com/iobroker-community-adapters/ioBroker.openhab/master/io-package.json",
     "icon": "https://raw.githubusercontent.com/iobroker-community-adapters/ioBroker.openhab/master/admin/openhab.png",
     "type": "iot-systems"
+  },  
+  "openknx": {
+    "meta": "https://raw.githubusercontent.com/iobroker-community-adapters/ioBroker.openknx/master/io-package.json",
+    "icon": "https://raw.githubusercontent.com/iobroker-community-adapters/ioBroker.openknx/master/admin/openknx.png",
+    "type": "iot-systems"
   },
   "openligadb": {
     "meta": "https://raw.githubusercontent.com/oweitman/ioBroker.openligadb/master/io-package.json",


TA2k:
RE-CHECK!

GermanBluefox:
## Automated adapter checker

### [ioBroker.openknx](https://github.com/iobroker-community-adapters/ioBroker.openknx)
[Downloads](https://www.npmjs.com/package/iobroker.openknx) Number of Installations (latest) 
[NPM](https://nodei.co/npm/iobroker.openknx/)

:thumbsup: No errors found
- [ ] :eyes: [W400] Cannot find "openknx" in latest repository


Apollon77:
Here some review comments:
* your build step seems broken :-)
* I'm very unsure about the knx.0 override feature because this breaks completely the namespace independence of adapters and instances. Also when you write that objects and structures are incompatible, what this would bring? Also pot data formats and such when thinking about historization will pot. be incompatible. My personal opinion is that we should not start with this. If a user wants to move over from knx.0 to your adapter then he needs to adjust the references (or start using aliases that are exactly there fur such things).
* ... min/max as number for a boolean state? https://github.com/iobroker-community-adapters/ioBroker.openknx#iobroker-communication-object-description ... but when I read code it would end up this way potentially. Please makse sure to only use min/max with number states tht need it. not for boolean.
* You do not need to list yourself as restartAdapters https://github.com/iobroker-community-adapters/ioBroker.openknx/blob/master/io-package.json#L73
* https://github.com/iobroker-community-adapters/ioBroker.openknx/blob/master/main.js#L225 whats that for? normally you use ack=true for own values and ignore it that way. Misusing the comment is not a best practice?
* You use the knx package, but also have a version included? Strange construction ... is it a patched version? ... but ok if you want to do it that way

Please check and adjust.

boellner:
@Apollon77 
point 6: I provide a patched version of knx, i also require it to resolve dependencies; the mid term solution is to have the knx implement my features and bugfixes

TA2k:
* build is fixed
* it was idea for advanced user to reduce the effort to change the vis without creating alias. Because the naming is identical only the extra properties are different. You can easily switch between the adapters while deleting and reimport. Can we give it a try to receive feedback or a no go?
* we will fix the min max for boolean
* restartAdapters removed
* we will check if state.ack is also working
* it is adjusted version for the project import

Apollon77:
For such a situation I would propose an own fork of knx lib with your changes and use this as dependency. All iobroker installations should have git installed already because installer do install it. Easier to manage also for fixes and such in my eyes and you need that to track the PR to upstream project anyway.

Alternative is to publish an own scoped package and reference this. 

So as "best practice" ... will not be a decline reason for repo tho ? But nicer. :-)

Apollon77:
@TA2k for the namespace override I personally would not like to have it. It might be compatible now but no one knows about in a day. There will be pot dead objects or or or. Unthinkable it is cleaner to have it separated. I will have discussion in core group about it

boellner:
Point 2:
I have a lot of consumers of knx.0 objects. Both adapters create an identical object tree, except the adapter name.
I delete the knx.0 tree and let openknx generate it, with its own object attributes. Major transmission and reception interfaces are equal, so that both adapter become exchangeable with each other.
With this approach I could easily exchange the underlying knx communication of a complex setup.

foxriver76:
https://github.com/iobroker-community-adapters/ioBroker.openknx/blob/c383547de3a3bc84cd3e87ef3153e880090c6ce2/io-package.json#L65
https://github.com/iobroker-community-adapters/ioBroker.openknx/blob/c383547de3a3bc84cd3e87ef3153e880090c6ce2/main.js#L351
shouldn't it be push?

Apollon77:
@boellner @TA2k We discussed the idea to "take over an other namespace" in the core group and we do not want to start this. Each adapter own it's namespace, structures and such and this is an important paragigm in iobroker and we do not want to change that.

I clearly understand your idea for "user convenience" in this regard but the issues this can cause - especially on the long run - are bigger. This also includes the "need" for you to mirror any change from knx (closed source) project from now on ... I think this makes simpyl no sense at all. 
Additionally we introduced Aliases as way to abstract exactly this interoperability to have only one place to define which adapter is responsible. Yes this is still relatively new but in our eyes this is the way to go to reach such an interoperability and yes for users it is an effort one time and will "only" help in the future.

When it comes to vis or javascripts there are easy ways to do "mass replacements" when doing such a switch to Alias or between adapters like in this case.

Please remove this option from the adapter and that you for respecting our decision.

AlCalzone:
Not sure it was mentioned, but taking over the foreign namespace has some more practical implications:
- Uninstalling the knx adapter will delete the tree along with all objects openknx has created there
- On the other hand, uninstalling openknx will **not** delete the objects it has created under the knx namespace.

This is just chaos.

boellner:
please re-review
everything except 
https://github.com/iobroker-community-adapters/ioBroker.openknx/issues/1
remove static inclusion of knx library
should have been fixed now

Apollon77:
Ok, had another look,
* you could also use the "Local" variants from the object methods (https://github.com/iobroker-community-adapters/ioBroker.openknx/blob/master/main.js#L131 ...) and someother state calls - but up to you.
* One question on https://github.com/iobroker-community-adapters/ioBroker.openknx/blob/master/main.js#L128-L145 ... I do not really get the difference? extend will also create the object if it is not existing and "fully defined", else will "update" it - but this requires controller 3.2 to be really 100% accurate. In order to reach what I think you want here you would need to add a "getObject" and according to result (object is returned or not) you wuld know if the object exists or not ...

Apollon77:
Please verify that the second part is really "how you want it" ... then we can add it to the repo

boellner:
@Apollon77 
Point 1: forneign can be removed eveywhere...
Point 2: 
With not set onlyAddNewObjects:
the setObject overwrites all user made changes to the object before. Its a fresh installation, loosing previous settings.

With set onlyAddNewObjects:
Use case is user adds new GAs in his ETS, now wants to import them, keeping his changes to other object.
Extend for safety, if adapter has new defines.

I tested it and remember it worked like this. Please let me know If this design makes sense to you

Apollon77:
The issue with "setObject" is - especially in controller <3 - that sometimes also custom settings to the objects like "history adapter settings" are overwritten ... in other versions of controller also on setObject for the "overwrite case" the cistom settings are automatically preserved toprevent exactly such situations. I'm not sure which other "user made changes" you refer to in your infos above.

Also as said: extendObject will also add objects when they are fully defined in the call, so also an extend on an non existing object wouldcreate this one - And will override all data that are in the structure and so updates the objects. Depending on which "user made changes" you refer to it should end up in kind of the same results (assuming a controller 3.2+ ... in older versions extend could behave slightly different)

If the "user made" cahanges are about "additional data in the native section of the object" then these would be preseved on extend and removed on setObject - this would be right

TA2k:
I think preserving values is only possible with a custom method
1. GetObject
2. Check for different values than the importer would write
3. Preserve differents.

For now we should stick with not exist and can improve this maybe later.

@Apollon77  All issues should now be fixed.

Apollon77:
@TA2k @boellner I'm a bit confused now ... I see opne repo https://github.com/iobroker-community-adapters/ioBroker.openknx and one repo here https://github.com/boellner/ioBroker.openknx and none is a fork or such ... whats the main repo now? 

Else please publish the version with the adjustemnts to NPM and then we can continue with the merge. That you for your support.

TA2k:
https://github.com/iobroker-community-adapters/ioBroker.openknx 

the other will be deleted tomorrow

boellner:
my initial repo https://github.com/boellner/ioBroker.openknx is now archieved; 

main repo https://github.com/iobroker-community-adapters/ioBroker.openknx 
is connected to NPM https://www.npmjs.com/package/iobroker.openknx
and rel. 0.1.8 was made from there; 
(Dont know why NPM references the old repo, package is updated, need to check tomorrow)

Apollon77:
0.18 is from 2 days ago, maybe it was the last push before all the changes (that happend in last two days?) ;-) Sois this really the new version with all stuff discussed/adjusted here?

boellner:
you're right, i have created a new release 0.1.9 from here, links are now correct and doc is up to date

Apollon77:
oooohhh .... I saw one last thing (sprry for not noticing earlier)

We have a type mismatch in io-ppackage vs repo PR. I think sgouldbe both iot-systems or protocols (with first one should be the correct) ... so io-package in repo needs to be adjusted :(

TA2k:
Is fixed

boellner:
not 100% sure what type means, but isnt protocol better since it is definitely not a iot system?

Apollon77:
The Type is used in Admin to categorize the adapters ... knx is also in iot-systems .... protocols is more fpr "low level protocols" like mqtt or such that can be anything or nothing. When the "protocol" already structures stuff and knows devices then "iot systems "is better

PR #1579, opened 2022-01-06
add wiobrowser

Bettman66:
control a windows fullscreenbrowser


ldittmar81:
@Bettman66 before you add it to stable, you should add it to latest - sources-dist.json - and after a while, it can be added to stable

Bettman66:
I can do that, but the adapter has been in use for 1 month without errors by 2 users and me.

PR #1588, opened 2022-01-09
Update HeyTech Adapter for nodejs 14 compatibility

ansgarschulte:


Apollon77:
Is this proofed enough to already update stable? Additionally the adapter testing is currently not in place in the reposiotory

Apollon77:
Ok, version is 8 days old and have "enough" installations ... so fine, but Please reacrtivate adapter etsting before the next version bump!

PR #1598, opened 2022-01-12
ems-esp adapter please add to repository

tp1de:
So finally I hope that everything is ok now ....

### DIFF
--- a/sources-dist-stable.json
+++ b/sources-dist-stable.json
@@ -369,7 +369,7 @@
     "meta": "https://raw.githubusercontent.com/tp1de/ioBroker.ems-esp/master/io-package.json",
     "icon": "https://raw.githubusercontent.com/tp1de/ioBroker.ems-esp/master/admin/ems-esp.png",
     "type": "climate-control",
-    "version": "1.0.2"
+    "version": "1.0.4"
   },
   "energymanager": {
     "meta": "https://raw.githubusercontent.com/unltdnetworx/ioBroker.energymanager/master/io-package.json",
--- a/sources-dist-stable.json
+++ b/sources-dist-stable.json
@@ -377,7 +377,7 @@
     "type": "multimedia",
     "version": "1.0.2"
   },
-    "ems-esp": {
+  "ems-esp": {
     "meta": "https://raw.githubusercontent.com/tp1de/ioBroker.ems-esp/master/io-package.json",
     "icon": "https://raw.githubusercontent.com/tp1de/ioBroker.ems-esp/master/admin/ems-esp.png",
     "type": "climate-control",
--- a/sources-dist-stable.json
+++ b/sources-dist-stable.json
@@ -2181,5 +2181,11 @@
     "icon": "https://raw.githubusercontent.com/AlCalzone/ioBroker.zwave2/master/admin/zwave2.svg?sanitize=true",
     "type": "hardware",
     "version": "2.4.1"
+  },
+    "ems-esp": {
+    "meta": "https://raw.githubusercontent.com/tp1de/ioBroker.ems-esp/master/io-package.json",
+    "icon": "https://raw.githubusercontent.com/tp1de/ioBroker.ems-esp/master/admin/ems-esp.png",
+    "type": "climate-control",
+    "version": "0.9.9"
   }
 }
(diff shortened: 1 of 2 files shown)


Apollon77:
RE-CHECK!

ldittmar81:
[W505] setTimeout found in "main.js", but no clearTimeout detected

https://github.com/tp1de/ioBroker.ems-esp/blob/6a090c91b2d9923ee499eca91738d7b8c6cf7e01/main.js#L40
https://github.com/tp1de/ioBroker.ems-esp/blob/6a090c91b2d9923ee499eca91738d7b8c6cf7e01/main.js#L320

You have to assign your timer to a variable and clear it on unload.

tp1de:
> [W505] setTimeout found in "main.js", but no clearTimeout detected

We had a long discussion in Discord that this is not needed a week ago !

tp1de:
image

ldittmar81:
Please close https://github.com/ioBroker/AdapterRequests/issues/302 when done

tp1de:
RE-CHECK!

ldittmar81:
@Apollon77 Auto Review done - no error - RE-CHECK doesnt works when more tha one file is changed.

Apollon77:
@tp1de Here is my review feedback:

* One last think for the "unload callback timeout". If the callback is not called within 500ms then the unload will continue, so this 1000ms will be simply "ignored" unless you set in io-package "common.stopTimeout" to e.g. 1500 to make sure that your callback is called before the default timeout from the system ... So it only makes sense together
* Why you do this that way? https://github.com/tp1de/ioBroker.ems-esp/blob/6a090c91b2d9923ee499eca91738d7b8c6cf7e01/main.js#L50 ... state && !state.ack would be "the way to go".
* console.log will not be visible to the user ... intended? https://github.com/tp1de/ioBroker.ems-esp/blob/6a090c91b2d9923ee499eca91738d7b8c6cf7e01/main.js#L122
* The encrypt/decrypt could be easier with no code, but this is just convenience and is nothing you need to do :-) (btw you miss the fallback secret when decrypting ...)
* You told in last review that per host the adapter should be only there once? YOu can force that as told by adding "common.singleton: true" in io-package
* If you like you can also set https://github.com/ioBroker/adapter-core/blob/master/README.md#automatic-backup-of-data-files to get your files also backed up if relevant (optional)

Only first two topic are the relevant ones. Please have another look there. Rest are proposals from my sidecto have less issues with users. Thank you for checking it again :-)

tp1de:
@Apollon77  Ingo thanks for feedback. I changed:
1. I reduced the "sleeps" to a minimum and all below 300 msec and added common.stopTimeout
2. I changed to state && !state.ack (I had some setStates without ack = true)
3. console.log was a mistake. Corrected
4. encrypt/decrypt: I do not understand what I should change
5. I made the adapter to be used in multi-instances 
6. I added the folder for backup.

If everything is fine for you I will generate a new npm (new version)

Apollon77:
Please release and poke here, all is fine. Thank you for your support!

For encrypt decrypt: You do it "manually old style" :-) In the meantime you can simplify it 
* you can remove all the manual encrypt oder decrypt functions or code in your adapter and also admin html file
* you add the relevant field like here https://github.com/TA2k/ioBroker.psa/blob/master/io-package.json#L84-L89
* you add depenedncies for controller and admin to these values (at least) https://github.com/TA2k/ioBroker.psa/blob/master/io-package.json#L73-L82
And then all encryption and decryption is handled by Admin and the controller 100% automatically ... you can simply work with the values.

Also the "old style way" works, but this is one of these "we saw the pain of the devs and made it better" cases :-)

just poke here when npm is done and I'm more then happy to merge the PR

tp1de:
@Apollon77 I release new version 1.0.5 manually. I added the release script and tried to use it on www.iobroker.dev. It was running without errors but did not change anything.
I need to find the reason ...

Please merge the PR and thanks for support.

Apollon77:
The reason why your deploy do not work is that the release script is not fully integrated with the 3.x

you basically need:
* https://github.com/Apollon77/ioBroker.tuya/blob/master/.releaseconfig.json this file
* add these lines https://github.com/Apollon77/ioBroker.tuya/blob/master/package.json#L43-L44
then it should work again

PR #1610, opened 2022-01-19
adapter huum-sauna added

Chris-656:
Adapter for sauna control using HUUM device

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,9 +1,6648 @@
 {
   "name": "iobroker.repositories",
   "version": "1.0.4",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
   "requires": true,
+  "packages": {
+    "": {
+      "name": "iobroker.repositories",
+      "version": "1.0.4",
+      "dependencies": {
+        "request-promise-native": "^1.0.9"
+      },
+      "devDependencies": {
+        "@alcalzone/release-script": "^3.5.2",
+        "axios": "^0.23.0",
+        "chai": "^4.3.4",
+        "eslint": "^8.0.1",
+        "gulp": "^4.0.2",
+        "image-size": "^1.0.0",
+        "minimist": "^1.2.5",
+        "mocha": "^9.1.3",
+        "node.extend": "^2.0.2",
+        "request": "^2.88.2",
+        "request-promise-native": "^1.0.9",
+        "semver": "^7.3.5",
+        "unzipper": "^0.10.11"
+      },
+      "engines": {
+        "node": ">=12"
+      }
+    },
+    "node_modules/@alcalzone/pak": {
+      "version": "0.8.0",
+      "resolved": "https://registry.npmjs.org/@alcalzone/pak/-/pak-0.8.0.tgz",
+      "integrity": "sha512-ExytxVXzXsrRWLf35BC+EFJIDoRUB42mfHeczzOyyQLR8RjkMA86Em3WfbA+OyN0vd06XqSHr4JOMphIc/GnXw==",
+      "dev": true,
+      "dependencies": {
+        "axios": "^0.21.1",
+        "execa": "^5.0.0",
+        "fs-extra": "^9.1.0"
+      }
+    },
+    "node_modules/@alcalzone/pak/node_modules/axios": {
+      "version": "0.21.4",
+      "resolved": "https://registry.npmjs.org/axios/-/axios-0.21.4.tgz",
+      "integrity": "sha512-ut5vewkiu8jjGBdqpM44XxjuCjq9LAKeHVmoVfHVzy8eHgxxq8SbAVQNovDA8mVi05kP0Ea/n/UzcSHcTJQfNg==",
+      "dev": true,
+      "dependencies": {
+        "follow-redirects": "^1.14.0"
+      }
+    },
+    "node_modules/@alcalzone/pak/node_modules/fs-extra": {
+      "version": "9.1.0",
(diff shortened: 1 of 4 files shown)


Apollon77:
RE-CHECK!

Apollon77:
@Chris-1965 Please update this PR and sort in your adapter alphabetically :-)

GermanBluefox:
## Automated adapter checker

### [package.json](https://github.com/main/package.json)
[Downloads](https://www.npmjs.com/package/package.json) 
[NPM](https://nodei.co/npm/package.json/)

- [ ] :heavy_exclamation_mark: 404: Invalid URL


Apollon77:
Here also additional review comments:

* Please encrypt the stored password. For instructions how to use the auto encrption feature please check the best practices in the readme  of this repo
* please fix the testing https://github.com/Chris-1965/ioBroker.huum-sauna/runs/4878813294?check_suite_focus=true
* Reason is that you need to add all your required npm modules to the package.json in "dependencies" (e.g.https://github.com/Chris-1965/ioBroker.huum-sauna/runs/4878813294?check_suite_focus=true#step:2:394) Additionally to suncalc2 this should also be axios
* Please only react on stae Changes with ack=== false because ack= true are no commands in https://github.com/Chris-1965/ioBroker.huum-sauna/blob/main/main.js#L282
* Maybe it is a good idea to defeine a minimum value for the interval to prevent that someone tries to ask the cloud server every second or such (also because it will lead to bigger issues if a request would take longer then the interval time because then you start to DOS the cloud server. In general "Intervals" are problematic when it comes to external server communications because they ignmore timeouts (and also you did not define any timeout to the axios request). See bet practices!
* Please check all object roles in https://github.com/Chris-1965/ioBroker.huum-sauna/blob/main/io-package.json#L129 and following and get correct values from https://github.com/ioBroker/ioBroker/blob/master/doc/STATE_ROLES.md

Thank you for considering these comments and adjust accordingly

Chris-1965:
> PR

Done

Chris-1965:
I corrected all issues hopefully created a new version 0.1.5 
regards Chris

Apollon77:
Puuhhh ... your PR did move around many thing ... wtf? Can you please recreate it

Apollon77:
Hi small things left,

* for encryption you chose the most complex way ... I think the readme should have pointed you to the auto encrypt feature , or? But ok ... your decision :-))
* Please only react on stae Changes with ack=== false because ack= true are no commands in https://github.com/Chris-1965/ioBroker.huum-sauna/blob/main/main.js#L322 ... or is there any reason to also process ack=true?
* I would simply use value here or such ... https://github.com/Chris-1965/ioBroker.huum-sauna/blob/main/io-package.json#L234 ... media is not really matching, or?
*

Apollon77:
For Encryption ... hm yes I need to update the specs there ... meross is not really best anymore :-)
In fact:
* add the fields that way https://github.com/TA2k/ioBroker.psa/blob/master/io-package.json#L84-L89
* add these deps at least https://github.com/TA2k/ioBroker.psa/blob/master/io-package.json#L73-L82
* remove all custom encrypt/decrypt stuff

I do not understand what you write above for state Change. Normally the adapter sets the "real values" with ack=true. ack=false is used from outside to control values and tell the adapter to control something

Chris-1965:
> For Encryption ... hm yes I need to update the specs there ... meross is not really best anymore :-) In fact:
> 
>     * add the fields that way https://github.com/TA2k/ioBroker.psa/blob/master/io-package.json#L84-L89
> 
>     * add these deps at least https://github.com/TA2k/ioBroker.psa/blob/master/io-package.json#L73-L82
> 
>     * remove all custom encrypt/decrypt stuff
> 
> 
> I do not understand what you write above for state Change. Normally the adapter sets the "real values" with ack=true. ack=false is used from outside to control values and tell the adapter to control something

ok I will change to new encrypt version.

Regarding the state change:
When I switch on sauna I want to switch on the light after sunset automatically. so I use 
``` 
this.setState("switchLight", state.val, true);
```
in the code as usual. When I trigger the light switch from vis or state from objects iobroker that will we done with ack===false.
So it would not be possible to swicth the light in the adapter unless i use
```
this.setState("switchLight", state.val, false);  // line 335
```

Apollon77:
If the adapter kind of controls itself then why not send an ack=false comand to itself ... then you only ned the logic once and same logic is used as from user ... this would be the clean way ...

Chris-1965:
Updated all comments and fixed them
reagrds Chris

Chris-1965:
created new version and tried to fix issues
regards Chris

Chris-1965:
"RE-CHECK!"

Apollon77:
RE-CHECK!

Chris-1965:
RE-CHECK!

Chris-1965:
Is there anything for me to do here ?

ldittmar81:
@Chris-1965 no. It's our turn to check the adapter.

Apollon77:
Cool, looks good, welcome to repository!

Apollon77:
BUT, please fix/recreate the PR because you now added the release script to the repo in your PR??? @Chris-1965

Apollon77:
@Chris-1965 better recreate the PR ...

Chris-1965:
> @Chris-1965 better recreate the PR ...

you mean close the PR and create a new one ?

Apollon77:
@Chris-1965 maybe that the best idea ... simply refeence this PR then we have a connection

Chris-1965:
creating new PR

PR #1682, opened 2022-02-18
new adapter ioBroker.minuaru

SVallant:
Result of Adapter Checker:

- [E506] "admin/jsonCustom.json" not found, but custom support is declared
- [E300] Not found on travis. Please setup travis

both false positive:

- jsonCustom.json is there
- please add travis.com to checker together with travis.org

### DIFF
--- a/sources-dist.json
+++ b/sources-dist.json
@@ -39,15 +39,15 @@
     "icon": "https://raw.githubusercontent.com/misanorot/ioBroker.alarm/master/admin/alarm.png",
     "type": "alarm"
   },
-  "alexa2": {
-    "meta": "https://raw.githubusercontent.com/Apollon77/ioBroker.alexa2/master/io-package.json",
-    "icon": "https://raw.githubusercontent.com/Apollon77/ioBroker.alexa2/master/admin/alexa.png",
-    "type": "iot-systems"
-  },
   "alexa-timer-vis": {
     "meta": "https://raw.githubusercontent.com/MiRo1310/ioBroker.alexa-timer-vis/main/io-package.json",
     "icon": "https://raw.githubusercontent.com/MiRo1310/ioBroker.alexa-timer-vis/main/admin/alexa-timer-vis.png",
     "type": "logic"
+  },
+  "alexa2": {
+    "meta": "https://raw.githubusercontent.com/Apollon77/ioBroker.alexa2/master/io-package.json",
+    "icon": "https://raw.githubusercontent.com/Apollon77/ioBroker.alexa2/master/admin/alexa.png",
+    "type": "iot-systems"
   },
   "alias-manager": {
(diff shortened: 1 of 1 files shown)


Apollon77:
RE-CHECK!

GermanBluefox:
## Automated adapter checker
No changed adapters found


SVallant:
Hi @Apollon77 

here some review comments:

 _Please add adapter testing also to the travis job (or better go to Github Actions?) - right now only package testing is done_
yes, is planned before adding to stable. Can you give me some example of testing admin/custom.json ?

_Is the adapter type storage really making sense_ 
for me it is the same like "ioBroker.history", storing and displaying alarm-messages
What other type would you suggest ???

_Please remove admin from the "dependencies" list_ , _minimum js-controller 3.3 as dependency_
 done

data-directory ( now .../iobroker-data/minuaru.0) changed, please check if config 
https://github.com/minukodu/ioBroker.minuaru/blob/bde7c341d48c60561fdce6d27f10b785feae7ed5/io-package.json#L127
is correct for backup

Thank you for this good support

Apollon77:
@SVallant 

1.) Testing custom ... in fact you can create an object with the relevant custom entries before adapter get started and such. In fact we have very detailed testing for history and such (eg. https://github.com/ioBroker/ioBroker.history/blob/master/test/testAdapter.js#L100-L116), but this uses the "legacy testing scripts" and not the new testing lib. But basically "the same" is possible there too...
2.) I would see it more "alarm" or "misc-data" dependeing on what you want to focus on. I think the main focus is more about "doing something based on actions" and not "dump storage of data"
3.) Backup/Data Dir: seems fine.. maybe test it :-) you can check that it restored on "restore" or that it is contained in the backup file

Apollon77:
Additionally for testing: even if you do not add real custom testing it is better then nothing to just proof that install and start basically works. But yes you can also go deep into functional testing ;-)

SVallant:
adapter's type is "misc-data" now, since alarm-adpters or mostly "security"-Adapters
Backup works like you said 
wirting tests will be a big challenge next weeks ... ;-)

Apollon77:
For tests: As said: For the repo addition it would be enought to just see adater starting basically (so execute the integration tests - done). More is optinal after that

Apollon77:
@SVallant Anything we can support with? Just poke when ready for re-review.

SVallant:
No, thanks, just no available time at the moment .... other projects are more important

Apollon77:
@SVallant if it is only about the testing thenthis would just be requirement that "it runs and starts the adapter", custom testing is formally not required

SVallant:
@Apollon77 Adapter is ready for latest now. Do you a new PR?

Apollon77:
resolve the conlfucts or do a new PR (but please add a linkto here) ... up to you :-)

PR #1703, opened 2022-02-27

sbormann:
IQontrol 2.0.0

### DIFF
--- a/sources-dist-stable.json
+++ b/sources-dist-stable.json
@@ -867,7 +867,7 @@
     "meta": "https://raw.githubusercontent.com/sbormann/ioBroker.iqontrol/master/io-package.json",
     "icon": "https://raw.githubusercontent.com/sbormann/ioBroker.iqontrol/master/admin/iqontrol.png",
     "type": "visualization",
-    "version": "1.8.2"
+    "version": "2.0.0"
   },
   "jarvis": {
     "meta": "https://raw.githubusercontent.com/Zefau/ioBroker.jarvis/master/io-package.json",


Apollon77:
How compatible is this version to 1.x ? I ask because e.g. on free/pro cloud server there will still be the 1.x frontend that then communicates with  an 2.x backend ... will that work? If not you might want to add an info for the users?

sbormann:
Should work, but lacks the new features. What is necessary to update the cloud version?

Am 27.02.2022 um 12:38 schrieb Ingo Fischer ***@***.***>:

﻿

How compatible is this version to 1.x ? I ask because e.g. on free/pro cloud server there will still be the 1.x frontend that then communicates with an 2.x backend ... will that work? If not you might want to add an info for the users?

Apollon77:
Updating cloud version best create an issue in the iobroker.cloud repo but normally we do that once the version is longer in stable ... because when we update cloud version the above question reverses: What breaks in FE is 2.x and Backend still 1.x ... so in fact you need to find a balance :-)
But when the old FE wil the new backend works then updating stable without info is easy ... 

Wasjst a question whats better for the user communication

sbormann:
Ok, thanks for explaining me that. i think we should update cloud version in a few days. Most of old FE is compatible with new BE. Just a few option will have no function, but it won’t break it.

Am 27.02.2022 um 12:59 schrieb Ingo Fischer ***@***.***>:

﻿

Updating cloud version best create an issue in the iobroker.cloud repo but normally we do that once the version is longer in stable ... because when we update cloud version the above question reverses: What breaks in FE is 2.x and Backend still 1.x ... so in fact you need to find a balance :-)
But when the old FE wil the new backend works then updating stable without info is easy ...

Wasjst a question whats better for the user communication

Apollon77:
Thank you

PR #1722, opened 2022-03-13
Add Apple-Find-Me Adapter

PfisterDaniel:
Add Apple-Find-Me Adapter and remove Find-My-Iphone Adapter from Apollon77

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2427,12 +2427,6 @@
       "integrity": "sha1-peZUwuWi3rXyAdls77yoDA7y9RM=",
       "dev": true
     },
-    "json-schema": {
-      "version": "0.2.3",
-      "resolved": "https://registry.npmjs.org/json-schema/-/json-schema-0.2.3.tgz",
-      "integrity": "sha1-tIDIkuWaLwWVTOcnvT8qTogvnhM=",
-      "dev": true
-    },
     "json-schema-traverse": {
       "version": "0.4.1",
       "resolved": "https://registry.npmjs.org/json-schema-traverse/-/json-schema-traverse-0.4.1.tgz",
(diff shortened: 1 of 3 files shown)


Apollon77:
RE-CHECK!

GermanBluefox:
## Automated adapter checker
No changed adapters found


Apollon77:
@PfisterDaniel Can you please redo/adjust this PR? Please do not change package or package-lock

Additionally here my review feedback:
* Please fix tests .. they are red - i thinl you need to commit the package-lock.json also; also update testing lib, else will also fail: https://github.com/PfisterDaniel/ioBroker.apple-find-me/blob/master/package.json#L31
* Hint: You can remove all that manual encrypt/decrypt code if you use the automatic handling which exists since js-controller 3 and admin 4.0.9+ ... see readme of this readme. But works also zthe way you did it for now
* Please consider to change from node_schedule to an interval or better timeout (https://github.com/PfisterDaniel/ioBroker.apple-find-me/blob/master/main.js#L586). The reason is that when doing this the start time of the adapter defines the interval which spreads calls over time better then a hard schedule that runs on second 0 on the given minutes (by default every 5 minutes ... for all users of the adapter ... this can generate request peaks and can give issues (even if the servers are the ones from Apple :-) )
* Please remember if objects were already created and do not usse "setObjectNotExists" every x mins ... this is just unneeded database load
* https://github.com/PfisterDaniel/ioBroker.apple-find-me/blob/530cd05d94e3c2c4d03696318a63b521967ec8aa/main.js#L160-L174 such constructs will generate warn messages on first object creations because you execute the setState before the objects exists (because created async. So eigther do setState in callback or use asyn/await please
* please check the roles of your states for better options (e.g. value.battery or value.gps.latitude ...) https://github.com/ioBroker/ioBroker/blob/master/doc/STATE_ROLES.md ; also please store dates better as role date (see also state riole page for specs)
* unload MUST stop the schedule! https://github.com/PfisterDaniel/ioBroker.apple-find-me/blob/530cd05d94e3c2c4d03696318a63b521967ec8aa/main.js#L47
* please remove https://github.com/PfisterDaniel/ioBroker.apple-find-me/blob/530cd05d94e3c2c4d03696318a63b521967ec8aa/main.js#L62-L70 if not needed

Apollon77:
I hope closed to recreate and not because shocked by feedback? ;-)

PfisterDaniel:
Hi,

no I’m not shocked, I will fix all issues and then I recreate the PR.

Br Daniel

Von: Ingo Fischer ***@***.***>
Datum: Montag, 14. März 2022 um 11:07
An: ioBroker/ioBroker.repositories ***@***.***>
Cc: Daniel Lippert ***@***.***>, State change ***@***.***>
Betreff: Re: [ioBroker/ioBroker.repositories] Add Apple-Find-Me Adapter (PR #1722)

I hope closed to recreate and not because shocked by feedback? ;-)

Apollon77:
oK, fine, please do me a favour and link to this PR ... so I remember my review comments :-) Safes me some time for the re-review

PR #1796, opened 2022-05-04
Add bidirectional-counter to latest

BenAhrdt:
Please add my adapter ioBroker.bidirectional-counter to latest.


Apollon77:
RE-CHECK!

GermanBluefox:
## Automated adapter checker
No changed adapters found


Apollon77:
Hi,

here some review comments:
* [E145] No news found for actual version 1.12.8
* [E201] Bluefox was not found in the collaborators on NPM!. Please execute in adapter directory: "npm owner add bluefox iobroker.bidirectional-counter"
* [E506] "admin/jsonCustom.json" not found, but custom support is declared
* Please add the usual badges to Readme
* Remove the WORK IN PROGRESS from iitial release ... this will make issues with the release scriüt
* If you only support jsonconfig yopu bneed to add common.globalDependencies to io-package with admin >= 5.1.0 and https://github.com/BenAhrdt/ioBroker.bidirectional-counter/blob/main/io-package.json#L118 should be false
* I think a connection identifier makes no sense, because you do not connect to anything. Please remove
* def 0 is only allowed if the type is number ... but your type seems to be dynamic?? Then def might be wrong https://github.com/BenAhrdt/ioBroker.bidirectional-counter/blob/main/main.js#L151 
* When you subscribe to states consider >100 states that it is performance wise better to do a subscribe *
* Do you write back a change into the "foreign" namespace??? https://github.com/BenAhrdt/ioBroker.bidirectional-counter/blob/main/main.js#L295 with ack=true. Please NEVER do that!! Only the adapter that owens this state should do that

Please check and adjust

BenAhrdt:
1. Add news: => ok, i will do
2. Add Bluefox: => ok, i will do.
3. jsonCustom is there, why you can´t found it?
4. Readme => i will check
5. Work in progress, i will delete
6. materialize = flase ?????
7. i use the connection identifier for indicate that the init of the adapter was correctly. is it wrong?
8. i have to remove identifier?
9. its just possible to configure states with type number, so i will change typ into number
10. i subscribe only the custom configed states. so i thing that these arte less than 100
11. i will check it with setState (But the state i write are only adapter internal states.
12. i just used setForeignState because i think in case of setState i have to substring the id (delete namespace)

Garfonso:
3: I think this is an error in adapter checker. admin/jsonCustom.json is in repository.

6: materialize is the default UI in admin 4. If you only support admin 5 jsonConfig, then materialize is not there, so do not set it to true. Makes sense. Not? There is no index_m.html in your admin directory, which would be used for materialize.

7/8: That is not the meaming of info.connection. It means the connection to a device / service is working. If not, adapter will still be running, but admin indicates a connection problem (i.e. instance turns 'yellow'). In the case of your adapter, you just set it to true at the [end of onReady](https://github.com/BenAhrdt/ioBroker.bidirectional-counter/blob/main/main.js#L92). What can happen in your onReady function that is not an exception (i.e. would end your adapter) but that it would not reach the point where you set connection to true? In the case of your adapter the red/green logic is sufficient, if everything is fine, adapter will be green, if there is an error, adapter throws exception and ends and turns red. -> so please remove info.connection. There is an object (system.adapter.biderectional-counter.0.alive) that will indicate if the adapter is running or not.

11/12: No, in fact setState will handle that for you, so you can supply either bidirection-counter.0.xyz or xyz as id, and it will both set the same state. So, please use setState for your own objects.

BenAhrdt:
where i have to delete info.connection?
if i delete this section in io-package.json:
`{
			"_id": "info.connection",
			"type": "state",
			"common": {
				"role": "indicator.connected",
				"name": "Device or service connected",
				"type": "boolean",
				"read": true,
				"write": false,
				"def": false
			},
			"native": {}
		}`
the app crashes at startup

BenAhrdt:
Everything is changed.
Can you check it please?

BenAhrdt:
Whats the meaning of it?
why are there some marks at alexa?
alexa is not my adapter.

Apollon77:
I will check it again  -and the other adapters are checked too because your PR touched them (most likely resorting) ... 

Can youplesase check the conflice in your PR and respove? Thank you

Garfonso:
> why are there some marks at alexa?

It seems the tool you used to edit the json file thought alexa2 was not correctly sorted in. Probably becase of the `-` signs. Did you use the portal to make the PR for the repositories repository?

Apollon77:
@Garfonso This is m,eehhh topic ... it seems that the "default way" using the npm command (which is also iused by dev-portal and such) sorts depending on the set "locale" ... also so a "-" is sometimes sorted before or after the letters :-(

BenAhrdt:
i use https://iobroker.dev/ site to do the PR.

BenAhrdt:
@Apollon77 I dont know what i have to do to resolve the conflicts?
If i click on resolve then i was linked to some alexa code range?
This is not in my adapter.... ????

Apollon77:
I will take care once we merge

Apollon77:
Ok, next round:
* Please change adapter type in io-package and this PR to "misc_data"
* without the connection state please also remove the info channel ... 
If you only support jsonconfig yopu bneed to add common.globalDependencies to io-package with admin >= 5.1.0
* why you did this? https://github.com/BenAhrdt/ioBroker.bidirectional-counter/blob/main/main.js#L189 This will pot not reflect the real loglevel because can be change dynamically. Makes no sense. Do you log that miuch that it makes issues?

BenAhrdt:
@Apollon77 
in this adapter i also use a state in info channel.
is this wrong? what can i do? Move the state creation into io-package?

debug level: i do this if because in some cases i log many entries.) is it ok so?
i checkd it and it works finde with loglevel

BenAhrdt:
Everything should be changed now.
I can't do anything about the conflict with alexa2.

Apollon77:
welcome in the repo

Issue #1782, opened 2022-04-27
Check for archived repositories and remove them from the adapter list


klein0r:
e.g. https://github.com/Kartax/ioBroker.binance

https://github.com/ioBroker/ioBroker.repositories/blob/1fb816c41a67098b2ed3256bca7c1ee1706acfe6/sources-dist-stable.json#L146-L151

klein0r:
Please don't close this

mcm1957:
PR #2321
new script lib/checkArchived.js
The script scans stable and latest repository and creates issues at ioBroker.repositories if:
-) an adapter is listed at stable but not at latest
-) an adapter repository is archived
-) an adapter repository is not accessible

The script does NOT manipulate repository files (*.json) in any way. Removed of adapters must be done manually. As this is a ruther seldom task, automatisation is too risky, espscially as an archived repository could also indiocate that an adpater should not be removed but moved to community area.

The script will be running once a week

PR #1799, opened 2022-05-05
Removed binance

klein0r:
- see #1782


klein0r:
RE-CHECK!

GermanBluefox:
## Automated adapter checker
No changed adapters found


Apollon77:
@klein0r pvforecast had an own PR already ... can you please rebase yours? Then only binance is reloved or? Should we add a admin notification to let users know that it is EOL?

PR #1803, opened 2022-05-09
Add tronity to latest

sebastian-tronity:
Please add my adapter ioBroker.tronity to latest.

Like wished here:
https://gitlab.com/tronity-public/iobroker.tronity/-/issues/3

### DIFF
--- a/sources-dist-stable.json
+++ b/sources-dist-stable.json
@@ -2036,7 +2036,7 @@
     "meta": "https://raw.githubusercontent.com/tronity/ioBroker.tronity/main/io-package.json",
     "icon": "https://raw.githubusercontent.com/tronity/ioBroker.tronity/main/admin/tronity.png",
     "type": "vehicle",
-    "version": "1.1.0"
+    "version": "1.1.1"
   },
   "tunnelbroker-endpoint-updater": {
     "meta": "https://raw.githubusercontent.com/PowerPan/ioBroker.tunnelbroker-endpoint-updater/master/io-package.json",
--- a/sources-dist-stable.json
+++ b/sources-dist-stable.json
@@ -2032,6 +2032,12 @@
     "type": "multimedia",
     "version": "0.0.5"
   },
+  "tronity": {
+    "meta": "https://raw.githubusercontent.com/tronity/ioBroker.tronity/main/io-package.json",
+    "icon": "https://raw.githubusercontent.com/tronity/ioBroker.tronity/main/admin/tronity.png",
+    "type": "vehicle",
+    "version": "1.1.0"
+  },
   "tunnelbroker-endpoint-updater": {
     "meta": "https://raw.githubusercontent.com/PowerPan/ioBroker.tunnelbroker-endpoint-updater/master/io-package.json",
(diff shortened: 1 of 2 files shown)


Apollon77:
RE-CHECK!

GermanBluefox:
## Automated adapter checker

### [ioBroker.tronity](https://github.com/tronity/ioBroker.tronity)
[Downloads](https://www.npmjs.com/package/iobroker.tronity) Number of Installations (latest) 
[NPM](https://nodei.co/npm/iobroker.tronity/)

:thumbsup: No errors found
- [ ] :eyes: [W400] Cannot find "tronity" in latest repository


sebastian-tronity:
@Apollon77 something todo on our side?

Apollon77:
Nope, just "me" (most likely) will do a review

Apollon77:
@sebastian-tronity Hi, here is some review feedback:

* https://github.com/tronity/ioBroker.tronity/runs/6355909499?check_suite_focus=true#step:2:493 ... you need to create the objects toset the state. You can easiely do this for these static objects in io-package like https://github.com/Apollon77/ioBroker.homekit-controller/blob/main/io-package.json#L165-L187
* I think it makes sense to encrypt the client secret, or? https://github.com/tronity/ioBroker.tronity/blob/main/io-package.json#L104. To do that you just need to:
    * require the minimum admin version for this https://github.com/TA2k/ioBroker.psa/blob/master/io-package.json#L81-L85
    * add the field to encrypt to config https://github.com/TA2k/ioBroker.psa/blob/master/io-package.json#L87-L92
    * Rest will happen automatically ... users need to reenter the password once after this change. If you want to know options to communicate that simply tell me and I will provide support
* maybe add some infos to Readme on how to obtain clientid/secret?
* Please consult https://github.com/ioBroker/ioBroker/blob/master/doc/STATE_ROLES.md to add better roles for the created objects at https://github.com/tronity/ioBroker.tronity/blob/8e26f4110c0ad303d5b0aa9d098427c2551223c0/src/main.ts#L26-L38 ; especially "data" is no allowed rile
* With https://github.com/tronity/ioBroker.tronity/blob/8e26f4110c0ad303d5b0aa9d098427c2551223c0/src/main.ts#L111 ... wpuld charging be ever set to "false"? ;-) maybe better check for !== undefined or such? Same for plugged any maybe others. Also value 0 or '' would be filtered that way
* Please respect state.ack in onStateChange and only allow to control state when state.ack===false!
* Maybe it makes sense to trigger a direct data update e.g. 1-2s after the user controlled "commaon.Charging" to get a proper direct data update?

Thank you for checking and adjusting the point. Contact me on any questions,

Ingo

sebastian-tronity:
@Apollon77 thanks for the feedback. Hope now it's better

Apollon77:
Hi @sebastian-tronity,

I had another look and here some suggestions:
* Do really all objects are writeble? I would assume  not :-) https://github.com/tronity/ioBroker.tronity/blob/main/src/main.ts#L53-L54 (code wise only Charging state is writeable) ... no blocker but maybe something your want to adjust before too many users have it installed because setObjectNotExists would not update the objects! (you could switch to extendObject instead that also works but would adjust).
Welcome in the repository

PR #1809, opened 2022-05-13
Add discord to latest

crycode-de:
Repo: https://github.com/crycode-de/ioBroker.discord

Adapter Request: https://github.com/ioBroker/AdapterRequests/issues/322

Channel on ioBroker Discord server: [`#discord`](https://discord.gg/ZdkzZvKwmD)

Forum testing thread: https://forum.iobroker.net/topic/54928/test-adapter-discord-v1-0-x

_Note:_ Node.js >= 16.9 is required for this adapter, since it a requirement of the used discord.js library.

### DIFF
--- a/sources-dist.json
+++ b/sources-dist.json
@@ -324,6 +324,11 @@
     "icon": "https://raw.githubusercontent.com/ioBroker/ioBroker.digitalstrom/master/admin/digitalstrom.png",
     "type": "iot-systems"
   },
+  "discord": {
+    "meta": "https://raw.githubusercontent.com/crycode-de/ioBroker.discord/main/io-package.json",
+    "icon": "https://raw.githubusercontent.com/crycode-de/ioBroker.discord/main/admin/discord.png",
+    "type": "messaging"
+  },
   "discovergy": {
     "meta": "https://raw.githubusercontent.com/DrozmotiX/ioBroker.discovergy/main/io-package.json",
     "icon": "https://raw.githubusercontent.com/DrozmotiX/ioBroker.discovergy/main/admin/discovergy.png",


Apollon77:
RE-CHECK!

GermanBluefox:
## Automated adapter checker

### [ioBroker.discord](https://github.com/crycode-de/ioBroker.discord)
[Downloads](https://www.npmjs.com/package/iobroker.discord) Number of Installations (latest) 
[NPM](https://nodei.co/npm/iobroker.discord/)

- [ ] :heavy_exclamation_mark: [E506] "admin/jsonCustom.json" not found, but custom support is declared
- [ ] :eyes: [W400] Cannot find "discord" in latest repository


crycode-de:
`admin/jsonCustom.json` exists, but is not detected by the adapter checker 🤔

Apollon77:
Known, ignore

Garfonso:
@Apollon77 hope you are not faster, agein. :-p

Garfonso:
Hi,

thanks for the adapter. A lot of features. :-) Generally I think the adapter is of high quality.

Here are some things I found. None of this are dealbreakers, just things to consider (and maybe I even misunderstood some things in some parts): 
* [ ] Is it correct, that this select does not start with _ in front of it: https://github.com/crycode-de/ioBroker.discord/blob/main/admin/jsonConfig.json#L609 ? (just because it is the only thing on that tab without a _ and not in native).
* [ ] Maybe await setting info.connected here https://github.com/crycode-de/ioBroker.discord/blob/main/src/main.ts#L135 - could in rare cases last longer than for example exit because of wrong token or setting it to true later.
* [ ] I'd move the creation of the parent channel out of the Promise.all: https://github.com/crycode-de/ioBroker.discord/blob/main/src/main.ts#L402 -> this currently is checked nowhere (AFAIK), but in future it might be required that all parent objects exists and this is not guaranteed with server.{guild.id} also in the Promise.all call.
* [ ] if `write: true` is correct here https://github.com/crycode-de/ioBroker.discord/blob/main/src/main.ts#L402 then the role `indicator` does not fit (indicator is read-only). I'd advise using role `switch`.
* [ ] Same as above here https://github.com/crycode-de/ioBroker.discord/blob/main/src/main.ts#L561
* [ ] All the sendXY-States have `read: true` set. From my point of view this should be `false` instead, because the content is not updated by the adapter but only used to control (i.e. send) something, right? (similar to button with read: false). I'd change that.
* [ ] Can this happen before all the user objects are created: https://github.com/crycode-de/ioBroker.discord/blob/main/src/main.ts#L264 (i.e. updating states that are not yet created)? Or this https://github.com/crycode-de/ioBroker.discord/blob/main/src/main.ts#L268 ? 
* [ ] Same as  above with this one https://github.com/crycode-de/ioBroker.discord/blob/main/src/main.ts#L246 (probably more secure, though).
* [ ] Since the result will be displayed in UI (right?) can't you add a bit more information for the error reason, here: https://github.com/crycode-de/ioBroker.discord/blob/main/src/main.ts#L1891 ? - Or will it be obvious what is wrong?
* [ ] What about number as string here? https://github.com/crycode-de/ioBroker.discord/blob/main/src/main.ts#L2127
* [ ] In https://github.com/crycode-de/ioBroker.discord/blob/main/src/commands.ts quite some async calls are not waited for. I did not check if this can cause trouble, or not.

Best regards.
Achim

Apollon77:
Just one addition from me:
* Formally common.states are only defined to map numbers to display/UI strings ... So using it with strings in your objects is "officially out of definition" ...

Apollon77:
For common.states if you read also what you posted it is currently defined for number. The "value" in example was meaned as a placeholder. 

But leave as is and read protocol of the dev meeting on Wednesday (or join) for a change on that topic ;-))

Apollon77:
Thank you

PR #1824, opened 2022-05-30
Add wallpanel to latest

xXBJXx:
Please add my adapter ioBroker.wallpanel to latest.

### DIFF
--- a/sources-dist.json
+++ b/sources-dist.json
@@ -1146,10 +1146,10 @@
     "type": "iot-systems"
   },
   "message-queue": {
-     "meta": "https://raw.githubusercontent.com/MK-2001/ioBroker.message-queue/main/io-package.json",
-     "icon": "https://raw.githubusercontent.com/MK-2001/ioBroker.message-queue/main/admin/message-queue.png",
-     "type": "communication"
-   },
+    "meta": "https://raw.githubusercontent.com/MK-2001/ioBroker.message-queue/main/io-package.json",
+    "icon": "https://raw.githubusercontent.com/MK-2001/ioBroker.message-queue/main/admin/message-queue.png",
+    "type": "communication"
+  },
   "meteoalarm": {
     "meta": "https://raw.githubusercontent.com/jack-blackson/ioBroker.meteoalarm/master/io-package.json",
     "icon": "https://raw.githubusercontent.com/jack-blackson/ioBroker.meteoalarm/master/admin/meteoalarm.png",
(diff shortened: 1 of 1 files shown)


Apollon77:
RE-CHECK!

Apollon77:
Hi, than you for the adapter submission,

* Your name replace function goes a bit short ,... https://github.com/xXBJXx/ioBroker.wallpanel/blob/7fb304f444eb03722747b948db212fe8a219735a/build/lib/replaceFunktion.js#L23 maybe check the official one :-) https://github.com/ioBroker/ioBroker.js-controller/blob/master/packages/common/src/lib/common/tools.ts#L54 you get this in adapter.FORBIDDEN_CHARS
* Getting the loglevel from that state https://github.com/xXBJXx/ioBroker.wallpanel/blob/main/src/main.ts#L116 must not be accurate, especially not when started on CLI with debug loglevel. But there is no easy way right now to get the real loglevel :-(

Rest looks fine. Please adjust the replace logic and we can merge

xXBJXx:
Hello Apollon77,
> Your name replace function goes a bit short ,...

I don't quite understand this, what should be too short I **replace** all **ä ö ü ß** with **ae oe ue ss** and with **/[^a-z0-9]/g** all special characters with **_**.
Only letters and numbers allowed, everything else is replaced by a _

CleanShot 18-06-2022 at 00 15 02

> maybe check the official one 
```js
// We allow:
// · Ll = lowercase letters
// · Lu = uppercase letters
// · Nd = numbers
// · ".", "_", "-" (common in IDs)
// · "/" (required for designs)
// · " :!#$%&()+=@^{}|~" (for legacy reasons)
//
/** All characters that may not appear in an object ID. */
export const FORBIDDEN_CHARS = /[^._\-/ :!#$%&()+=@^{}|~\p{Ll}\p{Lu}\p{Nd}]+/gu;
```
I don't understand that, above in **We allow** are all things that are allowed and below in the regEx FORBIDDEN_CHARS then again FORBIDDEN_CHARS should be things that are not allowed or am I wrong there

Apollon77:
The FORBIDDEN_CHARS lists basically anything which is allowed but then it is all "negated" with the first ^ in the character group, so matches all characters except te listed allowed ones. When you are sure that all characters allowed by the wallpanel devices are the normal ones then your regex might be ok ... else the official one is "the better"

xXBJXx:
just tried again to replace the special characters with the FORBIDDEN_CHARS and what can I say, it doesn't really work with it was only *<>?,;[] replaced all other characters stayed there with /[^a-z0-9]/ was all replaced

CleanShot 18-06-2022 at 01 24 29

Apollon77:
Formally the other characters are allowed, but maybe I did not saw something ... so which replace you like to keep?

xXBJXx:
I would leave the method that is now in there so far I have had no problems.

Apollon77:
Then welcome in the repository

PR #1830, opened 2022-06-03
Add janitza-gridvis to latest

BenAhrdt:
Please add my adapter ioBroker.janitza-gridvis to latest.

### DIFF
--- a/sources-dist.json
+++ b/sources-dist.json
@@ -905,6 +905,11 @@
     "icon": "https://raw.githubusercontent.com/iwg-vpn/ioBroker.iwg-vpn/main/admin/iwg-vpn.png",
     "type": "network"
   },
+  "janitza-gridvis": {
+    "meta": "https://raw.githubusercontent.com/BenAhrdt/ioBroker.janitza-gridvis/main/io-package.json",
+    "icon": "https://raw.githubusercontent.com/BenAhrdt/ioBroker.janitza-gridvis/main/admin/janitza-gridvis.png",
+    "type": "energy"
+  },
   "jarvis": {
     "meta": "https://raw.githubusercontent.com/Zefau/ioBroker.jarvis/master/io-package.json",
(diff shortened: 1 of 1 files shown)


Apollon77:
RE-CHECK!

BenAhrdt:
@Garfonso Thanks for the hints.
I will check and change them in a fiew days / weeks :-D

here my comments on the first look:

Change to setObjectNotExistsAsync: OK, top. thanks

[https://github.com/BenAhrdt/ioBroker.janitza-gridvis/blob/main/admin/jsonConfig.json#L178](https://github.com/BenAhrdt/ioBroker.janitza-gridvis/blob/main/admin/jsonConfig.json#L178)

BenAhrdt:
I changed the major things

Garfonso:
> => I create this value with the cronjob creater in the config. => ????

Ok, ignore my comment. Seems like I misread something. 

> So now i have to use the words.js?
As discussed in telegram, of course not. But strings that show up in UI should be either translated (if possible) or English.

BenAhrdt:
i have implemented the translation.
so the Adapter will be included?

Apollon77:
Re-Review will be done soon

Apollon77:
hi,

thank you @Garfonso for the great check!

Here my my review feedback whats left:
* when you only have jsonconfig add a globalDependencies for admin >=5.0.0
* do you really log partly in a localized way? You can do that ,.. I wish you much fun when you get teh first chinese debug log to research an issue *gg* thats the reason why logs are usually always english
* Is that number in the State ID really unique? What happens in the grid config or such when you delete something ... might get the ID reused? Else you end in total data mix ...
*

BenAhrdt:
- global dependencies will be added.
- i will change the logs into english => your write.
- the id is unique. if the user deletes a device and creats a new one a used id will not be used again.

BenAhrdt:
Everything should be changed now.
I do some changes i reaction of a beta tester

BenAhrdt:
Is something to do?

PR #1843, opened 2022-06-17
Added badges with last refresh timestamps

klein0r:


klein0r:
Example: last refresh beta last refresh stable

Apollon77:
Cool idea, thank you!

PR #1848, opened 2022-06-26
Adapter matrix-org added

oelison:


Apollon77:
RE-CHECK!

oelison:
Release v0.0.4 https://github.com/oelison/ioBroker.matrix-org/releases/tag/v0.0.4 could now be reviewed. I think it is in a productive state.

Garfonso:
RE-CHECK!

oelison:
Thank you for the very good hints. All are included in the new version. Except of the "encodeURI". It is a bit difficult, especially the "#" was not replaced and resulted in a wrong request. The last I have done, but not understood. Why ack === false, I would expect true. But it works.

oelison:
Okay, understood now. Thank you for explaining. With the encodeURI is the problem, that it doesn't replace # with %23. When I do it before, the function replace the %23 with %2523. So I replace the # now after the encodeURI and it is working. Rudi1982 has tested it. The actual v0.0.6 is working fine.

Apollon77:
RE-CHECK!

Apollon77:
Hi,

here also some review feedback from me, Thank you @Garfonso for the frst round

* Is the reat UI included really needed seems dummy stuff ad you ave a jsonconfig .. (https://github.com/oelison/ioBroker.matrix-org/blob/main/admin/src/components/settings.jsx#L148-L149) maybe remove that part
* You create several objects in io-ackage ... but then also in adapter code ;-) https://github.com/oelison/ioBroker.matrix-org/blob/main/main.js#L40-L50 ... up tp you in the end, but interesting mix

Rest is fine ... welcome in the repo

oelison:
Hi Apollon77, Thank you for your review. I followed the video from klein0r (haus_automation) for this (my first adapter). The json aproach seems to be the new and better way, but the template generator don't support it. 
To point 1: For me it looks like, that this file is completely not needed admin/src/components/settings.jsx. Or at least the class settings. I don't see where it is called.
To point 2: where I shoud create objects? Allwais in io-package? This was the example for the send message to the adapter. I haven't seen, that I create an object there.
Thank you for merging.

Apollon77:
Hi,

* for react admin stuff: Yes nowhere called, but still there :-) Maybe remove it!? basicall yall that or? https://github.com/oelison/ioBroker.matrix-org/tree/main/admin/src
* other onbjects you already cteate in https://github.com/oelison/ioBroker.matrix-org/blob/main/io-package.json#L157 ... thats easier for "static object"

oelison:
Great. So this could be closed then. I will delete the whole folder src in admin and create the object in io-package.

PR #1873, opened 2022-07-24
Add oekofen-json to latest

chaozmc:
Please add my adapter ioBroker.oekofen-json to latest.

### DIFF
--- a/sources-dist.json
+++ b/sources-dist.json
@@ -1405,7 +1405,7 @@
   "oekofen-json": {
     "meta": "https://raw.githubusercontent.com/chaozmc/ioBroker.oekofen-json/main/io-package.json",
     "icon": "https://raw.githubusercontent.com/chaozmc/ioBroker.oekofen-json/main/admin/oekofen-json.png",
-    "type": "communication"
+    "type": "climate-control"
   },
   "oilfox": {
     "meta": "https://raw.githubusercontent.com/iobroker-community-adapters/ioBroker.oilfox/master/io-package.json",
--- a/sources-dist.json
+++ b/sources-dist.json
@@ -1402,6 +1402,11 @@
     "icon": "https://raw.githubusercontent.com/crycode-de/ioBroker.odl/master/admin/odl.png",
     "type": "misc-data"
   },
+  "oekofen-json": {
+    "meta": "https://raw.githubusercontent.com/chaozmc/ioBroker.oekofen-json/main/io-package.json",
+    "icon": "https://raw.githubusercontent.com/chaozmc/ioBroker.oekofen-json/main/admin/oekofen-json.png",
+    "type": "communication"
+  },
   "oilfox": {
     "meta": "https://raw.githubusercontent.com/iobroker-community-adapters/ioBroker.oilfox/master/io-package.json",
     "icon": "https://raw.githubusercontent.com/iobroker-community-adapters/ioBroker.oilfox/master/admin/oilfox.png",


Apollon77:
Hi, thank you for your adapter. Here are some review comments:

* Should the password be stored encrypted? if yes simply add a key "encryptedNative" and "protectedNative" as array with the field name ("password") and ad min will store it encrypted and decryotion will also be automatic ... it will just be a breaking change for the current users (but I think better now then later) ... like here https://github.com/TA2k/ioBroker.psa/blob/master/io-package.json#L87-L92 (just the default should be then removed)
* You can define static objects ti be created directly in io-pakage like https://github.com/TA2k/ioBroker.psa/blob/master/io-package.json#L100-L122 instead "in code" (and a´you also missed to create the info channel object :-)
* an "await method().then" (e.g. https://github.com/chaozmc/ioBroker.oekofen-json/blob/main/main.js#L111-L112) is an anti pattern .. if you use "then" then you should not also use await. or use await but then no "then"
* The current code would start the update interval also if the initial scan was not successfull. is this wanted? I would expect not because objects are missing then
* The creation of the main objects is all done in parallel. This could be toomuch for smaller systems. Please use Async/await here!
* As optimization you can simply remember the created objects to remember the factor. Migth be betetr then redine the objects on every update ...
* When I read the one issue which is open and read about 1,6k objects then please add a safeguard and remember if adapter was unloaded (unload was called) and check that when setting values (because after unload the DB is gone 500ms later ... with such number of objects setting all the values can eaiely take longer!)
* https://github.com/chaozmc/ioBroker.oekofen-json/blob/main/main.js#L349 ... you do not clear the interval - and si as soon as some triggers this the interval is duplicated, ending in two intervals running and requesting ... please add handling here, else triggering this often enough will kill the device and the iobroker system ;-)

Thank you for checking and adjusting the topics,

Ingo

chaozmc:
Hi Ingo, 

thank you for checking and feedback. I'll have a look into the things you've pointed out and get back soon

chaozmc:
Hi Ingo, 

would you check the latest code again? I've also added a self-termination of the adapter if an error occurs on retrieval, this should also prevent the extensive creation of wrong objects. That (still open) issue I can't reproduce or think of how this could happen and without any further feedback of the User, which I asked for but not received, I also couldn't troubleshoot that behavior any further. 

I guess I've covered everything beside a local copy of the objects. 
In regards of this, I'm thinking about how this could be changed at all, and maybe rewrite the update procedure in a next version and compare for example the last values with the current retrieved ones and only write those objects with changed values to ioBroker. But I think this will require more testing as I'm a bit concerned about the memory demand as soon as I create a local copy of all objects with the current and last retrieved values to compare them afterwards.

If you don't think that would be a crucial requirement for now and the code rewrite can be postponed as an improvement for a future version, I'd say this release is finished. Except I've missed something again but at least for me the adapter runs without any issues.


Thank you, 
Max

GermanBluefox:
RE-CHECK!

GermanBluefox:
## Automated adapter checker

### [ioBroker.oekofen-json](https://github.com/chaozmc/ioBroker.oekofen-json)
[Downloads](https://www.npmjs.com/package/iobroker.oekofen-json) Number of Installations (latest) 
[NPM](https://nodei.co/npm/iobroker.oekofen-json/)

:thumbsup: No errors found
- [ ] :eyes: [W400] Cannot find "oekofen-json" in latest repository


Apollon77:
Hi, 

sorry that my re-review took that long.

In fact many things are adjusted. Thank you for that. 

But the main initialization look still does  "setObjectNotExists" calls and they all run in parallel. If we really talk about many objects this can crash or overload smaller systems. Please replace the forEach there with for Loops and simply use "await setObjectNotExistsAsync()" because then they are all created sequencially ... 

Thank you for adjusting this.

Ingo

Apollon77:
@chaozmc Anything we can support with?

chaozmc:
Hi @Apollon77 - sorry my Laptop was broken and my whole VS setup on it etc. was gone with it too... 
In regards of the for-loop I'll have a look how I could do that. 

Is a foreach loop handled in parallel and not sequential? Would it be sufficient to replace [this](https://github.com/chaozmc/ioBroker.oekofen-json/blob/d904b23817e64ced2883ea36f552a31b092a3fe0/main.js#L262) with await this.setObjectNotExistsAsync()?

chaozmc:
Hi @Apollon77 again, I've searched around in the internet about the object.Keys().forEach() topic and the parallelism and understood what you've mentioned. I've changed the functions to use a for...of loop which should target what you've meant.

Apollon77:
I meaned exactly that :-)

But now after having another look ... what is https://github.com/chaozmc/ioBroker.oekofen-json/blob/main/main.js#L4 seems unused and also such a package is not declared in your package.json ... Please check this.

Rest fine now

chaozmc:
removed, idk where this came from, maybe VSCode has putted that in automatically due to IntelliSense

Apollon77:
Welcome in the repository, sorry that it took that long

Apollon77:
BTW ... stttoooop. One last thing ... Adapter type "communication" is not really matching ... Please check for a better matching type ... household? hardware? Please adjust in youtr io-package and in this PR. Thank you!

chaozmc:
I've looked around other adapters with familiar capabilities, "climate-control" seems to be the best category as there are other heating-system adapters already. I've adjusted the category accordingly in both locations

Apollon77:
Welcome in the repo then!

PR #1878, opened 2022-07-31
Add device-watcher to latest

ciddi89:
Please add my adapter ioBroker.device-watcher to latest.

### DIFF
--- a/sources-dist.json
+++ b/sources-dist.json
@@ -334,6 +334,11 @@
     "icon": "https://raw.githubusercontent.com/xenon-s/ioBroker.device-reminder/master/admin/icon.png",
     "type": "energy"
   },
+  "device-watcher": {
+    "meta": "https://raw.githubusercontent.com/ciddi89/ioBroker.device-watcher/main/io-package.json",
+    "icon": "https://raw.githubusercontent.com/ciddi89/ioBroker.device-watcher/main/admin/device-watcher.png",
+    "type": "misc-data"
+  },
   "devices": {
     "meta": "https://raw.githubusercontent.com/ioBroker/ioBroker.devices/master/io-package.json",
     "icon": "https://raw.githubusercontent.com/ioBroker/ioBroker.devices/master/admin/devices.png",


o0shojo0o:
I have been using the adapter for a while and have not been able to find any errors in the code either.
Where I'm not sure again is the io-package.json, as I'm not up to date.

From my point of view it is ready for the beta repo :)

GermanBluefox:
RE-CHECK!

GermanBluefox:
## Automated adapter checker

### [ioBroker.device-watcher](https://github.com/ciddi89/ioBroker.device-watcher)
[Downloads](https://www.npmjs.com/package/iobroker.device-watcher) Number of Installations (latest) 
[NPM](https://nodei.co/npm/iobroker.device-watcher/)

:thumbsup: No errors found
- [ ] :eyes: [W400] Cannot find "device-watcher" in latest repository


Apollon77:
Hi,

thank your your adapter submission. Here is some review comments:
* You use the release script in your dependencies. You also need to add the iobroker plugin and a release config file. see the release script changelog from version 3! Else it will not work
* Please check your roles. Indicator is only a read-only boolean! not strings.
* on start you execute main() and then refresh() method ... and refresh does main() again directly?= Really needed twice at tghe beginning? ;-)
* Also in general the adapter works very unperformant and create big load spikes on the system of the users. you should consider changing the following things:
  * The main logic is in "main" and here you do the object creation any time it is called. YOu should remember that objects are created or just do that on start!
  * The main logic is always reading aaalll the states needed ... I saw logic to handle subscriptions but commented out? Why you not simply read them once, put in a map/object by their ID and subscribe for changes and just update that object and have the counting logic working on that object? This would be way more performant and can update states "on change" and also not needed as interval. But without state changes (then please remove the stateChange method) yu can do it that way
* becauseof the fact that your current way to get all data can take long please set a variable in "unload" that unload was called and add a check in your data processing loops to cancel further processing of the current run if unload was called. Especially do not set a new timeout then! (because if unload is called while one processing is running you will set a new timeout at the end :-)

Thank you for checking my comments and adjustments.

Simply poke here when done and I do a re-check.

Ingo

ciddi89:
Hi Ingo,

thanks for your review. I worked out all of your points. And I hope it's okay for you now. I made already some changes to make the adapter performant and in the last ten days the ram usage was only between 80 and 90 percent on my system. 

Because of the point with subscription on state changes. For now I removed it now. But in near future I will use it as suggested by you. But I hadn't time enough in the last weeks to implemented it. Hope it's okay for you. 

If there is anything I forgot please give me an advice again. 

Thanks 

Christian

Apollon77:
Hey, I did a recheck and beside the already discussed downfalls of the general approach  it is ok now. I still do not "like" that all states are crerated in any interval-run, but also here ... acceptable ;-) Future optimizations would be cool.

But the release script will not work as it is now ... nothing changed there, the iobroker plugin is still missing ... but thats up to you how you release.

ciddi89:
I'm glad to hear that. Yes I know, I will change it as soon as possible with the Intervall. 

But can you explain where I have to add the iobroker plugin for the release script? I choosed the option in the adapter creation so I thought this is all clear and will work correctly. (And it did first time except the last time). I looked already in the changelog and to other Adapter which uses the release script but doesn’t find it where I have to add this plugin.

Apollon77:
Yes the creator aded the release script package in v2.x ... you upgraded it to 3.x but this is a MAJOR change ... Major normally means that something is breaking! So you need to check those very careful.

see https://github.com/AlCalzone/release-script#plugins

ciddi89:
okay I think I've added it now correctly. Thanks for your advice.

PR #1887, opened 2022-08-05
added wamo

smarthausleben:
Please add my adapter „wamo” to the latest.

1.	I’m aware about the “Timer” and “Timeout” problematic, but this should be under control in any circumstances.
2.	I’m aware about the dynamic state handling, but this may be necessary in further releases anyway because I plan to make them selectable in adapter options in a further release.

The adapter is not mentioned in the ioBroker forum yet because it was never planed to have this adapter in the official list. All adapter users have reported their problems by e-mail and within Github issues. 

But since the SYR company (builder of the hardware) contacted me and gave very positive feedback and also technical support, I thought it may be good to push this project a little bit further and try to get the adapter into the official list. Even if my time is verry limited and I’m really struggling with all the procedures to get the adapter in the official adapter list, I went this road in order to give something back to the SYR company.

Even if there are still a lot of cosmetical issues to address, the current version is running stable without any problems and device disconnections etc. are fully addressed and under control. The adapter is already running on several ioBroker systems and as you can see in Github issues, I have taken care of the problems.

Adapter Ceck on ioBroker.dev is showing an error ([E606] Current adapter version 0.2.5 not found in README.md) but I can’t really fix it because from my point of view everything is OK in README.md. Maybe you can tell me what’s wrong there.

So, thanks for your feedback and for listing the adapter in latest.

Thanks
Claus


GermanBluefox:
RE-CHECK!

GermanBluefox:
## Automated adapter checker

### [ioBroker.wamo](https://github.com/smarthausleben/ioBroker.wamo)
[Downloads](https://www.npmjs.com/package/iobroker.wamo) Number of Installations (latest) 
[NPM](https://nodei.co/npm/iobroker.wamo/)

:thumbsup: No errors found
- [ ] :eyes: [W400] Cannot find "wamo" in latest repository


smarthausleben:
Anything I have to take care?

Apollon77:
Yes me as soon as some time critical topics are solved

smarthausleben:
> Yes me as soon as some time critical topics are solved

No rush, just thought I’m missing something I should have to do.
Thanks for your reply.

Apollon77:
Hi,

Thank you for your adater submission. Here are my review comments:
* Please remove the package-lock.json changes from this PR here.
* Please also create a channel info for https://github.com/smarthausleben/ioBroker.wamo/blob/main/io-package.json#L177
* Your logic in timerStarts uses sleep to wait 3s several times. But this gets problematic if the adapter gets unloaded in between. Please add a variable to remember that unload was called and make sure to stop and clear all stuff in such a case. Alternatively use "adapter.delay" and "adapter.setInterval" available since js-controller 4.0 because they are unload-safe
* doing setObjectNotExists in every data update loop is not really a perfoermant approach ... consider creating objects once or remember which were already created.
* please do not invent new roles (like https://github.com/smarthausleben/ioBroker.wamo/blob/main/lib/device-parameters.js#L458) ... please use the defined ones only https://github.com/ioBroker/ioBroker/blob/master/doc/STATE_ROLES.md

Thank you for checking and adjusting, just poke here again for a re-check

Ingo

smarthausleben:
Hello Ingo,

thanks for your review and your suggestions:

I fixed all your mentioned points except number 4 
> * doing setObjectNotExists in every data update loop is not really a perfoermant approach ... consider creating objects once or remember which were already created.

To be honest, I really like the way I’m doing this because I’m somehow on the save site if a user deletes an object manually. But if this has such a big impact on the performance, I probably should use this method just once at startup and then I have to deal with errors from “setState” if I try to put a value into a not existing object.

Would this be a more preferred way to deal with “dynamic states”?
Or maybe there is a better way to do this I haven’t thought about yet?

Thanks for your support.

Claus

Apollon77:
In fact it all depends on the number of objects we talk about and how often it is done. If this is done for each refresh then each object is checked for existence once before the state is written (what would do that again).

In my eyes: If an user deletes objects manually without restarting the adapter he is responsible for the effects ;-)ANd in fact a restar of the adapter will then automatically fix all these issues.... so it is self healing with a restart ... 

but up to you: Doing it like this is not really a blocker for the repo acceptance, see it as review comment :-)
Ingo

Apollon77:
Ok, I recheck the rest tomorrow. Yes this one topic is not a blocker

Apollon77:
One last small thing: now with using "this.delay" you also need to increase the js-controller dep in https://github.com/smarthausleben/ioBroker.wamo/blob/main/io-package.json#L177 to >=4.0.0

Please one last "fix" and release .. then we are ok

smarthausleben:
brrrr... dependencies :-) I changed it to >=4.0.0 as you requested

Sorry it took a bit but I’m still fighting with my release-script process, Because since a while I’m getting strange answer on **“executing "git push --tags"**
- \* [new tag]         v0.2.12 -> v0.2.12
-  ! [rejected]        v0.2.3 -> v0.2.3 (already exists)

and I have no clou why.

So I had to do it manually.

Thanks for your **re-check**

Apollon77:
maybe you have a local tag which is ut of sync with github ... delete this local tag ...

smarthausleben:
Thanks for the hint. I deleted local tag v.0.2.3 ... I'm curious if it works next time.

smarthausleben:
Is there anything else I have to fix?
Could you already re-check the adapter?

smarthausleben:
I just released version 0.2.13 where I implemented your suggestion of “**setObjectNotExistsAsync**” improvement. Objects will now be checked/created during adapter startup only.

It would be great if you can **re-check** the adapter 😊

Maybe you can also tell me when I approximately can expect to see the adapter in the official list?

I highly appreciate your work here 👍 and many thanks for your answer.

Apollon77:
Welcome in the repo

smarthausleben:
> Welcome in the repo

Good news 😃​ Thanks a lot 🤗​

PR #1929, opened 2022-09-09
Add ecoflow to latest

Newan:


Apollon77:
RE-CHECK!

GermanBluefox:
## Automated adapter checker

### [ioBroker.ecoflow](https://github.com/Newan/ioBroker.ecoflow)
[Downloads](https://www.npmjs.com/package/iobroker.ecoflow) Number of Installations (latest) 
[NPM](https://nodei.co/npm/iobroker.ecoflow/)

:thumbsup: No errors found
- [ ] :eyes: [W400] Cannot find "ecoflow" in latest repository


Apollon77:
Hi,

thank you for your submission and here are some review comments:
* It seems that some states are incorrectly defined, see your testing results: https://github.com/Newan/ioBroker.ecoflow/actions/runs/3013617524/jobs/4842974462#step:2:520 ... you should set "def" values matching the defined datatype :-)
* makes it sense to encrypt the secretkey? Can be handled automatically ... just defined as nativeEncrypted (see best practices)
* please also create channels (info/status) for your static objects
* i think you have a bnad typo in your deps ... https://github.com/Newan/ioBroker.ecoflow/blob/main/package.json#L20 ... this should be "axios" or? ;-)
* The release script will not work as it is right now because when using v3 syou need another plugin dep and a config file. see the v3 changelog off release script
* you should set a timeout for the external communication which is always lower then your interval (or not use an interval at all - see best practices, but also then a timeout is a good idea)

Thank you for checking and adjusting,

Ingo

Newan:
fixed

Newan:
> Re-Review done. Please release the new version :-) Then I can merge.

New Version released

PR #1941, opened 2022-09-12
add yahoo-stock-market to latest

Newan:


Apollon77:
RE-CHECK!

Apollon77:
Hi,

release comments incoming :-)
* please elete .commitinfo ( seems from an unsuscessfull release)
* add the release script plugins and config for 3.x - seems that you upgraded the release script dep and did not checked major changes
* you can not clear the timeout from https://github.com/Newan/ioBroker.yahoo-stock-market/blob/main/src/main.ts#L74 how you do it ... just make a normal setTimeout and not that promise stuff in this place because it also makes no sense
* There are jsonconfig types that offer a "toast" like entering of multiple values and you get it as array or comma separated ... but easier for user then a normal "text". maybe check that out.

Please check and adjust the comments

Ingo

Newan:
@Apollon77  Thanks for the feedback.

- "add the release script plugins and config for 3.x - seems that you upgraded the release script dep and did not checked major changes" -> ok i use the adapter creator and then the release script. this failed wih the currentversion, so i musst update. so please update this too.
https://github.com/ioBroker/create-adapter/issues/824

- "There are jsonconfig types that offer a "toast" like entering of multiple values and you get it as array or comma separated ... but easier for user then a normal "text". maybe check that out." -> Please give me a hint. I only know this .. but no "toast":  https://github.com/ioBroker/ioBroker.admin/blob/master/src-rx/src/components/JsonConfigComponent/SCHEMA.md

Apollon77:
to 2.) sorry I used wrong word ... meaned "chips" (can be found in jsonconfig schema

Newan:
fixed

Newan:
ok the timeout option in the yahoo lib is on TODO:   [timeout?: number; // TODO](https://github.com/gadicc/node-yahoo-finance2/blob/55298986e92de3f91f783324fd8919745a6c6e8e/src/lib/queue.ts#L11)

new release online

PR #1951, opened 2022-09-26
Update tonew stable version

BenAhrdt:
Updating operating-hours to 0.0.15

=> Automatic reset implemented


Apollon77:
From where do you know that this version is "stable"? Normally we require stable to have at least some installations or time passed by, so that you can be sure to not have a broken version.

In special cases on very important bugfixes we allow faster stable update ... But this seems not the ase here, or?

BenAhrdt:
OK, i have checked it.
also it is possible to set it to latest.

BenAhrdt:
OK, so tomorrow the user can install beta. wright?

BenAhrdt:
i closed this pR because i will check it with testers in Beta and maby do some changes

PR #1955, opened 2022-10-02
Bosch Smart Home Bridge Adapter stable 0.2.0

holomekc:
Bosch Smart Home Bridge Adapter stable 0.2.0


Apollon77:
0.2.0 is from 6h ago ... really to stable that fast? I think it can not be proofed that fast by users or?

holomekc:
Hi no there was not enough time. I thought you also need time to check changes. If this is not the approach I will create the pr later the next time.

Apollon77:
The adapter is only reviewed in very first latest additionl ... after that not (with >500 adapters this would never be manageable!). So normally stable PRs are just merged ... sometimes I also check since when the version is online ... ;-)

holomekc:
I see it is fine. Is it ok if I comment here when I get some feedback or recreate the pr?

holomekc:
Ah and maybe one thing. I was not sure regarding the admin changes. It works fine but the example confused me, because I could really see which files I need.

Apollon77:
We can leave open and you comment in some days ...

What admin changes you mean? Please send links rto the relevant commits

holomekc:
Basically those two:
https://github.com/holomekc/ioBroker.bshb/commit/d954fd424dc06ce6cc6ffa87e24f780c1b2f4753
https://github.com/holomekc/ioBroker.bshb/commit/ddaae4e1a6df4b4b91503f8a2b949c094f5d152b

Not sure how the approach is even named, but basically create the admin ui from a json file.

Apollon77:
In general (rought view over it) looks ok. in fact I ssume you tested that config works. But nowyou only have jsonconfig, correct? Then you need to add to io-packge.sjon an "common.globalDependencies" section with admin >= 5.0.0 as dependency because your adapter can only be configured with admin 5+

holomekc:
Thank you. I updated the file.

PR #1961, opened 2022-10-07
New version of adapter Growatt 2.0.0

PLCHome:


Apollon77:
Hm ... a new major version after <1 day as stable? Is the version stable proofed? Or is it needed because of old version is not working anymore? Normally stable after 1-2 weeks of latest or significant installation numbers without reported issues

PLCHome:
No it's not that important. I was asked about the function. The critical part of the adapter had not changed. Now you can make a few settings in the firmware of the device, which is why I upgraded the version. The functionality has been available for some time and has been tested in the API. So I'm pretty sure it won't cause any problems.

Apollon77:
So you reopen it in some days?

PR #1963, opened 2022-10-08
Bump Adapter nibeuplink to 1.2.1

sebilm:


Apollon77:
How logn this version is out? Iassume reletiavely short, or?

Stable updates we do normally after several days in latest withour reported issues and some installations. normally after 15 days.
We can be faster if there is a need like an important bugfix or such fpr it.

Is it that way here?

sebilm:
I changed very little code and mainly updated dependencies. Also added support for Node 18. And that's what someone asked for 4 days ago. So I thought it would be no problem to declare the version stable after one week. I have also tested with Node 14, 16 and 18 myself and also still with JS Controller 3.3.23.

PR #1973, opened 2022-10-13
Add new version of iobroker.apcups

XHunter74:


Apollon77:
Really a new stable version 1h after release ;-) Normally stable versions stay in latest untill thes are proofed some days by users and got some installations ... or is that an important bugfix?

XHunter74:
Just bug fixing that should not impact functionality.

Apollon77:
Ok I merge now, but in the future please give new versions some days in Beta/Latest before pushing to stable ( and you do not wanna hear all own/other stories how "should not" can be wrong ;-)

XHunter74:
Ok, this moment of publishing flow was not clear to me.

PR #1974, opened 2022-10-13
added alternative pjlink

Bannsaenger:


Apollon77:
RE-CHECK!

Bannsaenger:
die Warning [W145] Each "common.news" should be translated into all supported languages (en, de, ru, pt, nl, fr, it, es, pl, zh-cn)
ist wohl nicht mehr richtig. Das translate script übersetzt auch in ukrainisch. Die common.news hat jetzt 11 Sprachen.

Apollon77:
jupp, Known, so ignore

Apollon77:
Hi,

here are my review comments:
* a default value for a password makes no sens if you use automatic encryption because then your default one will be decrypted - rendering it to glibberish ...
* Whats that? https://github.com/Bannsaenger/ioBroker.pjlink/blob/main/main.js#L242 (and many more places) setTimeout shoukd only return a "Number" ... calling "refresh" on it does what?
* Please make sure to clear all timeouts in unload ... I think at least one is missing.

Bannsaenger:
Ok,

- default password is removed
- i changed all timeouts to intervalls. All are cleared on unload
- the setTimeout return a timer object. Calling refresh on it restarts the time. But it is now obsolete.

i hope this meets the needs

Apollon77:
All good and fine. Welceome in the repo

PS: Thank you for the "refresh" info ... Didn't know that

PR #1975, opened 2022-10-14
Adding Wattpilot Adapter to the source list and stable repo

tim2zg:
Hello, 
The adapter is now ready and tested, I added the adapter to the normal repo and stable repo. 
Here is the forum thread: https://forum.iobroker.net/topic/55343/fronius-wattpilot-test-3-0-x

Best regards, Tim

### DIFF
--- a/sources-dist-stable.json
+++ b/sources-dist-stable.json
@@ -630,7 +630,7 @@
     "meta": "https://raw.githubusercontent.com/tim2zg/ioBroker.fronius-wattpilot/main/io-package.json",
     "icon": "https://raw.githubusercontent.com/tim2zg/ioBroker.fronius-wattpilot/main/admin/fronius-wattpilot.png",
     "type": "energy",
-    "version": "4.1.0"
+    "version": "4.2.0"
   },
   "frontier_silicon": {
     "meta": "https://raw.githubusercontent.com/halloamt/ioBroker.frontier_silicon/master/io-package.json",
--- a/sources-dist-stable.json
+++ b/sources-dist-stable.json
@@ -629,7 +629,7 @@
   "fronius-wattpilot": {
     "meta": "https://raw.githubusercontent.com/tim2zg/ioBroker.fronius-wattpilot/main/io-package.json",
     "icon": "https://raw.githubusercontent.com/tim2zg/ioBroker.fronius-wattpilot/main/admin/fronius-wattpilot.png",
-    "type": "energy",
+    "type": "vehicle",
     "version": "4.3.0"
   },
   "frontier_silicon": {
--- a/sources-dist-stable.json
+++ b/sources-dist-stable.json
@@ -630,7 +630,7 @@
     "meta": "https://raw.githubusercontent.com/tim2zg/ioBroker.fronius-wattpilot/main/io-package.json",
     "icon": "https://raw.githubusercontent.com/tim2zg/ioBroker.fronius-wattpilot/main/admin/fronius-wattpilot.png",
     "type": "energy",
-    "version": "4.0.0"
+    "version": "4.1.0"
   },
   "frontier_silicon": {
     "meta": "https://raw.githubusercontent.com/halloamt/ioBroker.frontier_silicon/master/io-package.json",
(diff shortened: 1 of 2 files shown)


Apollon77:
RE_CHECK!

Apollon77:
RE-CHECK!

GermanBluefox:
## Automated adapter checker

### [ioBroker.fronius-wattpilot](https://github.com/tim2zg/ioBroker.fronius-wattpilot)
[Downloads](https://www.npmjs.com/package/iobroker.fronius-wattpilot) Number of Installations (latest) 
[NPM](https://nodei.co/npm/iobroker.fronius-wattpilot/)

:thumbsup: No errors found
- [ ] :eyes: [W171] "common.title" is deprecated in [io-package.json](https://github.com/tim2zg/ioBroker.fronius-wattpilot/blob/master/io-package.json)
- [ ] :eyes: [W105] "common.titleLang" should be translated into all supported languages (en, de, ru, pt, nl, fr, it, es, pl, uk, zh-cn)
- [ ] :eyes: [W109] "common.desc" should be translated into all supported languages (en, de, ru, pt, nl, fr, it, es, pl, uk, zh-cn)
- [ ] :eyes: [W156] Adapter should support admin 5 UI (jsonConfig) if you do not use a React based UI
- [ ] :eyes: [W145] Each "common.news" should be translated into all supported languages (en, de, ru, pt, nl, fr, it, es, pl, uk, zh-cn)
- [ ] :eyes: [W145] Each "common.news" should be translated into all supported languages (en, de, ru, pt, nl, fr, it, es, pl, uk, zh-cn)
- [ ] :eyes: [W145] Each "common.news" should be translated into all supported languages (en, de, ru, pt, nl, fr, it, es, pl, uk, zh-cn)
- [ ] :eyes: [W145] Each "common.news" should be translated into all supported languages (en, de, ru, pt, nl, fr, it, es, pl, uk, zh-cn)
- [ ] :eyes: [W145] Each "common.news" should be translated into all supported languages (en, de, ru, pt, nl, fr, it, es, pl, uk, zh-cn)
- [ ] :eyes: [W145] Each "common.news" should be translated into all supported languages (en, de, ru, pt, nl, fr, it, es, pl, uk, zh-cn)
- [ ] :eyes: [W400] Cannot find "fronius-wattpilot" in latest repository


Apollon77:
The notes about langiages from check can be ignored

Apollon77:
Hi,

here some review feedback:
* Why the adapter has itself as depenency? https://github.com/tim2zg/ioBroker.fronius-wattpilot/blob/main/package.json#L32
* Becuse you use encryptedNative you also need to set the dependencies correctly. js-controller >=3 and as "common.globalDependencies" set admin>4.0.9
* Please only react to state changes when ack===false.
* having the role "level" for all cases is not making sense. Please respect https://github.com/ioBroker/ioBroker/blob/master/doc/STATE_ROLES.md also about which roles are allowed for which data types
* When I get the code right you always do the setObjectNotExists call/check for every data package received, right? This is a lot of objects DB actions that are not needed if you remember whoich obvers where already created/checked once. Such code can make issues on smaller systems because of the high number of database calls.

Thank you for checking and adjusting the mentioned topics

tim2zg:
Thank you so much for the feedback. I fixed the issues you pointed out in version 3.3.0.

tim2zg:
Updated to new Version 4.1.0

tim2zg:
Updated to new Version 4.2.0

Apollon77:
@tim2zg for that the re-review took a bit. Small stuff left

* The admin dependency in io-package needs to go into "common.globalDependencies" (for context: dependencies means "needs to be on the same host"; globalDependencies means "needs to be in the system somewhere")
* Please do not provide a default value for password ... because it will be encrypted this default is destroyed anyway and might confuse users

For the roles I think there are better  ones here and there but I will accept the pragmatic solution you choose ;-)

After that we are ready to go. Please poke here again when changed and released as new version. Thank you!

tim2zg:
Done thank you!

Apollon77:
Ok, then we go ... interesting how "smae" that device looks like a go-e :-)

But because it is a wallbox mabe type "vehicle" is better?
Please adjust io-package on github and here in PR ... no release needed now, include in next

tim2zg:
Thank you! Done.
Yes, I think it’s the same hardware and software but another cloud and the API options are not unlocked. But don’t nail me on that.

PR #1979, opened 2022-10-21
Set operating-hours 0.0.19 into Stable

BenAhrdt:
New State implemented and works for the tester fine in beta


BenAhrdt:
@Apollon77 is it possible something went wrong?
In stable there is only 0.0.18 available and in beta there is just 0.0.17

Apollon77:
Are you sure you reloaded the repo on your system?

BenAhrdt:
Yes. On my Testsystem ist Works.
Nur ob Productive Not. Bit today in the Morning its ok. Not Self (i have to refresh)
Why there is an other adapter refreshing yesterday on my system and this adapter not? The merge Is not there at the same time on different systems?

Apollon77:
In your admin adapter you have a setting how often repo is updated ... compare this between your systems. the default is "daily" aka 24h ...

PR #1980, opened 2022-10-21
set janitza-gridvis 2.1.23 into stable

BenAhrdt:
warning (logging) for axios response 404 implemented. tested => works fine.
User need this information in case of error directly. Otherwise the reading of data will be blocked.


BenAhrdt:
one more feature will be implemented

PR #1982, opened 2022-10-21
set janitza-gridvis 2.1.25 into stable

BenAhrdt:
just updated some bugfixes / comments for logging --- no functional changes


Apollon77:
honestly ... yes also with "no functional changes" we should please keep the process to have  a new version in latest for some days ... I will merge now last time directly

BenAhrdt:
OK, thanks

PR #2093, opened 2023-01-06
Updated sainlogic to 0.10.0

phifogg:


Apollon77:
PS: @phifogg normally we do stable updates after a new version has been proofed working in latest/Beta fpor 1-2 weeks :-)

PR #2116, opened 2023-01-27
new stable version of iwg-vpn

iwg-vpn:


Apollon77:
A major update? Any breaking change a user should be informed about or such?

iwg-vpn:
No breaking changes, nothing a user has to do. The set of features and stability have reached the grade of the v1. That's the reason.

PR #2157, opened 2023-02-26
add elgato-key-light to Latest

xXBJXx:
Please add my adapter ioBroker.elgato-key-light to latest.

### DIFF
--- a/sources-dist.json
+++ b/sources-dist.json
@@ -464,6 +464,11 @@
     "icon": "https://raw.githubusercontent.com/marc2016/ioBroker.elero-usb-transmitter/master/admin/elero-usb-transmitter.png",
     "type": "iot-systems"
   },
+  "elgato-key-light": {
+    "meta": "https://raw.githubusercontent.com/xXBJXx/ioBroker.elgato-key-light/main/io-package.json",
+    "icon": "https://raw.githubusercontent.com/xXBJXx/ioBroker.elgato-key-light/main/admin/elgato-key-light.png",
+    "type": "lighting"
+  },
   "email": {
     "meta": "https://raw.githubusercontent.com/iobroker-community-adapters/ioBroker.email/master/io-package.json",
     "icon": "https://raw.githubusercontent.com/iobroker-community-adapters/ioBroker.email/master/admin/email.png",


Apollon77:
RE-CHECK!

GermanBluefox:
## Automated adapter checker

### [ioBroker.elgato-key-light](https://github.com/xXBJXx/ioBroker.elgato-key-light)
[Downloads](https://www.npmjs.com/package/iobroker.elgato-key-light) Number of Installations (latest) 
[NPM](https://nodei.co/npm/iobroker.elgato-key-light/)

- [ ] :heavy_exclamation_mark: [E114] No adapter are allowed in the repo without admin support (set "common.noConfig = true" and "common.adminUI.config = none" if adapter has no configuration)
- [ ] :eyes: [W171] "common.title" is deprecated in [io-package.json](https://github.com/xXBJXx/ioBroker.elgato-key-light/blob/master/io-package.json)
- [ ] :eyes: [W156] Adapter should support admin 5 UI (jsonConfig) if you do not use a React based UI
- [ ] :eyes: [W400] Cannot find "elgato-key-light" in latest repository


mcm1957:
RE-CHECK!

GermanBluefox:
## Automated adapter checker

### [ioBroker.elgato-key-light](https://github.com/xXBJXx/ioBroker.elgato-key-light)
[Downloads](https://www.npmjs.com/package/iobroker.elgato-key-light) Number of Installations (latest) 
[NPM](https://nodei.co/npm/iobroker.elgato-key-light/)

- [ ] :heavy_exclamation_mark: [E114] No adapter are allowed in the repo without admin support (set "common.noConfig = true" and "common.adminUI.config = none" if adapter has no configuration)
- [ ] :eyes: [W171] "common.title" is deprecated in [io-package.json](https://github.com/xXBJXx/ioBroker.elgato-key-light/blob/master/io-package.json)
- [ ] :eyes: [W156] Adapter should support admin 5 UI (jsonConfig) if you do not use a React based UI
- [ ] :eyes: [W400] Cannot find "elgato-key-light" in latest repository


mcm1957:
RE-CHECK!

mcm1957:

- [ ] errors reported by Apollon77 must be fixed

  see comment above
  
- [ ] adapter has 'singleton' set to true (aka only one instance per host allowed) - why?

  Why is 'singleton=true' attribute set. Thes limits the usage of the adapter to one single instance per host.

  Some characters are not allowed to be used as part of an object id. If an object id is not hardcoded or even depending on user input you should ensure that such an object id does not contain an invalid character.  Iobroker provides the constant ```adapter.FORBIDDEN_CHARS```.- You are using the serial number reported by the device. Please verify that this serial number does not contain any forbissen characters or spaces and replace such characterw i.e. by '_'.

- [ ] axios timeout

  axios calls should specify a specific timeout. Accoring to axios documentation the default for the option timeout is 0 (no timeout). This could result in an stalled adapter.

- [ ] role button

  states with role button should (mast) specify common.read=false.
  see https://github.com/ioBroker/ioBroker.docs/blob/master/docs/en/dev/stateroles.md
  https://github.com/xXBJXx/ioBroker.elgato-key-light/blob/43315a3a6059ddea5c83e4583205195bcc4a6816/src/main.ts#L363C1-L363C1

- [ ] reevaluate state roles

  Tests for node 16 is mandatory unless you require node 18 and higher.
  Tests for node 18 is mandatory as many users already use node 18 and node 16 will be eol and deprected Q3/2023.
  
McM1957

mcm1957:
Development of adapter suspended for a currently unknown timeframe.
PR cancelled

Please open a ne PR if development to be continued

PR #2235, opened 2023-04-18
Migrate ioBroker.beckhoff Adapter to iobroker-community-adapters

dkleber89:


mcm1957:
@dkleber89 
Do you plan to continue maintain this adapter in the future? Or is the migration to community-adapters a sort of controlled shutdown?

I'm asking as I would like to know whether I should/can add you as (primary) maintainer at iobroker.comminuty-adapters for this adapter or if this is inappropiate. Thanks for ashort reply.

dkleber89:
You can do it as u like ... The reason for this migration on my side is, that i don´t have too much time to maintain this adapter and especially make new features ... So in the community everyone (also myself) have easily the ability to work on it but im not alone responsible for this adapter.

PR #2275, opened 2023-05-09
Update oekofen-json to 1.0.3

chaozmc:
Please update my adapter ioBroker.oekofen-json to version 1.0.3.

### DIFF
--- a/sources-dist-stable.json
+++ b/sources-dist-stable.json
@@ -1621,6 +1621,12 @@
     "type": "misc-data",
     "version": "2.0.5"
   },
+  "oekofen-json": {
+    "meta": "https://raw.githubusercontent.com/chaozmc/ioBroker.oekofen-json/main/io-package.json",
+    "icon": "https://raw.githubusercontent.com/chaozmc/ioBroker.oekofen-json/main/admin/oekofen-json.png",
+    "type": "climate-control",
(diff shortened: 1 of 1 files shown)


mcm1957:
Normally a release should be added to stable only after it has been available at latest some time and has some installatiosn (and of course no known issues). You just created 1.0.3. I think you just updated trabnsaltion and some tools (release-script, test-amnd-release) but di not change anything at code. If this is correct I guess @Apollon77 will accept anyway.

In addition I added a issue to add node versions for tests. Please consider the suggestions. A new release is NOT required for only this suggestions:
https://github.com/chaozmc/ioBroker.oekofen-json/issues/19

chaozmc:
Hi, yes you're right i've run the adapter checker and got the Errors due to missing translations for uk in the admin folder and also the missing releaseconfig file. I thought that i have to create a new release with a new version as soon as i change something in the repository. The code itself is the same since Jannuary (1.0.1) and works fine without any issues since then.  

Regarding the new node versions, sure I'm going to add the tests for them

mcm1957:
Yes you are correct. You had to create a new release to get those fixes included. I only wanted to make sure that you die not functional change. Even minimal functional fixes can break the whole thing - but I'm sure you know that :-).

PR #2276, opened 2023-05-10
Updated hausbus_de adapter to actual stable version 1.2.1

hausbus:


mcm1957:
Please seperate changes to sources-dist and sources-dist-stable ALWAYS into seperate PRs.

And in this case please indiocate why your PR contains a addition of hausbus_de to sources-dist.json. As far as I see, this adapter is already available at latest repo. **So please remove sources-dist from this PR.**

I assume you used an old fork. Please use iobroker.dev to create PRs to avoid using outdated forks. It's only a one click action  there.

PR #2279, opened 2023-05-11
mediola gateway

oelison:
https://github.com/ioBroker/AdapterRequests/issues/47
#47

### DIFF
--- a/sources-dist-stable.json
+++ b/sources-dist-stable.json
@@ -1327,6 +1327,12 @@
     "type": "lighting",
     "version": "0.1.2"
   },
+  "mediola-gateway": {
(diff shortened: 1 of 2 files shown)


mcm1957:
Please seperate changes to sources-dist and sources-dist-stable ALWAYS into seperate PRs.

And in this case please do not enter the adapter to stable repository yet. As far as I can see, this adapter is new and not yet included into latest repository. New adapters need a review before they are added to latest repository. Thea mightr be added to stable only after some time at latest. See requirments listet at readme.md

**So please close this PR and create a PR requesting addition to latest.** You can (should) do this using developer portal.

PR #2287, opened 2023-05-16
Add/lightcontrol

Schmakus:


Schmakus:
Close because automatically with ioBroker.dev #2288

PR #2289, opened 2023-05-16
Update countdown adapter to version 2.0.0

jack-blackson:


mcm1957:
Latest 7.5.
Innstalled 77 or app 7%

mcm1957:
@jack-blackson 
This is a major Release. I coulg not read aby statement which breaking changes exist. Or an explicit statement that no incompatibilities exost. This could confise users.

@Apollon77 
Ok to merge anyway?

jack-blackson:
@mcm1957 The big change is that the whole order of what is done when - including "waiting"  was corrected, to avoid functions overtaking others, and too many updates of the same object. It is not a "completely new" one - but it should show that the adapter is still alive and I am motivated to work on new features - some of them already included in the 2.0.0.

jack-blackson:
Where should I put such a kind of statement regarding incompatibilities - or like here - that there are none? In the readme?

mcm1957:
> Where should I put such a kind of statement regarding incompatibilities - or like here - that there are none? In the readme?

Yes
The readme is the right place. I think a new release is not required. Would be sufficient to add it at readme in github head in 2
0.0 section so that users looking for incompatibilities can find it. 

@apollon77 ok?

Reason for asking is that a major release most likely contains breaking changes. If there are none a major release is ok if there are other reasons like this major reqrite. But user should get the info as a major release change should be done only after checking if its ok  for the personal environment.

Apollon77:
All good, and yes @jack-blackson with "semantic versioning" you should increase major only if you have breaking changes, minor for new features and micro/patch version for fixes only.

PR #2304, opened 2023-05-19
Update device-watcher to 2.9.2

ciddi89:
Please update my adapter ioBroker.device-watcher to version 2.9.2.


mcm1957:
2.9.2 was released at 19.05.2023
Normally a new release should be one or two weeks at latest before adding oit to stable.
If this release is a urgent hot fix please indiacte.

Otherwise I will delay merging a litle bit ..

ciddi89:
wtf? since when?? it's importent and fixes important issues for an smooth work.

Apollon77:
I can only repeat it also here:The question from @mcm1957 would have been the same I would have asked ... lso updating "just dependencies" introdued a lot of fun in some cases ...

PR #2314, opened 2023-05-28
add web frontend usage to README.md

oelison:
* add adapter to lists with web frontend [no-ticket]


mcm1957 (commented):
Short review from my side:
Thanks for contribution, content looks good 

I would suggest the following small adaptions:
- [ ] sentences should start with capital letter like at other paragraphs
- [ ] usage for webfrontend for stable should be added too (might be second PR, could be done by me later too)

Issue #2338, opened 2023-06-06
readyForStable: log clickable link


mcm1957:
The script should log a clickable linkt for those adapters which seem to be ready for update to ease access to repo for detail investigations. Currently only owner and name of adapter are logged.

PR #2341, opened 2023-06-06
readyForStable.js: log url next to adaptername

mcm1957:


mcm1957:
solves: https://github.com/ioBroker/ioBroker.repositories/issues/2338

PR #2366, opened 2023-06-16
Added awtrix light to beta

klein0r:
See https://forum.iobroker.net/topic/65571/test-adapter-awtrix-light


klein0r:
RE-CHECK!

klein0r:
RE-CHECK!

mcm1957:
RE-CHECK!

mcm1957:

- consider using adapter.setTimeout / this.setTimeout instead of (standard) setTimeout
  
- check and limit configurable timeouts / intervals
  
  historyAppsRefreshInterval seems to have no upper limit which could cause an overflow and a n immidate trigger - at least in Theory if user enters some unreasonable big values. Consider to limit maximum value.

- adapt testing to supported node releases

  Tests for node 16 is mandatory unless you require node 18 and higher. (already OK)
  Tests for node 18 is mandatory as many users already use node 18 and node 16 will be eol and deprected Q3/2023. (already OK)
  
**NOTE: none of those suggustion should be blockimng for merging the PR.** 

McM1957

klein0r:
@mcm1957 Thanks, fixed in 0.1.1

PR #2376, opened 2023-06-20
promote ioBroker.ical to stable 0.13.3 version

jens-maus:


mcm1957:
1.13.3 from 20.6.2023 1071 (6%) user
OK

PR #2377, opened 2023-06-20
Set stable version of adapter Alpha-ESS to 1.0.0

Gaspode69:


mcm1957:
1.0.0 from 20.6. But only version and rn changes since veto from end of nax

ÖĶ

PR #2378, opened 2023-06-20
add gigaset-elements 0.3.0 to stable

matthsc:
Forum thread: https://forum.iobroker.net/topic/51917/neuer-adapter-gigaset-elements

It has been in latest for way to long and should finally be moved to stable.

### DIFF
--- a/sources-dist-stable.json
+++ b/sources-dist-stable.json
@@ -698,6 +698,12 @@
     "type": "geoposition",
     "version": "1.2.4"
   },
+  "gigaset-elements": {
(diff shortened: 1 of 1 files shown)


matthsc:
RE-CHECK!

matthsc:
I've updated the LICENSE year in git, is that enough or do we really need a new release?

The build/main.js checks seem a little to eager, those are build files and part of the npm package, but I did not commit them to git.

mcm1957:
Correction the copyright year is pk. Needs NO new release.

mcm1957:
@Apollon77 
Please review

Adapter is new to stable. Semms to have same env as #2379 but checker is satisfied here.

My possibilitues at mobile are too limited for analyses

PR #2379, opened 2023-06-20
add innoxel 0.3.1 to stable

matthsc:
Forum thread: https://forum.iobroker.net/topic/50871/neuer-adapter-innoxel
No replies there unfortunately in nearly one and a half years.

I would suggest to move it to stable anyways, in order to increase potential visibility of this adapter. Maybe it will find some users besides myself. I've been using this adapter for about 4 years now.

### DIFF
--- a/sources-dist-stable.json
+++ b/sources-dist-stable.json
@@ -1016,6 +1016,12 @@
     "type": "iot-systems",
     "version": "1.2.6"
   },
+  "innoxel": {
+    "meta": "https://raw.githubusercontent.com/matthsc/ioBroker.innoxel/main/io-package.json",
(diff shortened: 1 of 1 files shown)


matthsc:
The build/main.js checks seem a little to eager, those are build files and part of the npm package, but I did not commit them to git.

mcm1957:
@apollon77
Please review and decide further steps.

Adapter is new for stable and seems tonhave some nonstandard env. See errors reported by checker.

Apollon77:
@matthsc In most cases we also check in build files into git to allow users to install directly from GitHub. Without the file this is not possible.

PR #2387, opened 2023-06-26
bump nuki-extended to 2.6.5

mcm1957:


mcm1957:
issues caused by migration to commmunity-adapters.
Will be fixed with next release

PR #2393, opened 2023-06-27

foxthefox:
update lifx 1.0.1


mcm1957:
1.0.1 from 20.09.2022
OK

PR #2398, opened 2023-06-29
Add ioBroker.residents to stable repo

jpawlowski:
First stable version


mcm1957:
Released 29.6.232
11 (40%) User

OK

mcm1957:
RE-CHECK!

PR #2399, opened 2023-06-30
add adapter hydrawise to latest

SentiQ:
https://github.com/ioBroker/AdapterRequests/issues/195

### DIFF
--- a/sources-dist.json
+++ b/sources-dist.json
@@ -882,7 +882,7 @@
   "hydrawise": {
     "meta": "https://raw.githubusercontent.com/SentiQ/ioBroker.hydrawise/master/io-package.json",
     "icon": "https://raw.githubusercontent.com/SentiQ/ioBroker.hydrawise/master/admin/hydrawise.jpg",
-    "type": "lighting"
+    "type": "garden"
   },
   "hyperion": {
     "meta": "https://raw.githubusercontent.com/ruhigundrelaxed/ioBroker.hyperion/master/io-package.json",
--- a/sources-dist.json
+++ b/sources-dist.json
@@ -879,6 +879,11 @@
     "icon": "https://raw.githubusercontent.com/Chris-656/ioBroker.huum-sauna/main/admin/huum-sauna.png",
     "type": "climate-control"
   },
+  "hydrawise": {
+    "meta": "https://raw.githubusercontent.com/SentiQ/ioBroker.hydrawise/master/io-package.json",
+    "icon": "https://raw.githubusercontent.com/SentiQ/ioBroker.hydrawise/master/admin/hydrawise.jpg",
+    "type": "lighting"
+  },
   "hyperion": {
     "meta": "https://raw.githubusercontent.com/ruhigundrelaxed/ioBroker.hyperion/master/io-package.json",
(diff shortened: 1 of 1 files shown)


mcm1957:

- [ ] Readme.md is very short

  The README.md of this adapter is very minimalistic. Theres no information about the functionality and the configuration inside.

  Please add a short description about the functionality of the adapter,  add links the the manufacturer site (if applicable) and add a description of the configuration of the adapter. If you provide an external documentation - which should be located at doc/xx directory - please add links to at least the english documentation. Also describe how to get the api-key required for configuration.

  ```
  "encryptedNative": [
    "password",
    "data2"
  ],
  "protectedNative": [
    "password",
    "data2"
  ],
  ```
  I refer to the apiKey which seems to be an id which is (or at least should) not be public.

```
function name2id(pName) {
}
```
  You retrive date from the api and  create states depending on the data returned. This data must be checked to contain only valid characters.

- [ ] reevaluate state roles

  In addition the roles MUST match the read/write functionality. As an example a role value.* requires that the state is a readable state. Input states (write only) should have roles like level.*. Please read the description carefully.
  
- [ ] error text seems to be wrong

   line 451 reads ```reject("Device IP is not configured");``` but the error should indoicate a missing api key.

- [ ] do not mix this.setTiemout / clearTimeout

   You are using this.setTimeout (what is OK). Bute please also use this.clearout to cancel those timers
 
  
McM1957

SentiQ:
@mcm1957 I went through your suggestions and made a few changes in my code

mcm1957:
RE-CHECK!

mcm1957:
- [ ] I detected some potential typo / cut and paste error:

  The id 'stopZone' indicates something is to be stopped. So the role 'button.start' seems to be incorrect. The same refers to  state 'schedule.stopall'. 'button.stop' is a valid role too so this role seems to match better if I understand the functionality correct.

- [ ] And please consider fixing
  👀 [W505] setTimeout found in "build/main.js", but no clearTimeout detected

  Although you are using this.setTimeout which is designed to clear pending timeouts (intervals) its considered a good praxis to explicitly clear all timeouts (intervals) at onUnload and leave the ioBroker class internal cleanup only as a safeguard.

  I leave this point to @Apollon77 to decide whether this is mandatory - or feel free to mention him directly for any discussion about this point.

SentiQ:
RE-CHECK!

SentiQ:
@mcm1957 I reviewed and fixed your suggestions

mcm1957:
OK, lgtm

mbhomie007:
Hello @Apollon77 @mcm1957

I hope you can integrate the adapter to latest without any trouble.
I will test it soon.
Thank you.

mcm1957:
@SentiQ 

SentiQ:
> @SentiQ
> 
> This adapter has been released to latest repository and should be visible within 24h maximum.
> 
> Please create a thread at https://forum.iobroker.net/category/91/tester titled like "Test Adapter " to collect some user feedback and provide a link to this topic when requesting addition to stable repository later.
> 
> Note: If an other testing topic already exists, ist OK to continue using this topic too.

It seems I don't have the right to create a new thread?

mcm1957:
Possible
Please contact HOMORAN at forum to get matching rights.
Either in any other topic or by chat.
And state that you are devbof adapter snd want to create test topic

Thanks

PR #2190, opened 2023-03-24
add ioBroker.apg-info to repo

HGlab01:

PR #2401, opened 2023-06-30
add apg-info to latest

HGlab01:
https://github.com/ioBroker/ioBroker.repositories/pull/2190 was closed by accident
this is so to say the reopen

### DIFF
--- a/sources-dist.json
+++ b/sources-dist.json
@@ -89,6 +89,11 @@
     "icon": "https://raw.githubusercontent.com/xhunter74/ioBroker.apcups/master/admin/ups.png",
     "type": "hardware"
   },
+  "apg-info": {
+    "meta": "https://raw.githubusercontent.com/HGlab01/ioBroker.apg-info/main/io-package.json",
+    "icon": "https://raw.githubusercontent.com/HGlab01/ioBroker.apg-info/main/admin/apg-info.png",
+    "type": "general"
+  },
   "apple-find-me": {
     "meta": "https://raw.githubusercontent.com/PfisterDaniel/ioBroker.apple-find-me/master/io-package.json",
     "icon": "https://raw.githubusercontent.com/PfisterDaniel/ioBroker.apple-find-me/master/admin/find-me.png",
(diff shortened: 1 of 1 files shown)


mcm1957:
> **- do not use process.exit to terminate an adapter** How can I then terminate a cron-job steered adapter after the execution? This is in my point of view common practice in iobroker-adapters, isn't it? (e.g. https://github.com/rg-engineering/ioBroker.daswetter/blob/a8eda122b7c53f0b64db7ce2bdb98e1f4a60bb72/main.js#L104 or https://github.com/HGlab01/ioBroker.fuelpricemonitor/blob/9341a26918aebb66ad2007b83ac08bfd47079ac2/main.js#L99)

OK - I'll leave it open to be checked by Apollon77. Maybe the compact flag is not correct in this case. But running in compact mode ADN process.exit ist still incompatible in my oppinion.
 
> **- using console.log and this.log.debug (info only / non blocking)** console.log is also available as sentry log which is super helpful in debuggin when it comes to having only the sentry info. I am using this in other adapters (e.g. iobroker.tado) and it helps a lot

OK, was an info only anyway

> **- invalid characters should be filtered from object ids** Is taken into account in the libary that creates the states https://github.com/HGlab01/ioBroker-jsonExplorer/blob/ae65d046dc27c25d6d77f9fec8cabd01fbb31fd6/jsonExplorer.js#L218-L220
> 
> **- state Ids should not contain space** the names in stateAttr.js is not the name of the "physical" state but a logical name You can also see this feature in the library https://github.com/HGlab01/ioBroker-jsonExplorer/blob/ae65d046dc27c25d6d77f9fec8cabd01fbb31fd6/jsonExplorer.js#L216

OK, I did notz see, that any non 0-0a-zA-Z characters are replaced by _ outside of this adapter.

Apollon77:
@HGlab01 in fact process.exit is th" hard" way and yes - because scheduled adapters do not use compact mode currently!! - it is ok-ish .

best practice would be like https://github.com/iobroker-community-adapters/ioBroker.weatherunderground/blob/59fc204266638f2e686252b836c0ba29298c3ea4/main.js#L106 ... unless you target an js.controller >=4 then simply adapter.terminate is also fine

HGlab01:
> @HGlab01 in fact process.exit is th" hard" way and yes - because scheduled adapters do not use compact mode currently!! - it is ok-ish .
> 
> best practice would be like https://github.com/iobroker-community-adapters/ioBroker.weatherunderground/blob/59fc204266638f2e686252b836c0ba29298c3ea4/main.js#L106 ... unless you target an js.controller >=4 then simply adapter.terminate is also fine

this is exactly what I do, isn't it?
https://github.com/HGlab01/ioBroker.apg-info/blob/6184c845ba87a303e421215510803c9c24c49172/main.js#L68
or should I remove the `0` in `this.terminate(0)`

Apollon77:
aahh yes, then this is fine ... removing 0 would be good because so the automtic error code is asigned by adapter.js. For me all fine. Thank you.

As said: I think (would need to check) we added terminate back in js.controle 3.x or such ... so if you set a dependency to js.cintroler >= 4 you do not need the exit fallback at all

HGlab01:
understood!
An in error cases like
https://github.com/HGlab01/ioBroker.apg-info/blob/6184c845ba87a303e421215510803c9c24c49172/main.js#L48-L55
?
Also just this.terminate() or still with a code as from adapter framework point of view there is no error (code)

PR #2297, opened 2023-05-18
Added Telegram-Menu to latest

MiRo1310:

PR #2403, opened 2023-06-30
Add telegram-menu to latest repo

MiRo1310:
#2297 

https://forum.iobroker.net/topic/64968/test-adapter-telegram-menu-v-0-3-x

### DIFF
--- a/sources-dist.json
+++ b/sources-dist.json
@@ -1140,14 +1140,14 @@
     "icon": "https://raw.githubusercontent.com/foxthefox/ioBroker.lifx/master/admin/lifx_logo.png",
     "type": "lighting"
   },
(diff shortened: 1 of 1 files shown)


mcm1957:
RE-REVIEW at new PR

  I cannot see any reference to ack flag near  https://github.com/MiRo1310/ioBroker.telegram-menu/blob/4c1277262f3996d44f6f83f4d016d2f0f353ae49/main.js#L143
  
McM1957

MiRo1310:
> 
Vielen Dank für das Feedback, 
ich habe den Code geändert und fange jetzt auch den Status des ack Flag ab.

mcm1957:
ack handling has been improved.

PR #2408, opened 2023-07-01
Intex 0.1.0 in service as a collaborator for TA2k

PLCHome:
The stable adapter is currently no longer usable due to the slow cloud from Intex. There is no connection anymore. 
This update uses a local connection and enables operation.


mcm1957:
0.1.0 released 25.06.2023
10 User (14%)

PR #2432, opened 2023-07-13
Bump DWD to 2.8.5

mcm1957:


mcm1957:
ref: https://github.com/ioBroker/ioBroker.dwd/issues/169

PR #2434, opened 2023-07-13
Bump VIS-LCARS to 1.1.1

mcm1957:


mcm1957:
ref https://github.com/ioBroker/ioBroker.vis-lcars/issues/15

PR #2437, opened 2023-07-14
Bump OCPP to 0.12.5

mcm1957:


mcm1957:
errors reported not relevant for stable

PR #2446, opened 2023-07-17
Bump TELEGRAM to 1.16.0

mcm1957:


mcm1957:
E506 is incorrect - everything is fine

PR #2448, opened 2023-07-17
Add TibberLink to latest

hombach:
Add new adapter TibberLink to latest repository

### DIFF
--- a/sources-dist.json
+++ b/sources-dist.json
@@ -2242,6 +2242,11 @@
     "icon": "https://raw.githubusercontent.com/ioBroker/ioBroker.text2command/master/admin/text2command.png",
     "type": "logic"
   },
+  "tibberlink": {
+    "meta": "https://raw.githubusercontent.com/hombach/ioBroker.tibberlink/master/io-package.json",
+    "icon": "https://raw.githubusercontent.com/hombach/ioBroker.tibberlink/master/admin/tibberlink.png",
+    "type": "energy"
+  },
   "tileboard": {
     "meta": "https://raw.githubusercontent.com/ioBroker/ioBroker.tileboard/master/io-package.json",
     "icon": "https://raw.githubusercontent.com/ioBroker/ioBroker.tileboard/master/admin/tileboard.png",


mcm1957:
This adapter is not yet at latest.

An adapter must be available at latest for some time before adding it to stable.
Please open a new issue after
-) your adapter is available at stabel
-) some users have used it without issues
-) a test thread at the forum is present and has some feedback.

see https://github.com/ioBroker/ioBroker.repositories#requirements-for-adapter-to-get-added-to-the-stable-repository

mcm1957:
reopened and title changed as the adapter is added to LATEST repository anyway.

hombach:
Hi Martin - First of all - THANK YOU for the time and effort you spend to verify this adapter. I really apreciate all the work you're doing with and for ioBroker!!

I fixed several of your proposals... but

- [x] crude sentence in readme


    -> looks like this is optional, but it's better to encrypt, I'll do this in a upcoming step


    -> I don't get your point - are you talking about state names like "1" ... that are the days per month I wanna keep this (caused by no better idea)
    -> Or did you notice "tibberlink.0.Homes.26b5774b-07ea-47f0-b816-124628...." that's the ID of a home in your Tibber account... could be changed in case there's only one home - but if you have several of it, changing the ID would be tricky

- [x] unused onStateChange handler

- [x] reevaluate state roles

- [x] log messages should / must be in english

BR, Christian

mcm1957:
-> I don't get your point - are you talking about state names like "1" ... that are the days per month I wanna keep this (caused by no better idea)
-> Or did you notice "tibberlink.0.Homes.26b5774b-07ea-47f0-b816-124628...." that's the ID of a home in your Tibber account... could be changed in case there's only one home - but if you have several of it, changing the ID would be tricky

Well maybe I did not understand your code in detail and everything is OK anyway. Both of your examples are OK.
It looked like you generate the id of multiple infomation at getStatePrefix. The paramater "space" is a little bit confusing too as one or more spaces between dots might cause trobles. But at a second look, you seem store the id seperated in the genearated object. And the description and name part is free text.

So if you the only variant is the tibber id in the format specified above thats completly OK.

hombach:
Well - now I understand.... 
`protected getStatePrefix(homeId: string, space: string, name: string):`
the space string is just the "folder"... like:  _tibberlink.0.Homes.26b5774b-0815.**LiveMeasurement**.lastMeterConsumption_

in version 0.1.4 also the API-Token is encrypted - so all of your proposals are DONE  ---  Thank you for your support and time!

hombach:
is "must be fixed" label still correct? What has to be changed?

mcm1957:
One remark to:

_sensitive information, especially passwords should be stored encrypted
-> looks like this is optional, but it's better to encrypt, I'll do this in a upcoming step_

As this change requires users to add the api token again (it will be invalidated when changing the encryption) its a good idea to either do it as soon as possible (with as less users as possible to be effected) or to decide that this informatoin is not critical.

mcm1957:
RE-CHECK!

PR #2449, opened 2023-07-17
updated kecontact adapter to 2.0.1

Sneak-L8:


mcm1957:
@Sneak-L8 
Please avoid issuing PRs with al lotz of outdated commits as this cause much more effort to check.

Either sync your fork befor changing stable.json or even mor comfortable use www.iobroker.dev, select you adapter and release and click at SET STABLE. You can also edit directly at the iobroker github repo which will also create a clean baranch.

Thanks in advance

Sneak-L8:
Thanks for your information. Sorry, I did a sync fork before editing. Perhaps I was too quick editing in IDE before all syncs were completed.
But setting stable versions via iobroker.dev is good news. I didn't know about it. But that's awesome. I will do it that way in future releases. Great stuff!

PR #2458, opened 2023-07-24
bump BLE to 0.13.4

AlCalzone:
The feedback seems to indicate that the latest `noble` version is finally working, so we can mark this as stable now.


AlCalzone:
looks like I have some cleaning up to do :D

AlCalzone:
Isn't most of that checked by looking at `io-package.json` on `master`?

mcm1957:
Your right. At least E150 and E152 are likely to be used from github. Not quite sure but likely.

The js controller check is low prio.

I l go on and merge when I m at PC. I avoid merging from Handy as I could not do muchbif any error occures.

mcm1957:
RE-CHECK!

PR #2488, opened 2023-08-06
Update sainlogic stable to 0.10.5

phifogg:


mcm1957:
1.10.4 created 23.7.2023
8 user

1.10.5 created 23.7.2023
77 user

mcm1957:
@phifogg 
Are you sure that an update to 1.10.4 is desired?
1.10.5 has been created at same date and hence 1.10.5 has been tested latest sinde 23.7.

mcm1957:
@phifogg
Are you sure that an update to 1.10.4 is desired?
1.10.5 has been created at same date and hence 1.10.5 has been tested at latest since 23.7.

phifogg:
AH.. good point. I'll update to 10.5 for stable. I didn't expect people to update that quickly :D

PR #2490, opened 2023-08-07
Update easee Adapter in stable

Newan:


mcm1957:
@Newan 
Looks like you deleted this PR on error.

I suggest to use www.iobroker.dev gui environment to add a new release into stable as this ensures creation of dedicated branch. And everything is done by a simple click. GO to adapter - releases - and click at "set as stable"

PR #2526, opened 2023-08-19
bump b-control-em to 0.3.1

mcm1957:
update required for js-controller 5


mcm1957:
ok lt. Apollon77 / telegramm

PR #2523, opened 2023-08-19
Update stiebel-isg to 1.7.7 and energymanager to 1.4.2

unltdnetworx:
New confirmed stable version for stiebel-isg and energymanager.

PR #2530, opened 2023-08-20
Update stiebel-isg to 1.7.7

mcm1957:
split off from #2523 as required for js-controller 5


mcm1957:
stiebel-isg
1.7.7 created 24.7.
26 user (5%)
OK

PR #2531, opened 2023-08-20
add frigate 0.2.5 to stable

Bettman66:


mcm1957:
2.0.5 added 20.07.2023
12 user (66%)
OK

PR #2541, opened 2023-08-24
bump history to 2.2.6

foxriver76:
needed for controller v5

### DIFF
--- a/sources-dist-stable.json
+++ b/sources-dist-stable.json
@@ -870,7 +870,7 @@
     "meta": "https://raw.githubusercontent.com/ioBroker/ioBroker.history/master/io-package.json",
     "icon": "https://raw.githubusercontent.com/ioBroker/ioBroker.history/master/admin/history.png",
     "type": "storage",
-    "version": "2.2.4"
+    "version": "2.2.6"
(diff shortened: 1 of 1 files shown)


mcm1957:
adapter checker error can be ignored for stable merge

mcm1957:
2.2.6 created 23.08.2023
157 user (0,5%)

@foxriver76 
When do you plan to release js-controller to stable?
I would suggest that histore should be available at latest at for a few (2 or 3 days) before going to stable, but if this is too long in your oppion I cn release it immidiatly too. You know best what has changed :-)

According to readme changes required for js-controll5 have been done end 2022 but never made into a relase. But I wonder why this did not pop up during latest tests of js-controller.

So please respond if you think history should go to stable immidiatly, otherwise I will do it at weekend

foxriver76:
Problem is, the previous version had adapter-core `^2.6.0` and in cases where it was not automatically installed to `2.6.8`, it has problems with controller v5, so super edgy.

foxriver76:
And weekend is fine.

mcm1957:
728 User
OK

PR #2550, opened 2023-08-27
update sureflap to 1.2.0

Sickboy78:
external API has become slower, need to increase timeout from 60s to 120s to avoid too much timeouts


mcm1957:
1.2.0 created 19.08.2023
28 user (>10%)
OK

PR #2551, opened 2023-08-27
Add amtronwallbox to latest

rg-engineering:
Please add my adapter ioBroker.amtronwallbox to latest.

### DIFF
--- a/sources-dist.json
+++ b/sources-dist.json
@@ -79,6 +79,11 @@
     "icon": "https://raw.githubusercontent.com/PArns/ioBroker.amazon-dash/master/admin/amazon-dash.png",
     "type": "hardware"
   },
+  "amtronwallbox": {
+    "meta": "https://raw.githubusercontent.com/rg-engineering/ioBroker.amtronwallbox/master/io-package.json",
+    "icon": "https://raw.githubusercontent.com/rg-engineering/ioBroker.amtronwallbox/master/admin/amtronwallbox.png",
+    "type": "energy"
+  },
   "anelhut": {
     "meta": "https://raw.githubusercontent.com/dan1-de/ioBroker.anelhut/main/io-package.json",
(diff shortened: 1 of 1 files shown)


mcm1957:
Related 
https://github.com/ioBroker/ioBroker.repositories/pulls?q=is%3Apr+amtronwallbox+is%3Aclosed

mcm1957:

- [ ] adapter requires node 16

  Due to usage of adaptercore 3.x.x this adapter requires node 16 or newer. Please merge PR https://github.com/rg-engineering/ioBroker.amtronwallbox/pull/74 to add engines requirement to package.json.  

- [ ] reevaluate state roles

  Please do not use general roles (like state, or value) only if no dedicated role is available.

- [ ] axios requires timeouts

  Please verify that you set a timeout at all axios calls. At least here (https://github.com/rg-engineering/ioBroker.amtronwallbox/blob/693e36944ae8173a360d57dd84ade94fed42aad3/main.js#L184) it seems to be missing. Axios defaults to "no timeout" which could case hanging adapters.

- [ ] adapter uses sentry

  As this adapter uses sentry evaluate whether you want to enble creating a sentry release  during deploy by uncommengting the relevant section: https://github.com/rg-engineering/ioBroker.amtronwallbox/blob/693e36944ae8173a360d57dd84ade94fed42aad3/.github/workflows/test-and-release.yml#L70 (only info, no blocking point)

  
McM1957

mcm1957:
RE-CHECK!

rg-engineering:
@mcm1957 with version 0.2.8 I've changed:
* link to amtron wallboxes added in readme
* check of node > 16 in package.json
* state roles overworked as good as I'm able to do (some states of adapter does not really match to iob spec)
* timeouts for axios configured

mcm1957:
RE-CHECK!

mcm1957:
ok

PR #2574, opened 2023-09-02
Bump TibberLink stable to 0.2.3

hombach:


hombach:
ok - this "brand-new" Label is wrong....

mcm1957:
Your right.
Mixed with 0.2.4 
Sorry

hombach:
I would think it' cause by my update commit in my fork... That was my fault... ;) Sorry!

PR #2577, opened 2023-09-04
Update ds18b20 stable to 2.0.4

crycode-de:
Major version increase since compatibility of old node.js versions is dropped and the admin ui is completely rewritten.
On first start after upgrading from v1.x the config will be migrated automatically.

v2.0.4 is required for js-controller 5 compatibility for users upgrading from ds18b20 1.x in some cases.


mcm1957:
2.0.4 created 20.08.2023
40 User ( 3,5%)
OK

PR #2579, opened 2023-09-04
bump linux-control to 1.1.5

mcm1957:


mcm1957:
special release for js-controller 5
checks ignoired

PR #2589, opened 2023-09-07
Add webui to latest

mcm1957:
**Originally created by [jogibear9988](https://github.com/jogibear9988)**

Reverts ioBroker/ioBroker.repositories#2588

This PR replaces the PR #2552 which has been merged on error

### DIFF
--- a/sources-dist.json
+++ b/sources-dist.json
@@ -2751,6 +2751,11 @@
     "icon": "https://raw.githubusercontent.com/Vertumnus/ioBroker.weblogin/master/admin/logo-login.png",
     "type": "utility"
   },
+  "webui": {
+    "meta": "https://raw.githubusercontent.com/iobroker-community-adapters/ioBroker.webui/master/io-package.json",
+    "icon": "https://raw.githubusercontent.com/iobroker-community-adapters/ioBroker.webui/master/admin/logo.png",
+    "type": "visualization"
+  },
   "webuntis": {
     "meta": "https://raw.githubusercontent.com/Newan/ioBroker.webuntis/main/io-package.json",
     "icon": "https://raw.githubusercontent.com/Newan/ioBroker.webuntis/main/admin/webuntis.png",


mcm1957:
Please see initial discussions at PR #2552

mcm1957:
Last comment by jogibear9988 https://github.com/ioBroker/ioBroker.repositories/pull/2552#issuecomment-1710350611


> First of all - THANK YOU for the time and effort you spend to maintain this adapter.
> 
> I would like to give some (preliminary) feedback based on my personal oppinion. This is NOT an offical review and @Apollon77 might have several additional suggestions or even a different oppinion to one or the other statement. In addtion this revie ONLY refers to the backend / adapter part - the frontend / www part must be reviewed by @Apollon77.
> 
> Please feel free to contact him (or wait for a response from him) if you cannot follow my suggestsions or before you spend major effort.
> 
> * [ ]  errors reported by adapter checker must be fixed / evaluated
>   `[E500] node_modules found in repo. Please delete it` refers to  https://github.com/iobroker-community-adapters/ioBroker.webui/tree/master/www/node_modules. node_modules inside adapter direvctory are invalid. Whether a node_moduls directory insode www tree is acceptable must be evaluated / decided by @Apollon77.
>   So please wait for his statement regarding this point.
> * [ ]  adapter has 'singleton' set to true / check compact attribute
>   As this adapter is allowed to have only one instance, creates subprocesses and seems to need significant resources please evaluate whether its really useful to enable compact mode for this adapter. Eventually set compact = false unless you see real benefits and have tested the adapter in compact mode.

-> I have tested it, and most of the time the adapter does nothing. Only when you add or remove a NPM package


> * [ ]  reevaluate read/write attribute of states
>   All states defined at io-package.json have write attribute set to true. Its unlikely, that all those states are designed to operate as input for the adapter and should be written ba other adapters, i.e. state.npm is a feedback state used during npm installation - if I read the code correctly. (https://github.com/iobroker-community-adapters/ioBroker.webui/blob/b49e969ebb297e288a0316a506bd4d95da12ed51/io-package.json#L251)

-> not all, some of them (the onse for the files have none, it's only the npm state who is wrong, I fixed it)

> * [ ]  onUnload code needs rewrite
> 
> https://github.com/iobroker-community-adapters/ioBroker.webui/blob/b49e969ebb297e288a0316a506bd4d95da12ed51/src/backend/main.ts#L201C1-L204C1
> 
> Rountine onUnload code must stop all running timers (seem none exist) and spawned subprocessed (i.e. npm installation). In addition onUnload must call the provide callback when completed. See templates or originally created adapter code for details.
> 
> In addtion evaluate the use of variable _onUnload. It is written once, but I did not find any usage.

-> I look that I cancel the processes

> 
> * [ ]  adapt testing to supported node releases
>   Tests for node 16 is mandatory unless you require node 18 and higher (at package.json).
>   Tests for node 18 is mandatory as this is the recommended node version and node 16 will be eol and deprected Q3/2023.
>   Consider adding node 20 tests unless you already know incompatibilities.
>   So the recommended testmatrix is [16.x, 18.x, 20.x] or [18.x, 20.x] depending on engines requirments setting at package.json
>   As node 16 is end-of-life now, it's ok to remove node 16 from testing. In this case please require node 18 as minimum node version at package.json. Currently node 16 is required / allowed) but not tested.
> 
> In general I must state once more, that this review ONLY refers to the adapter / backend part. The frontend / www part need to be reviewed by @Apollon77.
> 
> Thanks for reading and evaluating this suggestions. McM1957
> 
> **Please add a comment when you have reviewed and fixed the suggestionsor at least commented the suggestions and you think the adapter is ready for a re-review!**
> 
> **Additional notess addressed at @Apollon77 - please consider them during your review**
> 

-> I will not add test for 16

> * [ ]  delivering node_modules below www
>   Please check whether delivering node_modules below www directory tree is ok or not and eventually adapterchecker should be adapted to handle this (if delivery is ok)

-> I wll not look into adapterchecker

> * [ ]  using state as command interface
>   Is using several states as command interface ok? Or wouldn't this be handled by message in a better way?

-> I would it have similar to vis, so I leave it like this

> * [ ]  on the fly installation (running npm from adapter)
>   Is it ok that an adapter installs npm packages while the system is running?
>   Are there sufficient firewalls that the due to some error or user misuse the adapter cannot open scurity holes (i.e. installing anything without sufficient access rights of the user at the system) ?
>   Are there sufficient firewalls that npm installation does not interfere with other adapters or ioBroker core ?

-> I could use Security of iobroker (if there is one), but yes it is needed like it is in the adapter. 

> * [ ]  complete frontend (www) including installation of npm and uploading needs review. I cannot check this due to missing knowhow in this area. Apollon77 please take this part.

mcm1957:
-> I will not add test for 16

As I wrote already node 16 not required, but in this case adapter requirements to node 18. Adapetr should nozt allow installation with an untested node release. Sorry, if my statement wasn't clear enough.

mcm1957:
I repeat the last paragrapgh below. You might have missed the header. This üaragraph is addressed to Apollon77, no need to change anything without a feedback from him. I do not expect that you do anything at adapter checker of course.

Whether there could be a security hole created when installing the adapter must be evaluated by Apollon77 too (as far as I read your response, the adapter has no security and hence anybody having access to the web interface or anybody who could write to the states could install potential dangerous npm packages without having elevated rights (root, iobroker account, etc.). If this is correct it looks dangerous. But this discussion should be done with @Apollon77. I only want to create some focus on it - I cannot decide whats good practice within ioBroker.


**Additional notes addressed at @Apollon77 - please consider them during your review** 

- [ ] delivering node_modules below www

  Please check whether delivering node_modules below www directory tree is ok or not and eventually adapterchecker should be adapted to handle this (if delivery is ok)

- [ ] using state as command interface

  Is using several states as command interface ok? Or wouldn't this be handled by message in a better way?

- [ ] on the fly installation (running npm from adapter)

  Is it ok that an adapter installs npm packages while the system is running? 
  Are there sufficient firewalls that the due to some error or user misuse the adapter cannot open scurity holes (i.e. installing anything without sufficient access rights of the user at the system) ?
  Are there sufficient firewalls that npm installation does not interfere with oterh adapters or ioBroker core ?

- [ ] complete frontend (www) including installation of npm and uploading needs review. I cannot check this due to missing knowhow in this area. Apollon77 please take this part.

jogibear9988:
any news ?

mcm1957:
Review by Apollon77 accepted (via telegram)

PR #2610, opened 2023-09-09
Bump iwg-vpn stable to 2.0.4; changed type

iwg-vpn:


mcm1957:
RE-CHECK!

Apollon77:
@iwg-vpn As discussed in the forum already I would liek to request some more transparency.

Please add, at least, informations which data are transferred to the skill and who is the sole owner of the skill and lambda services to your Alexa Readme. 
Please also consider adding an enable(disable (disabled by default!) setting for this amazon integratzion. When I roughly understand the code right the objects are always pished to your lambda? Could this be? Or where it is checked that it is not enabled and so no data are exposed?

mcm1957:
**Please add a comment when you have reviewed and fixed the suggestions or at least commented the suggestions and you think the adapter is ready for a re-review**

iwg-vpn:
readme extended, configuration with per default disabled Alexa integration added. Everything now is in the 2.0.**_4_**

mcm1957:
Technical feedbak: Please update PR. It still contains 2.0.3
Not quite sure if all questions by @Apollon are answered. But this will be checked by him

Apollon77:
@iwg-vpn please solve conflicts and add 2.04 here. thank you

iwg-vpn:
version updated. not sure what conflicts you mean though

PR #2624, opened 2023-09-14
Bump pushbullet stable to 1.0.1

mcm1957:


mcm1957:
priority release due to js-controller 5

PR #2645, opened 2023-09-17
iwg-vpn bumped up to 2.0.4

iwg-vpn:


Apollon77:
can oyu please incorporate into #2610 ?

mcm1957:
I will close here
Please adapt #2610 (= change version to 2.0.4) which is still open and resolve conficts there as @Apollon77 suggested.

PR #2647, opened 2023-09-17
bump admin stable

foxriver76:


mcm1957:
6.10.1 created 11.9.
OK

PR #2670, opened 2023-09-21
Bump TibberLink stable to 0.3.2

hombach:


mcm1957:
0.3.2 created at 18.09.2023
20 user
OK as 4.0.0 already block further testing of 0.3.x

PR #2671, opened 2023-09-22
Update discord stable to 2.1.2

crycode-de:


crycode-de:
W171 and W170 will be fixed in next release as in my opinion it's not worth creating a new release for this.

W505 -> `Adapter.setTimeout(...)` is used, so no `clearTimeout(...)` should be needed.

mcm1957:
2.1.2 created 04.09.2023
OK

PR #2682, opened 2023-09-24
Bump oilfox to 4.2.0 at stable

mcm1957:


mcm1957:
4.2.0 created 08.09.2023
46 User (>10%)
OK

PR #2686, opened 2023-09-25
Bump growatt 3.2.1 to stable

PLCHome:


mcm1957:
3.2.1 created 08.09.2023
82 User
OK

mcm1957:
Please add node 20 test for next release
Please add node 16+ requirement for next release or create a new one including the requirement now if you are afraid that users are trying to install at node 14 (and fail)

PR #2688, opened 2023-09-25
Update iobroker.residents to 0.1.1

jpawlowski:


mcm1957:
Ok - already done as I noticed

PR #2694, opened 2023-09-29
Add renacidc to latest

raschy:
Please add my adapter ioBroker.renacidc to latest.


mcm1957:

  The README.md of any adapter related to device or multiple devices of a manufacturer should contain a link the manufacturers website and/or to a description of the device. This is to enable new users to easily check wether they associate the adapter with the correct device(s). (See https://github.com/ioBroker/ioBroker.repositories#requirements-for-adapter-to-get-added-to-the-latest-repository paragraph 4.)

- [ ] module apiClient has empty lines

  consider removing some empty lines from apiCLient (non blocking)

```
function name2id(pName) {
}
```
  As object ids are built using stationId and key you should filter for invalid charactes transmitted by web service. There's always a possibility that the external service changed the data returned and even such a case should not generate invalid states.


- [ ] reevaluate state roles

  In addition the roles MUST match the read/write functionality. As an example a role value.* requires that the state is a readable state. Input states (write only) should have roles like level.*. Please read the description carefully. States with role 'button' should (must) have attribute 'common.read=false' set. A button ( "Taster" in german) only triggers when you press but else has no state), so it should also be not readable. This is also like HomeMatic does it. A switch has clear states true or false and so can be read.

  Please avoid using general roles (state, value) whenever a dedicated role exists. It looks like you are always using role = value. Setting correct roles (and units) should be possible using a table for a list known keys. If the webservice adds new keys later, one could add the role/unit Info with a new adapter release later.

McM1957

mcm1957:
@raschy 
Do you need any help?

raschy:
I have edited the comments in the first adapter review in the points 'Readme.md', the interception of 'Forbidden_Chars' and the specification of the 'stateroles' to the best of my ability. I could not understand the reference to the 'setTimeout'. 
While working through the empty lines in 'apiClient.js', I came across the warning in VSC: "Property 'default' does not exist on type 'AxiosStatic'. Did you mean 'defaults'?". (See also in starters-adapter-development'.) I can't find a solution to this.

raschy:
@mcm1957 Hopefully I have corrected all the deficiencies. Please review again.

mcm1957:
> I could not understand the reference to the 'setTimeout'.

js provides a set of functions setTimeout / clearTimeout / setInterval / clearInterval

ioBroker adaptercore provides warppers for thos functions named identically but located at the adapter namespace. So ypu can use them by writing adapter.setTimes (classicmode) or this.setTimeout (class mode). The benefit of using the wrappers is, that the adaptercore has some firwealls built in to handle "forgotten" cleanups and has (or will have) checks for parameter valueas so that i.e. values too big do not result in immidate execution or something like this. Simpl adding a tzhsi. in formt of setInterval and clearInterval should do the magic. 

I do not consider this blocking - take it as a suggestion for next revision.

mcm1957:
BUT I think I detected a new issue.

At https://github.com/raschy/ioBroker.renacidc/blob/bd4327bed1585b12b7186aa9f02f9d1926f0ccef/main.js#L115 you remove invalid characters from the key. But I think this conversion is missing for deleteDeviceState. Please check - I do not consider this as blocking too but only as a normal issue / bug (if ist not ok at all anyway).

mcm1957:
RE-CHECK!

mcm1957:
Your adapter has been released to latest repository lately.

Please create a forum topic at https://forum.iobroker.net/category/91/tester named 'Test adapter xxxx' to invite users to test the adapter. Some feedback at forum is desired before adapter is released to stable repository in future.

If some other thread discussing the adapter already exists, its ok too - you do not need to open a second topic in this case.

PR #2724, opened 2023-10-06
Bump Plex to 1.1.3 at stable

ticaki:
no hurry :)


mcm1957:
1.1.3 created 05.10.2023

mcm1957:
@ticaki 
Nothing special - your PR was only used as a test for the new auto-info function :-)

PR #2735, opened 2023-10-09
Update lifx to 1.0.3 at stable

foxthefox:
update lifx 1.0.3

### DIFF
--- a/sources-dist-stable.json
+++ b/sources-dist-stable.json
@@ -1289,7 +1289,7 @@
     "meta": "https://raw.githubusercontent.com/foxthefox/ioBroker.lifx/master/io-package.json",
     "icon": "https://raw.githubusercontent.com/foxthefox/ioBroker.lifx/master/admin/lifx_logo.png",
     "type": "lighting",
-    "version": "1.0.1"
+    "version": "1.0.3"
   },
   "lightcontrol": {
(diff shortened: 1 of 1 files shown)


mcm1957:
@foxthefox 
Your adapter https://github.com/foxthefox/ioBroker.lifx does not use standard iobroker testing using node 16/18/20

Please see requirements https://github.com/ioBroker/ioBroker.repositories#requirements-for-adapter-to-get-added-to-the-latest-repository which is also required for stable repository.

Adapter needs to have at least Adapter basic testing (installing, running) using GitHub actions. More information in forum from apollon77 (Just take from other adapters the samples)

Currently your testmatrix only contains node 18. Iobroker officially supports node 16 / 18 / 20. Tests should be run for every node version supported. So please add node 16 and node 20 to your test matrix and retrigger a test. 

If you do not or cannot support node 16, it's ok the add an 'engine: >=18' clause to package.json so that node 18 or newer ist required during installation.

If you know that you adapter fails with node 20 temporary omit node 20 tests but add an issue describing the problem and try to fix it as soon as possible.

Please feel free if you need any help setting up standard testing and / or if you prefer a that we / I create a PR to fix this.

Please add a comment when you have reviewed and fixed the suggestion or at least commented the suggestions and you think the adapter is ready for a re-review

foxthefox:
testing is now with 16,18,20

mcm1957:
1.0.3 created 24.09.2023
OK

PR #2751, opened 2023-10-12
Update fritzdect to 2.5.2

foxthefox:


mcm1957:
2.5.2 created 12.10.2023

mcm1957:
@foxthefox 
There seems to be some issue with 2.5.2
https://github.com/foxthefox/ioBroker.fritzdect/issues/459
https://github.com/foxthefox/ioBroker.fritzdect/issues/460
https://forum.iobroker.net/topic/44473/fritzbox-dect-adapter/19

Please try to analyze / fix the problem. Does not make sense to push 2.5.2 to stable as long as that problem is not anylzed or solved.

**Please add a commetn as soon as the above problem is solved or you want to push 2.5.2 to stable anyway**

mcm1957:
@foxthefox
Problems mentioned above look like fixed with 2.5.3.

Please let me know wther a stable upodate to 2.5.2 would be really useful now. I suggest to wait for 2.5.3 to be old enough for stable and to close here.

foxthefox:
yes, 2.5.3 would be the next version for stable.

mcm1957:
Then lets close here.

Please open a new PR as soon as 2.5.3 is old eneogh and you think ist Ok too.

THANKS for sepnding your time for this adapter and ioBroker.

PR #2752, opened 2023-10-13
add homekittiles to latest

Standarduser:


Standarduser:
RE-CHECK!

Standarduser:
RE-CHECK!

Apollon77:
Please rename to haveve "vis-" in front of the name as with all other widgets. This is important for the users to understand that this is not a standalone adapter but a vis "addon". thank you

PR #2761, opened 2023-10-16
Add soliscloud adapter to latest

Trixx34:


Trixx34:
RE-CHECK!

mcm1957:
RE-CHECK!

mcm1957:

- [ ] Is 'tab:materialize' needed?

   Please check whether 
  ```
            "tab": "materialize"
  ```
  is really needed at io-package,json (https://github.com/Trixx34/ioBroker.soliscloud/blob/0d0ffb0f9eeddf93e68188638e0c7be16f9664e0/io-package.json#L84C1-L85C1)


```
function name2id(pName) {
}
```
  The best approach for new adapters is to allow a-z,A-Z,0-9,- and _ only. Please filter the user configurable platId and convert alle non a-zA-Z0-9-_ chars into i.e. an _

- [ ] reevaluate state roles

  I.e. 
  ```
  value.energy.consumed - energy consumed (unit=Ws, Wh, kWh)
  value.energy.produced - power produced (unit=Ws, Wh or kWh)
  ```
  will fit for some states. Please check all states for matching roles.


  Limit all intervals used to access cloud services to a reasonable value. If users configure an inapprobiate access rate this could lead to suspending the service by the provider for all ioBroker users.  
  
- [ ] standard github testing is missing

  standard github testing must be active and pass. You hab an incorrect head branch name at test-and.release workflow. I've generated a PR trying to fix this. Please merge and check that test is running and succeds (green checkmark) for all node releases (16 / 18 / 20)

 - [ ] adapte engines requirements

   As ioBroker reuires node 16 minimum, please adapt engines clause at package.json to require node 16 minimu
https://github.com/Trixx34/ioBroker.soliscloud/blob/0d0ffb0f9eeddf93e68188638e0c7be16f9664e0/package.json#L22
 
McM1957

mcm1957:
Did you have time to review commenst?

Trixx34:
Hi,

I did check them and was able to already fix a few things.
I hope I get to the other issues in the next few days.

Thanks again for your help in checking my code.

Regards.

<http://www.avg.com/email-signature?utm_medium=email&utm_source=link&utm_campaign=sig-email&utm_content=webmail>
Virus-free.www.avg.com
<http://www.avg.com/email-signature?utm_medium=email&utm_source=link&utm_campaign=sig-email&utm_content=webmail>
<#DAB4FAD8-2DD7-40BB-A1B8-4E2AA1F9FDF2>

Op zo 29 okt 2023 om 20:52 schreef Martin M. ***@***.***>:

Trixx34:
> * [ ]  Is 'tab:materialize' needed?
>   Please check whether
>   ```
>             "tab": "materialize"
>   ```
>   
-> Removed.

> * [ ]  invalid characters should be filtered from object ids
-> Fixed by only allowing numerical values for the plantID.

> 
> * [ ]  reevaluate state roles
>   Only the values specified here (https://github.com/ioBroker/ioBroker.docs/blob/master/docs/en/dev/stateroles.md) may be used as state roles. Do not invent new names as this might disturb the functionalyity of other adapters or vis.
>   In addition the roles MUST match the read/write functionality. As an example a role value.* requires that the state is a readable state. Input states (write only) should have roles like level.*. Please read the description carefully. States with role 'button' should (must) have attribute 'common.read=false' set. A button ( "Taster" in german only triggers when you press but else has no state), so it should also be not readable. This is also like HomeMatic does it. A switch has clear states true or false and so can be read.
-> I think these should be correct now.

> * [ ]  check and limit configurable timeouts / intervals
> * [ ]  check and limit intervals used to access cloud services
-> pollIntervall will only allow values between 45 & 1800 seconds.

> * [ ]  standard github testing is missing
>   standard github testing must be active and pass. You hab an incorrect head branch name at test-and.release workflow. I've generated a PR trying to fix this. Please merge and check that test is running and succeds (green checkmark) for all node releases (16 / 18 / 20)
-> PR fixed this.

> * [ ]  adapte engines requirements
>   As ioBroker reuires node 16 minimum, please adapt engines clause at package.json to require node 16 minimu
>   https://github.com/Trixx34/ioBroker.soliscloud/blob/0d0ffb0f9eeddf93e68188638e0c7be16f9664e0/package.json#L22
-> updated to node 16 minimum.

> * [ ]  some labels seem to have no translation into all languages
>   It seems that some labels are not translated into all supported languages, i.e. ...
>   Please consider adding missing translationsThanks for reading and evaluating this suggestions.
-> Fixed translations.

mcm1957:
RE-CHECK!

mcm1957:
Some (more) feedback

- [ ] -> Fixed by only allowing numerical values for the plantID. (not blocking)
  You should consider to add this check into the code. Config data could be written directly by changing the object tto bypassing any checks at jsonConfig.json. This is not blocking. Buts a giid praxis to check all parameters at adapter start (value and type) and not to trust any confgi limits.

-  [ ] write attribute must be false for "value*" **(must be fixed)**
   see https://github.com/Trixx34/ioBroker.soliscloud/blob/796fa27180e168d05932bc892976e0157aca04c7/src/main.ts#L28
  Roles look good. But most likely states are no input states and hence must have write:false attribute. IF they are really input states (states which ar used to control the inverter) the roles muste be LEVEL.*

- [ ] limiting intervalls (not blocking)
  as written above, such limits should be coded into source to avoid malfunction due to incorrect config. (i.e. a short time ago, admin had a bug so that it stored numeric values as strings. This caused really unexpected bahavior at adapters. If the adapter would have checkd the parameters a much more clear error message would have been logged)

- [ ] consider using integrated deployment (not blocking)
  Consider adding standard deploy workflow into test-and-release. This will work flowlessly with alcalzone/release-script.
  Of course this is a suggestion only and this will not block addition to repositories.

Trixx34:
@mcm1957 
Thanks for checking (again :) )

> * [ ]  -> Fixed by only allowing numerical values for the plantID. (not blocking)
>   You should consider to add this check into the code. Config data could be written directly by changing the object tto bypassing any checks at jsonConfig.json. This is not blocking. Buts a giid praxis to check all parameters at adapter start (value and type) and not to trust any confgi limits.

->You are correct, I didn't look at it that way. I've now implemented a replace using FORBIDDEN_CHARS before the plantid is created.

> * [ ]  write attribute must be false for "value*" **(must be fixed)**
>   see https://github.com/Trixx34/ioBroker.soliscloud/blob/796fa27180e168d05932bc892976e0157aca04c7/src/main.ts#L28
>   Roles look good. But most likely states are no input states and hence must have write:false attribute. IF they are really input states (states which ar used to control the inverter) the roles muste be LEVEL.*

-> switched the attributes to false.

> * [ ]  limiting intervalls (not blocking)
>   as written above, such limits should be coded into source to avoid malfunction due to incorrect config. (i.e. a short time ago, admin had a bug so that it stored numeric values as strings. This caused really unexpected bahavior at adapters. If the adapter would have checkd the parameters a much more clear error message would have been logged)

-> implemented an additional function to check if the config is correct and contains allowed values before setting the interval.

> * [ ]  consider using integrated deployment (not blocking)
>   Consider adding standard deploy workflow into test-and-release. This will work flowlessly with alcalzone/release-script.
>   Of course this is a suggestion only and this will not block addition to repositories.

-> Added and just used this for the first time, thanks for the suggestion :)

mcm1957:
RE-CHECK!

mcm1957:
> > * [ ]   consider using integrated deployment (not blocking)
> >   Consider adding standard deploy workflow into test-and-release. This will work flowlessly with alcalzone/release-script.
> >   Of course this is a suggestion only and this will not block addition to repositories.
> 
> -> Added and just used this for the first time, thanks for the suggestion :)

Just one remark: (NOT BLOCKING)

I do not see that you added anything to test-and.release workflow and I do not see a deploy job at actions. I'm quite sure that you pushe the latesr release suing npm publish manually.

I referred to the workflow parts you can find i.e. at this link
https://github.com/iobroker-community-adapters/ioBroker.pid/blob/b59fc9089f52147248408753c1d38eb0621ef84d/.github/workflows/test-and-release.yml#L54

If you add this part, the wirkflow will publish a new release without any extra action if checks pass.

But as already mentioned, ist up to you hwo you would like to process you releases. If you need any help, let me know. If you prefer I cann provide a PR with the "missing" parts.

mcm1957:

PR #2763, opened 2023-10-16
Bump lgtv to 2.1.1 at stable

mcm1957:


mcm1957:
2.1.1 created 06.10.2023
219 user
OK

PR #2806, opened 2023-10-19
Update sureflap to 1.2.1

Sickboy78:


mcm1957:
@Sickboy78 
As 1.2.2 has been released already do you really want to upgrade stable to 1.2.1 now instead of waiting for 1.2.2 to have enough feedback and update to 1.2.2 directly?

Please advise.

mcm1957:
1.2.1 created 03.10.2023
37 user (>10%)
OK

Sickboy78:
> @Sickboy78 edit: I just noticed that 1.2.2 is NOT released but only looks like ready at github

Correct, 1.2.1 is the latest version released on npm.
1.2.2 is work in progress.

> Please add node 20 tests for next release. I've created a PR for you.

Thx. Will do.

PR #2808, opened 2023-10-19
add iopooleco to latest

mule1972:
Add iopooleco Adapter

### DIFF
--- a/sources-dist.json
+++ b/sources-dist.json
@@ -1066,8 +1066,8 @@
     "type": "household"
   },
   "iopooleco": {
-    "meta": "https://raw.githubusercontent.com/ioBroker/ioBroker.iopooleco/main/io-package.json",
-    "icon": "https://raw.githubusercontent.com/ioBroker/ioBroker.iopooleco/main/admin/iopooleco.png",
+    "meta": "https://raw.githubusercontent.com/mule1972/ioBroker.iopooleco/main/io-package.json",
+    "icon": "https://raw.githubusercontent.com/mule1972/ioBroker.iopooleco/main/admin/iopooleco.png",
     "type": "meetering"
   },
   "iot": {
--- a/sources-dist.json
+++ b/sources-dist.json
@@ -1065,6 +1065,11 @@
     "icon": "https://raw.githubusercontent.com/TA2k/ioBroker.intex/master/admin/intex.png",
     "type": "household"
   },
+  "iopooleco": {
+    "meta": "https://raw.githubusercontent.com/ioBroker/ioBroker.iopooleco/main/io-package.json",
+    "icon": "https://raw.githubusercontent.com/ioBroker/ioBroker.iopooleco/main/admin/iopooleco.png",
+    "type": "meetering"
+  },
   "iot": {
     "meta": "https://raw.githubusercontent.com/ioBroker/ioBroker.iot/master/io-package.json",
     "icon": "https://raw.githubusercontent.com/ioBroker/ioBroker.iot/master/admin/iot.png",


mcm1957:
please fix urls at PR
```
https://raw.githubusercontent.com/ioBroker/ioBroker.iopooleco/main/io-package.json
````
repo owner is NOT iobroker but (most likely) mule1974

**Comment by mcm1957** [comment]
RE-CHECK!

**Comment by mcm1957** [comment]

  ```
  "encryptedNative": [
    "password",
    "data2"
  ],
  "protectedNative": [
    "password",
    "data2"
  ],
  ```
  The paramater 'apikey' sound like something which should be protected. Please check.

  To avoid publishing sensitive information when forwarding logs, those information should never be written into the logs but either omitted or at least partly masked so that the information cannot be misused. This relates at least to usernames, passwords, uniqui (user assigned) keys and tokens. In addition personal information like name, address, gegraphical coordinates should not be logged too. Refers to 'apikey'.

- [ ] avoid not required logging

  Logging at info, wraning or error level should only be performed if the information is really useful for the user.
  'this.log.info(`check if exists pooldevice: ${id.id}`)' is most likely a sort of debug information. Please log such info using level 'debug'

```
function name2id(pName) {
}
```
  You retrieve ids from the response of the website. Please check those ids and filter out invalid characters. Characters matching FORBIDDEN_CHARS must be filtered. As a good praxis for new adapter filetr out all characters other than a-zA-Z0-9,_ adn replace them with _ .

- [ ]processing of last timestamp
  Why do you retrive the state latestMeasure.measuredAt and process other attributes only if it changes? If you configure appropiate poll rates, it should not be a problem to update values even if they did not change, 

- [ ] folders must be created

  You create a state '.latestMeasure.temperature' but I do not see a plce where you create the folder '.latestMeasure'. see https://github.com/mule1972/ioBroker.iopooleco/blob/f32d3be7befbf20a393ea6359fbad45a19bb0e02/main.js#L95 

- [ ] reevaluate state roles

- [ ] unused variables

  Some variables seem to be unsused
  ```
		this.pollInterval = 0;
		this.pollTimeout = null;
  ```

- [ ] latestMeasure.isValid

  It looks like isValid is never set to false as states are only update if latestMeasure.isValid is trur. Please check.

- [ ] scheduled adapters should have random delays
  
- [ ] do not use cron for adapter which start very often
  
  You cron cponfig schedules the adapter to start every minute. If this is really desired / required (which seesm to be inappropiate) consider to move to a deamon mode. Overhead starting an adapter so often is too big. Schedules adapters should not start more then every 5 to 10 minutes.
  
  Limit all intervals used to access cloud services to a reasonable value. If users configure an inapprobiate access rate this could lead to suspending the service by the provider for all ioBroker users.  Polling every minute seems to be much to often. None of th epool paramaters will change significant witrhin such a short time
  
- [ ] do not access cloud services at fixed timestamps
  
McM1957

mule1972:
- [ ] Readme.md:
has been updated with link to manufacturer web page

- [ ] sensitive information: 
API key is now being encrypted and protected

- [ ] avoid not required logging: 
checked and updated

- [ ] invalid characters should be filtered from object ids: 
fixed

- [ ] processing of last timestamp: 
I would like to ensure that states only get updated if new measurement data is being received. For e.g. pools it is usual to close them for winter and meter devices like ECO from iopool are therefore set into a winter mode which means there is no more measurement data for months. Other scenarios are failure or maintenance of the meter device. For those scenarios I think it would not be a good idea to always update the states on regular basis without any new measurement data.

- [ ] folders must be created: 
fixed

- [ ] reevaluate state roles: 
checked it but did not find what you would like to change me. Therefore please point me to an example of what is wrong or should be changed.

- [ ] unused variables: 
fixed

- [ ] latestMeasure.isValid: 
Only valid measurements should be saved. Therefore latestMeasure.isValid has been removed from iobroker data set.

- [ ] scheduled adapters should have random delays: 
fixed

- [ ] check and limit intervals used to access cloud services: 
Official API guidelines from iopool defines min request frequence to 1 second (https://github.com/iopool/community/discussions/8) but I now changed the interval to every 15 minutes (usual measure frequence of the devices from iopool) and synced the scheduler with the measuredAT timestamp in order to get most current measurement data 

- [ ] do not access cloud services at fixed timestamps: 
see comments before

mcm1957:
RE-CHECK!

mcm1957:
- [ ] Please review loggings once more (non blocking)
  I do not consider it blocking, but logging i.e. start end end of every request is very likely filling the log without real information.
  see i.e. https://github.com/mule1972/ioBroker.iopooleco/blob/0e6158f68ce83ffaaa64098c97e9e30464708d7c/main.js#L35
  Logging at start of a daemon adapter would be acceptable - a scheduled adapter should only log whts really improtant as it would log with every start.

mcm1957:
@mule1972 

PR #2810, opened 2023-10-20
Fix icon path for govee-local (latest)

boergegrunicke:
Fix the icon path for govee-local, as this makes the repo-checker fail


mcm1957:
Thanks

PR #2818, opened 2023-10-21
bump hue to 3.10.1 at stable

foxriver76:


mcm1957:
3.0.1 created 17.10.2023
480 User
OK

PR #2820, opened 2023-10-22
Update panasonic-comfort-cloud to 2.2.4

marc2016:
AppVersion changed and there was an error in the automatic loading of the app version from Github. The previous adapter versions will no longer work if users do not manually enter the new AppVersion. With this version the automation works again.


mcm1957:
2.2.4 created 18.10.2023

PR #2822, opened 2023-10-23
Push 0.0.6 to stable

PLCHome:
44 Installation no issues.
The discussion in the forum is positive.
https://forum.iobroker.net/topic/68707/


PLCHome:
RE-CHECK!

PR #2824, opened 2023-10-23
Emergency Update viessmannapi to 2.3.0

TA2k:
Vendor requested a login flow update to reduce load on IAM system


mcm1957:
Thanks for adding releasenotes to README.md

PR #2827, opened 2023-10-26
Update Device-Reminder V3.0.0

Xenon-s:
Numerous changes Device-Reminder, because the old version was no longer executable with the new JS controller Update to V3.0.0


Apollon77:
fast-track because js-controller 5 issue

mcm1957:
@Xenon-s 

Thanks for fixing js.controlelr 5 issues.

**Please fix the following testing issues for next release:** (i've created a issue at the adapter repo as reminder)

- [ ] update standard testing (https://github.com/Xenon-s/ioBroker.device-reminder/pull/217).  PR is older than a year in the meantime.

- [ ] include node 20 into tests as node 20 is already released as stable verion unless this adapter fails using node 20. In this case please create a issue and try to fix ist as soon as possible.

- [ ] add engine>16 into package.json when updateing adapter core to 3.x.x as adapter core 3.x.x fails at node 14 and earlier.

Thanks in advance

PR #2831, opened 2023-10-27
Add zendure-solarflow to latest

nograx:
Please add my adapter ioBroker.zendure-solarflow to latest.


mcm1957:
Thanks for spening you time to develop an adapter for ioBroker.

I'll du a detailled review the next days, but in the meantime fix issues causing the standard github tests to fail.
- fix linter errors and warningd
- ensure that integration tests do not fail (Adapter should not terminate with error if config is missing, logging an error log ishould be enough
- please note that standard github tests are mandator, so removing them is not an option.
- 
image

mcm1957:
Just for Inof:
Theres no need to close the PR to process feedback

PR #2834, opened 2023-10-27
Add zendure-solarflow to latest

nograx:
Please add my adapter ioBroker.zendure-solarflow to latest.

### DIFF
--- a/sources-dist.json
+++ b/sources-dist.json
@@ -1917,11 +1917,6 @@
     "icon": "https://raw.githubusercontent.com/PLCHome/ioBroker.pylontech/master/admin/pylontech.png",
     "type": "energy"
   },
-  "radar2": {
-    "meta": "https://raw.githubusercontent.com/frankjoke/ioBroker.radar2/master/io-package.json",
-    "icon": "https://raw.githubusercontent.com/frankjoke/ioBroker.radar2/master/admin/radar2.png",
-    "type": "network"
-  },
   "radar-trap": {
     "meta": "https://raw.githubusercontent.com/Steiger04/ioBroker.radar-trap/main/io-package.json",
     "icon": "https://raw.githubusercontent.com/Steiger04/ioBroker.radar-trap/main/admin/radar-trap.png",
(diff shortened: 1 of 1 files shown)


nograx:
Fixed the issues on GitHub actions. I will bump the version number if first test from other users are successful.

mcm1957:
@nograx 

  ```
  "encryptedNative": [
    "password",
    "data2"
  ],
  "protectedNative": [
    "password",
    "data2"
  ],
  ```

- [ ] logging state changes at level info 
  Please check wether logging state changes at level info is required. I would suggest level debug to be sufficient unless you are sure, that only such states are subscribe which are very seldom changed.

  Some characters are not allowed to be used as part of an object id. If an object id is not hardcoded or even depending on user input you should ensure that such an object id does not contain an invalid character.  Iobroker provides the constant ```adapter.FORBIDDEN_CHARS```

  Your state Ids are composed of productKey and deviceKey. I did not see a verification that thise values do not contain illegal characters. Even if the current api ensures that all charactes are valid, this shoudl be checked at code as either the api could cahnge or due to some problem random data could be read.

- [ ] folder creation seems to be missing

  You use state ids like '    productKey + "." + deviceKey + "." + "lastUpdate", '

  So you must explicitly create folders (type: folder, device, channel) for productKey and deviceKey explicitly. Pleaase note that ioBroker standards require [device.[channel.[folder...]]].state sequence.

- [ ] reevaluate state roles and read/write flags

  Several states seem to have write flag set to true althoug the are not to be written by user for sure. Please set write to false except the state is to be written by user and handled at onStateChange.

  outputLimit must by of type level as it is an input value

- [ ] axios calls should have timeouts

  All axios calls should have a timeout specified as default for timeout according to axios docu is 0 = no timeout. ZThis could lead to stalled processing

- [ ] time / itervalls must be cleard at unload

  All timers and intervalltimers mut be cleared during execution aof onUnload. This is required as otherwise timers could lead to carshes when using compact mode where multiple adapters shre one process environment.
  
  
McM1957

mcm1957:
@nograx 

Do you need any help? 
Need some topics more explanation?

nograx:
Hi,
there where some points... not enough time, the adapter is running fine in my environment and it seems there is a small demand for this adapter... 

I've tried to fix all the issues mentioned above. 
Thanks,
Peter

mcm1957:
RE-CHECK!

mcm1957:
@nograx 
Thanks for review / changes,

Did you miss

- [ ] axios calls should have timeouts

  All axios calls should have a timeout specified as default for timeout according to axios docu is 0 = no timeout. ZThis could lead to stalled processing

or do **I** miss something. axios does not have a default timeout, so in case of a non or slow answering server this culd lead to hanging adapters. Especially combined with intall this is not safe. axios timeoput should be shorter than interval.

Please review or comment

nograx:
Hi,

you're right. Forgot to add the timeout. Fixed it in release 1.0.4.

Cheers, 
Peter

mcm1957:

PR #2838, opened 2023-10-29
Bumb TibberLink stable to 1.4.1

hombach:


hombach:
RE-CHECK!"

hombach:
RE-CHECK!

hombach:
RE-CHECK!

hombach:
RE-CHECK!

hombach:
must be checked label seems to be false positive because of verification of newer version 1.4.2 - the there found issue has been fixed in between, so also this error isn't found anymore

mcm1957:
1.4.1 created 25.10.2023
91 User
OK

mcm1957:
> must be checked label seems to be false positive because of verification of newer version 1.4.2 - the there found issue has been fixed in between, so also this error isn't found anymore

Those warnings / Errors are can be avoided if you do not change version numbers manually asynchronous to releasing a new version but change them using the releasescript during releasing the version.

But in anyway all errors / warnings will be manually reviewed and if the are casused by some changes for future versions they will be ignored. So no need to fic them explicitly.

hombach:
Brand new?? There are 900 people using the actual stable version 1.0.0 ?!?!?!

mcm1957:
> Brand new?? There are 900 people using the actual stable version 1.0.0 ?!?!?!

Please read the text above, If you have suggestions to make it more clear, they are welcome.
1.0.0 is irrelevant. 1.4.1 is requested to be listed at stable. And this release is only 4 days old.

PR #2862, opened 2023-11-01
Bump Kostal-Piko-BA stable to 3.0.7

hombach:


mcm1957:
3.0.7 created 01.10.2023
OK

PR #2866, opened 2023-11-03
Add reolink to latest

oelison:
Please add my adapter ioBroker.reolink to latest.

### DIFF
--- a/sources-dist.json
+++ b/sources-dist.json
@@ -1957,6 +1957,11 @@
     "icon": "https://raw.githubusercontent.com/TA2k/ioBroker.renault/master/admin/renault.png",
     "type": "vehicle"
   },
+  "reolink": {
+    "meta": "https://raw.githubusercontent.com/aendue/ioBroker.reolink/main/io-package.json",
+    "icon": "https://raw.githubusercontent.com/aendue/ioBroker.reolink/main/admin/reolink_logo.png",
+    "type": "alarm"
+  },
   "residents": {
     "meta": "https://raw.githubusercontent.com/jpawlowski/ioBroker.residents/main/io-package.json",
     "icon": "https://raw.githubusercontent.com/jpawlowski/ioBroker.residents/main/admin/residents.svg",


oelison:
Title is missleading, but simply add reolink to ioBroker adapters. I have just autogenerated from ioBroker Developer Portal.

mcm1957:
Don't know whats misleading. PR adds reolink to sources-dist.jsosn which definied which adapters are available at latest.
So everything seem OK.

oelison:
Ah yes, not stable, just latest...

mcm1957:

I would like to give some feedback based on my personal oppinion. This is NOT an offical review and @Apollon77 might have several additional suggestions or even a different oppinion to one or the other statement. Please feel free to contact him (or wait for a response from him) if you cannot follow my suggestions or want to discuss some special aspects.

- [ ] package.json should contain keywords

  Please add some matching keywords to package.json.
  'template' seems to be unchanged from creaton, realonk etc. would be useful.

- [ ] package.json should specifiy required nodejs version

  Please consioder adding en engines clause to package.json. Especially if you use adapter-core 3.x.x or newer nodegs 16 or newer must be reuested:
  ```
    "engines": {
    "node": ">= 16"
  },
  ```
  
- [ ] please verify that all dependencies are up to date

  Some dependencies are outdated. Please update to current releases if no special reason to keep an older version exists.
  adapter-core 2.6.0 and axios 0.26.2 are outdated for sure. Same for i.e. adapter-dev, ioborker/testing etc.

  ```
  "encryptedNative": [
    "password",
    "data2"
  ],
  "protectedNative": [
    "password",
    "data2"
  ],
  ```

- [ ] review translations for ids

  Please review translations encoded into io-package.json. I do not think that any camera will provide a 'gesichtsstütze' :-)
  As an alternative solution you could implement name attribute translation within code using i18n and weblate.
 
- [ ] reevaluate state roles

  Special remarks:
  - write:true must be set only, if state is to be written by external sources (user scripts, other adapters, ...) Those states must not have a role value.* but most likely level.*
  - all states with write:true must be handled at onStateChange
  - states *.support sound like an indocator, so check if write:true is needed there. Harddiscsizte does not make much sense to write too - or else?
  - please check state types, some type sound like a boolean indoicator (i.e. sensor.dog_cat.support). Sound like info if this feature is supported but not as a number of cats supported. If the state is logically a boolean, it should be declared and handled as such - and eventually processed / converted in code accordingly.
  - role:motion does not exist. Maybe sensor.motion is what you look for.
  
  Please reevaluate all roles used.
 
  Tests for node 16 is mandatory unless you require node 18 and higher (at package.json).
  Tests for node 18 is mandatory as this is the recommended node version and node 16 will be eol and deprected Q3/2023.
  
  So the recommended testmatrix is [16.x, 18.x, 20.x] or [18.x, 20.x] depending on engines requirments setting at package.json 

  As node 16 is end-of-life now, it's ok to remove node 16 from testing. In this case please require node 18 as minimum node version at pacakge.json.

- [ ] github dependabot support seems to be missing
  
  It looks like you did not yet setup dependabot support; at least .github/dependaboit.yml is missing.
  
  Please consider setting up dependabot support (and to ease life the workflow auto-merge.yml). Dependabot will ensure that your adapter uses the most recent versions of external packages and avoids problem due to outdted base libraries. Dependabit provides PRs at you repository to update the dependencies. Depending on you setting, especially on auto-merge.yml configuration no or only minor revisions will be outmatically merged into you main branch. All major revisions msiut (ans should) be reviewed by you. 
  
  
McM1957

oelison:
Thank you for the good review. All is easily solvable. I will came back when it is done.

mcm1957:
@oelison 
Still work-in-progress from your side or ready for re-review?

oelison:
Re-review could be done. With [PR 36](https://github.com/aendue/ioBroker.reolink/pull/36) all points from this ticket are solved.

mcm1957:
RE-CHECK!

mcm1957:
Some more feedback:

- [ ] incorrect role
  https://github.com/aendue/ioBroker.reolink/blob/42863bb6bd74194319503b855f95198849577cf9/main.js#L92C38-L92C55
  Role value must have write:false. Most likely this state is a switch or a button (would require red:false then)

- [ ] roles seems to be improveable:
  switchLed seesm to be a switch
  ledBrightness seems to be a level
  several values states seem to be indicator (due to boolen value)

Improving roles could help vis adapters to use better disply widgest. 

**Please fix at least to one definitly incorrect role with the next release.** I will no longer delay addition to latest anyway.

mcm1957:

**PLEASE consider fixing the notes I left here:**
https://github.com/ioBroker/ioBroker.repositories/pull/2866#issuecomment-1837124546

oelison:
@mcm1957 The forum thread allready exist here: [neuer-adapter-reolink-kamera](https://forum.iobroker.net/topic/56116/neuer-adapter-reolink-kamera)

PR #2886, opened 2023-11-08
bump TibberLink stable to 1.4.2

hombach:
Would be fine to get this earlier to stable, because of included fix for axios vulnerability


mcm1957:
1.4.2 created 03.11.2023
OK

PR #2890, opened 2023-11-10
Bump fullcalender to 2.2.6

mcm1957:


mcm1957:
2.2.6 created 27.07.2023
246 user
OK

PR #2891, opened 2023-11-10
Bump hm-rega to 4.0.0

mcm1957:


mcm1957:
4.0.0 created 04.09.2023
1500 user
OK

PR #2893, opened 2023-11-10
Bump mqtt to 5.1.0

mcm1957:


mcm1957:
5.1.0 created 11.10.2023
1900 user
OK

PR #2894, opened 2023-11-10
Bump openweathermap to 0.4.5

mcm1957:


mcm1957:
0.4.5 created 09.10.2023
705 user
OK

PR #2897, opened 2023-11-10
Bump web to 6.1.10

mcm1957:


mcm1957:
6.1.10 created 16.10.2023
3400 User
OK

PR #2907, opened 2023-11-11
Update knx to 2.0.23

chefkoch009:
set knx adapter 2.0.23 to stable


mcm1957:
2.0.23 created 11.10.2023
200 user
OK

PR #2917, opened 2023-11-11
Add samsung_tyzen to latest

mcm1957:


mcm1957:
This is a historical adapter.

PR #2918, opened 2023-11-11
Add Bosch Indego

TA2k:


TA2k:
RE-CHECK!

mcm1957:
RE-CHECK!

mcm1957:
replaced by PR #2919 as this PR initially had incorrect urls specified (master instead of main) and this will block automatic checks.
@TA2k please subscribe to #2919 to get notofications,

PR #2927, opened 2023-11-13
Update js-controller to 5.0.16 at stable

foxriver76:


mcm1957:
5.0.16 created 31.10.2023
OK

PR #2935, opened 2023-11-13
Bump Kostal-Piko-BA stable to 3.0.8

hombach:


mcm1957:
3.0.8 created 29.10.2023
17 user
OK

PR #2945, opened 2023-11-16
Update bidirectional-counter to 2.1.3

BenAhrdt:
Please update my adapter ioBroker.bidirectional-counter to version 2.1.3.


mcm1957:
2.1.3 create 14.11.2023

BenAhrdt:
I just added some debugloggings for Check an issue (at a User Installation)

mcm1957:
> I just added some debugloggings for Check an issue (at an User Installation)

Why do you think this requires fast lane passing?
User / Tester cooperating with you could easily install from beta repo. Und as far as I read your comment, the new release does NOT contain a fix to solve a high priority problem. So I think its ok to let the release be tested some days at latest repository.

BenAhrdt:
Ok. I just think Push this Version in stable, because Therme Is Nothilfen changed (just the logging) After the Last Release 2 days ago

PR #2964, opened 2023-11-19
bumb TibberLink stable to 1.5.0

hombach:


mcm1957:
1.5.0 created 14.11.2023
1.5.0 120 User

@hombach 
Is there any reason to NOT update directly to 1.5.0 which is already 6 days at latest and has 120 Users?

hombach:
Well, yes, thought the rule is minimum one week in latest?
If 1.5.0 is ok,   I would file a new request?

PR #2977, opened 2023-11-23
Updated device-watcher to v2.9.6

ciddi89:
Updated adapter device watcher from v2.9.4 to v2.9.6

thanks


mcm1957:
2.9.6 created 12.112023

PR #2983, opened 2023-11-24
Bump harmony to 1.3.0

mcm1957:


mcm1957:
08.11.2023,
400 user
OK

PR #3006, opened 2023-11-27
Bump TibberLink stable to 1.6.1

hombach:
1.6.1 isn't old enough, but a functional regression has been discovered in 1.5.0 - so I want to replace it earlier. Usage count of 1.6.0 and 1.6.1 could be added, cause there are no functional changes.


mcm1957:
1.6.1 created 26.11.2023
63 user

PR #3020, opened 2023-12-01
pls update icontwo to 0.98.0

skvarel:


mcm1957:
0.98.0 created 26.11.2023
OK

mcm1957:
@skvarel 

- [ ] Your adapter requests a restart of vis. As vis-2 is available in the meantime two, please extend the rsatrt clause to include vis and vis-2

- [ ] Yout adapter defined 'vis* as dependency. This causes problems when used with vis-2. This dependency is not needed at all, as the adapter only prrovides icons and can be installed at any time. Icons canbe used by any software / adapetr and are (most likely) not restricted to work with vis. So please remove vis from dependencies.

NOTE:
As dependencies are updated from grithub NO new release is required. Just update io.package.json please and 
**please drop a comment here when done**.

PR #3030, opened 2023-12-05
update lifx 1.0.4

foxthefox:


mcm1957:
1.0.4 created 03.11.2023
OK

PR #3033, opened 2023-12-05
add srm to latest

stephan1827:
Add adapter srm to latest

### DIFF
--- a/sources-dist.json
+++ b/sources-dist.json
@@ -218,7 +218,8 @@
     "meta": "https://raw.githubusercontent.com/iobroker-community-adapters/ioBroker.boschindego/main/io-package.json",
     "icon": "https://raw.githubusercontent.com/iobroker-community-adapters/ioBroker.boschindego/main/admin/boschindego.png",
     "type": "garden"
-  },  "bosesoundtouch": {
+  },
+  "bosesoundtouch": {
     "meta": "https://raw.githubusercontent.com/iobroker-community-adapters/ioBroker.bosesoundtouch/master/io-package.json",
     "icon": "https://raw.githubusercontent.com/iobroker-community-adapters/ioBroker.bosesoundtouch/master/admin/bosesoundtouch.png",
     "type": "multimedia"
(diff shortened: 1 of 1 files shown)


mcm1957:
Pr ok please ignore other adapters

ticaki:
As already explained in Telegram, a todo list with key points for the purpose of documentation.

- [ ] use Adapter.setTimeout()
- [ ] use FORBIDDEN_CHARS for not predefined data points
- [ ] only use roles from https://github.com/ioBroker/ioBroker.docs/blob/master/docs/en/dev/stateroles.md
- [ ] add xs:12 to jsonConfig.json for cellphones

stephan1827:
RE-CHECK!

mcm1957:
@stephan1827 

  ```
  function name2id(pName) {
  }
  ```
  You handle FORBIDDEN_CHARS during creation of mesh states but noit at setting the state values - if I'm correct

  https://github.com/stephan1827/ioBroker.srm/blob/b1b514e8f4d608c68e6814a22b31653dccadbec7/main.js#L235
  https://github.com/stephan1827/ioBroker.srm/blob/b1b514e8f4d608c68e6814a22b31653dccadbec7/main.js#L239

  You already use this.setTimeout. At least at one place you use clearTimeout, please adapt to this.clearTimeout. Additionally set/clearIntervall should be replaced by this.set/clearInterval.

  see https://github.com/stephan1827/ioBroker.srm/blob/b1b514e8f4d608c68e6814a22b31653dccadbec7/main.js#L168
  https://github.com/stephan1827/ioBroker.srm/blob/b1b514e8f4d608c68e6814a22b31653dccadbec7/main.js#L184
  https://github.com/stephan1827/ioBroker.srm/blob/b1b514e8f4d608c68e6814a22b31653dccadbec7/main.js#L92
  https://github.com/stephan1827/ioBroker.srm/blob/b1b514e8f4d608c68e6814a22b31653dccadbec7/main.js#L188
  https://github.com/stephan1827/ioBroker.srm/blob/b1b514e8f4d608c68e6814a22b31653dccadbec7/main.js#L92
  likely more plöaces - please check

- [ ] code for scheduled adapters
  Your adapter is running as daemon (https://github.com/stephan1827/ioBroker.srm/blob/b1b514e8f4d608c68e6814a22b31653dccadbec7/io-package.json#L114) - which is OK. Why do you use some code dedicated to schedule mode?
  https://github.com/stephan1827/ioBroker.srm/blob/b1b514e8f4d608c68e6814a22b31653dccadbec7/main.js#L172

  
McM1957

stephan1827:
Fixed version is up

mcm1957:
@stephan1827 

Some more feedback (non blocking) - eveluate for next release

- [ ] double replece is better than none :-)
   https://github.com/stephan1827/ioBroker.srm/blob/db321d7050c23ffadc7d5a6f5062ec8c99973747/main.js#L225

-[ ] cpu saving
   I don'nt know how much js jit compiler does an optimize, bute replacing once and using a variable could evntually save some cpu cycles
  https://github.com/stephan1827/ioBroker.srm/blob/db321d7050c23ffadc7d5a6f5062ec8c99973747/main.js#L225

mcm1957:
@stephan1827 

stephan1827:
Thx for the detailed review and hints

PR #3038, opened 2023-12-06
Add sun2000-modbus to latest (replacement for #3007)

mcm1957:
@daolis
This is a replacement PR for #3007 which has been closed due to merge conflicts causing reasonable wrong positives

### DIFF
--- a/sources-dist.json
+++ b/sources-dist.json
@@ -2342,6 +2342,11 @@
     "icon": "https://raw.githubusercontent.com/waoler/iobroker.stockmarket/master/admin/stockmarket.png",
     "type": "misc-data"
   },
+  "sun2000-modbus": {
+    "meta": "https://raw.githubusercontent.com/daolis/ioBroker.sun2000-modbus/main/io-package.json",
+    "icon": "https://raw.githubusercontent.com/daolis/ioBroker.sun2000-modbus/main/admin/sun2000-modbus.png",
+    "type": "energy"
+  },
   "sureflap": {
     "meta": "https://raw.githubusercontent.com/Sickboy78/ioBroker.sureflap/master/io-package.json",
     "icon": "https://raw.githubusercontent.com/Sickboy78/ioBroker.sureflap/master/admin/sureflap.png",


mcm1957:
@daolis

- [ ] unusual usage of admin columns

  Usage of adminColumns is very unusual. It looks like you are trying to display configuration data. As space at objects tab is very limits adminCOlumns sholud be used only if they provide a big beneit for users. Headers must be as short as possible and not be longer than typical data.
  I suggest to remove adminColumns completly for this adapter. Otehrwise please specify why you think thea provide what benefit to users.

  setObject(Async) creates a new object whenever it is called. You should avoid using setObject unless you really want to create new objects. Creating a new object deletes any userConfig, i.e. configures history logging which is most likely not desired. Most likely either setObjectNotExists oder (even better in most cases) extendObject should be used. Extend Object ensures that attributes like role, read/write flag etc. are set as specified and hence extendObjet shul dbe preferred when called onece during startup of the adapter.

  see https://github.com/daolis/ioBroker.sun2000-modbus/blob/bef1ac1388cf1f0893abd48f820bd38a703d4912/src/lib/states.ts#L283
 
- [ ] reevaluate state roles

  model name should be info.name
  serial number should be info.serial
  etc. 

  writeable states make no sense as no processing is implemented, i.e.
  https://github.com/daolis/ioBroker.sun2000-modbus/blob/bef1ac1388cf1f0893abd48f820bd38a703d4912/io-package.json#L288

  Please review and check all roles.

  
McM1957

mcm1957:
Any news?
Do you need help?
reminder 31.12.2023

daolis:
Hi, sorry, for the late answer.
Wanted to do it in vacation, but now i'm sick.
Will do it when im fit again.

mcm1957:
> @mcm1957 Hi, sorry, for the late answer. Wanted to do it in vacation, but now i'm sick. Will do it when im fit again.

No problem at all. Thanks for the feedback.

And first all get healthy again soon.
My best wishes

mcm1957

mcm1957:
reminder 31.01.2024

daolis:
@mcm1957 I created a PR with the requested changes. Could you pls take a look.. Thanks
https://github.com/daolis/ioBroker.sun2000-modbus/pull/27

If all is ok, then i will merge it and create a new release...

mcm1957:
Changes look good at a quick scan.

**Please add a comment when you have merged**

daolis:
@mcm1957 Merged and created release 0.0.2

mcm1957:
RE-CHECK!

mcm1957:

daolis:
Thank you Martin for the support!

PR #3060, opened 2023-12-12
Add blauberg-ventilation to latest

Dipl-Inge:


intellicasa:
RE-CHECK!

intellicasa:
RE-CHECK!

mcm1957:
@intellicasa 
You adapter might solve the this adapter request
https://github.com/ioBroker/AdapterRequests/issues/286
As theres a statement that the vents noted are compatible / identical to blauberg.

Please take a look at that reqeust and comment / confirm or rejec this assumtion based on your knowlegse.

mcm1957:

- [ ] Readme.md is very short

  The README.md of this adapter is very minimalistic. Please add a short description about the functionality of the adapter,  add links the the manufacturer site (if applicable) and add a description of the configuration of the adapter (if it's not completly self explainig). If you provide an external documentation - which should be located at doc/xx directory - please add links to at least the english documentation.

  ```
  function name2id(pName) {
  }
  ```
  You are creating states based on entry done by user (vent.name). This entry is already filtered/converted. Please ensure that it contains only valid characters (see FORBIDDEN_CHARS). In addtion you should ensure that all names are unique and even unique after conversion. 

- [ ] use async and  await with asynchronous calls

  Most of your routines should be asynchronous as they call asynchronous routines themself.

  You should use await at asynchronous calls unless you really do not care when the call is processed.
  i.E. setObjectNOtExistsAsync without an await might cause random effects as the creation of the object might be completed until is is accessed next - or not 
  see https://github.com/intellicasa/ioBroker.blauberg-ventilation/blob/5ef6714b58892ef160bcb5792a2858cde9a9f45d/main.js#L196 as an example.

  But this topic should be reviewed forthe whole code.

- [ ] reevaluate state roles

  Please note that role value required write=false. Input states (write:true) could be of role level as an example.
  But maybe you bind better matching roles. If possible try to configure all states i.e. using a table. Correct roles ease usage for vis applications and typedetector to decide what kind of a device the adapter handles.

- [ ] ensure that all operations are completed within intervalls

  As users could configure a encreasing number of vents it could be hard to ensure that querying all vents is compete within one interval period. If querying would last longer than one interval period multiple scnas would start to run in parallel and fianlly overloading the system.

  So please evaluate to either add a sort of firewall (busy flag) to avoid multiple paralel executions of the scan loop or (in most cases better) replace intervall by a timeout started at every scan completion.
   

  
McM1957

mcm1957:
reminder 27.12.2023

mcm1957:
@intellicasa 
Happy New Year !

How shall we continue?
Do you need any help?

**Please let me/us know if you encounter any problem and which timefram you expect to fix the mentionend issues.**

mcm1957

reminder 15.01.2024

mcm1957:
How should we continue?
Do you need any further help?

**Please let me/us know (by adding a comment) when all issues noted earlier are processed by you and a new review seesm to be appropiate.**

**If there is no response until 28.2.2024 this PR might be closed.**

reminder 15.2.2024

mcm1957:
How should we continue?
Do you need any further help?

**Please let me/us know (by adding a comment) when all issues noted earlier are processed by you and a new review seesm to be appropiate.**

**If there is no response until 28.2.2024 this PR might be closed.**

reminder 22.2.2024

mcm1957:
How should we continue?
Do you need any further help?

**Please let me/us know (by adding a comment) when all issues noted earlier are processed by you and a new review seesm to be appropiate.**

**If there is no response until 28.2.2024 this PR will be closed.**

reminder 28.2.2024

mcm1957:
As there was no reaction to the lastest status polls this PR will be closed now.

Feel free to create a new PR at any time if you plan to further maintain / fix this adapter.

PR #3070, opened 2023-12-15
Update webui to 1.20.6

jogibear9988:


mcm1957:
1.20.6 created at 12.12.2023

mcm1957:
@jogibear9988 
Please ensure that 1.20.6 will be available at latest for at least 7 days so that it can be installed by latest users and tested.
This requires, that you avoid publishing new releases as this would abort the test phase and users will no longer install or keep this release active.

Creating .alpha.0 releases would be ok as those release should not be published at latest when using the standard workflow.

jogibear9988:
For me this workflow does not work, sometimes I fix critical bugs, and after that there are still new releases. When I create a pull here it is tested enough.

mcm1957:
> e still new releases. When I create a pull here it is te

mcm1957:
> For me this workflow does not work, sometimes I fix critical bugs, and after that there are still new releases. When I create a pull here it is tested enough.

A critical bug will be provided very fast - you just  need to tell:

```
IMPORTANT:
Of course, it is possible to release a new version immediately if it is a hotfix for a serious problem, i.e. some errors cause adapter crashes or incompatible api changes of external websites blocking normal usage. In this case, please indicate this fact as a comment and mention mcm1957 and eventually Apollon77 explicitly. Please describe the reason (i.e. by referencing an issue). Hotfixes should minimize the changes, even dependency updates should be avoided if not related to the fix. New functionality and major (breaking) updates are most likely never a hotfix.
```

But 1.20.6 CANNOT be a critical ptach release to 0.18.4 as the minor version encreased by 2 numbers and the code conatins much more than a singel (or multiple) critical fixes. So let it run at latest for some time to be tested. 

Theres no need to produce  new releases while one is beeing tested at latest. Unless you have to fix something which of course starts the test phase again.

It no problem to test new code directly at your development environmant or from  npm. You only should avoid to publich it to latest unless you want to become this release the next test candidate. When using standard workflow using .alpha.x releases will avoid publishing at latest repository.

mcm1957:
As 0.20.7 has been release I cancel this PR.

Please create a new PR when 0.20.7 (or a leter version) has been available at LATEST for at least 7 days to be installed by users.

0.20.6 contains at least one issue which is fixed with 0.20.7 So it is obviously beeter to wait for the fixed release to be tested and to release this to stable. In addition 0.20.6 cannot be tested enough any more as you created a new release which aborts test phase.

PR #3071, opened 2023-12-16
add tedee to latest

TA2k:


TA2k:
RE-CHECK!

mcm1957:
RE-CHECK!

mcm1957:
RE-CHECK!

mcm1957:

- [ ] Are you sure that adapter uses a cloud service?

  "connectionType": "cloud", is specified at https://github.com/TA2k/ioBroker.tedee/blob/add612d9fa0cefb7203c7c3ad26af8976d04a574/io-package.json#L77 but code looks like it is using local connection. 

  Please check / adapt if local access only.


  ```
  "encryptedNative": [
    "password",
    "data2"
  ],
  "protectedNative": [
    "password",
    "data2"
  ],
  ```

  You have configured a parameter named 'password' to be encrypted. But you config uses a paramater named 'token' which should be protected like a password. 
  Please adapt io-package.json.


  ```
  function name2id(pName) {
  }
  ```
  You retrieve the 'id' used to create the device object by queryiing the (external) device and using the 'name'attribute. (https://github.com/TA2k/ioBroker.tedee/blob/add612d9fa0cefb7203c7c3ad26af8976d04a574/main.js#L107). As the content this data might change in future, please filter illegal characters before using the data to create states.


- [ ] reevaluate state roles

  role 'boolean' is invalid, maybe you menat button (needs read:false) or switch, see https://github.com/TA2k/ioBroker.tedee/blob/1b328f631363df0666baf7ba7802382b29228cde/main.js#L158

- [ ] check updateInterval / config.interval
  
  updateInterval is initialized and clered, but I did not see any place where it was used (no setInterval).

  Perameter config.interval exists and it limits to be in range. But I did not see any place where it is used.


- [ ] github dependabot support seems to be missing
  
  It looks like you did not yet setup dependabot support; at least .github/dependaboit.yml is missing.
  
  Please consider setting up dependabot support (and to ease life the workflow auto-merge.yml). Dependabot will ensure that your adapter uses the most recent versions of external packages and avoids problem due to outdted base libraries. Dependabit provides PRs at you repository to update the dependencies. Depending on you setting, especially on auto-merge.yml configuration no or only minor revisions will be outmatically merged into you main branch. All major revisions msiut (ans should) be reviewed by you. (Note: That's a suggestion only and not blocking)
    
  
McM1957

TA2k:
1. I added the information which devices a supported. But I will not add link or photos of the manufacture to prevent any risk of an `Unterlassungsklage`. Because this project is not official connected to the manufactures and I'm not allowed to give this impression when I use manufacture links. I also have not the right to use official device photos. A good example was the Vodafone speedtest adapter
2. fixed
3. fixed
4. sensitive information are not logged
5. fixed
6. fixed
7. I will use the native implementation to have the control over the behavior
8. fixed
9. GitHub dependabot brings no benefit in my opinion. It only spams the commit history and notifications. Major updates should not updated automatically and minor updates are applied automatically an installation because of the ^ syntax in the package.json. Normally a developer checks all dependencies before publishing a new release on npm

mcm1957:
RE-CHECK!

mcm1957:
Thanks for the ultrafast reaction.
Sorry, the logging issue was my fault; I forgot to remive this checkpoint from feedback.

I completly understand that using a companys logo might cause serious trobled. I personally do not think that a plain links to an official website of a manufacturer could be criticial - I did not read about any such claim. But I can and will respect your decision.

Same for Dependabot - I already mentioned that it is no blocking feedback.

mcm1957:

PR #3074, opened 2023-12-17
Add webcal 1.3.0 to stable

dirkhe:
Adapter is working several month stable on my side and also some other user are satifised, see https://forum.iobroker.net/topic/62479/test-neuer-adapter-webcal
So we should offer them more people

### DIFF
--- a/sources-dist-stable.json
+++ b/sources-dist-stable.json
@@ -2905,6 +2905,12 @@
     "type": "general",
     "version": "6.1.10"
   },
+  "webcal": {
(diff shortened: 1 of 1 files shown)


mcm1957:
Adapter available at latest since 17.12.2023.
So lets wait some time while it is loaded / tested from latest.

mcm1957:
reminder 31.12.2023

mcm1957:
1.3.0 created 01.11.2023
21 User (70%)
OK

PR #3087, opened 2023-12-19
Update webui to 0.20.7

jogibear9988:


mcm1957:
Please ensure that 0.20.7 will be available at latest for at least 7 days so that it can be installed by latest users and tested.
This requires, that you avoid publishing new releases or setting latest tag at npm to a newer release as this would abort the test phase and users will no longer install or keep release 0.20.7 active.

Creating .alpha.0 releases would be ok as those release should not be published at latest when using the standard workflow.

This PR will be closed if 0.20.7 cannot reach the installtion an test goal du to removal from latest repository.

mcm1957:
0.20.7 created 15.12.2023

mcm1957:
0.20.7 created 15.12.2023 and available at stable until 22.12.2023
23 User (>10%)

OK

0.21.3  released to latest now

PR #3135, opened 2023-12-26
Bump telegram-menu to 1.5.1

MiRo1310:
Please bump telegram-menu to 1.5.1 in stable repo


mcm1957:
1.5.1 created 20.12.2023
app 50 User
OK

Please assure that a release is available at least 7 days at latest repository. So do NOT create a new release (except -alpha.x) until you request a stable addition as the new reqeust will cancel availability at stable. And of course creating a new release to fix a bug is always ok - in this case the new release will be next at stable anyway.

PR #3151, opened 2023-12-29
bump kostal-piko-ba stable to 3.0.9

hombach:


mcm1957:
3.0.9 created 23.12.2023
10 user
OK

PR #3155, opened 2023-12-31
Update icons-mfd-svg to 1.2.0

mcm1957:
according to Apollon icons adapter always ok


mcm1957:
1.2.0 created 07.12.2023
OK

PR #3163, opened 2024-01-02
Added vis-2 and some widgets to stable

GermanBluefox:


GermanBluefox:
RE-CHECK!

mcm1957:
vis-2 2.9.10 creates at 02.01.2024

mcm1957:
RE-CHECK!

PR #3218, opened 2024-01-18
Add elgato-key-light to latest

mcm1957:


mcm1957:
[E114] seems to be false positive. No normal admin UI but admin tab present

PR #3224, opened 2024-01-20
Add lorawan to latest

BenAhrdt:
Please add my adapter ioBroker.lorawan to latest.

### DIFF
--- a/sources-dist.json
+++ b/sources-dist.json
@@ -503,7 +503,7 @@
     "meta": "https://raw.githubusercontent.com/iobroker-community-adapters/ioBroker.elgato-key-light/main/io-package.json",
     "icon": "https://raw.githubusercontent.com/iobroker-community-adapters/ioBroker.elgato-key-light/main/admin/elgato-key-light.png",
     "type": "lighting"
-  },  
+  },
   "email": {
     "meta": "https://raw.githubusercontent.com/iobroker-community-adapters/ioBroker.email/master/io-package.json",
     "icon": "https://raw.githubusercontent.com/iobroker-community-adapters/ioBroker.email/master/admin/email.png",
@@ -1300,6 +1300,11 @@
     "icon": "https://raw.githubusercontent.com/foxriver76/ioBroker.loqed/main/admin/loqed.png",
     "type": "hardware"
   },
+  "lorawan": {
+    "meta": "https://raw.githubusercontent.com/BenAhrdt/ioBroker.lorawan/main/io-package.json",
+    "icon": "https://raw.githubusercontent.com/BenAhrdt/ioBroker.lorawan/main/admin/lorawan.png",
+    "type": "protocols"
+  },
   "lovelace": {
     "meta": "https://raw.githubusercontent.com/ioBroker/ioBroker.lovelace/master/io-package.json",
     "icon": "https://raw.githubusercontent.com/ioBroker/ioBroker.lovelace/master/admin/lovelace.png",
@@ -1557,11 +1562,6 @@
     "icon": "https://raw.githubusercontent.com/iobroker-community-adapters/ioBroker.myenergi/master/admin/myenergi.png",
     "type": "energy"
   },
-  "myuplink": {
-    "meta": "https://raw.githubusercontent.com/sebilm/ioBroker.myuplink/master/io-package.json",
-    "icon": "https://raw.githubusercontent.com/sebilm/ioBroker.myuplink/master/admin/myuplink.png",
-    "type": "climate-control"
-  },
   "myq": {
     "meta": "https://raw.githubusercontent.com/StrathCole/ioBroker.myq/master/io-package.json",
     "icon": "https://raw.githubusercontent.com/StrathCole/ioBroker.myq/master/admin/myq.png",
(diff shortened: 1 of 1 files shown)


mcm1957:
reminder 03.02.2024

BenAhrdt:
@mcm1957 muss ich was fixen?

mcm1957:

You will find the results of the review and evntually issues / suggestings as a communt to this PR. So please keep this PR watched. 

mcm1957

BenAhrdt:
Hi @mcm1957,
sorry wenn ich nachfrage, aber gibts schon einen Fortschritt?
Ich werde dauernd gefragt :-D

mcm1957:
Adapter is not yet revied. Hope I can process it next week .
Users can install it from github or npm anyway for testing

BenAhrdt:
Hi. @mcm1957 bist Du weiter gekommen?

mcm1957:
No, did not have time for review.
Hope I can gest some time tomorrow as next week I do not have time for review for almost sure.

BenAhrdt:
Thanks for the info,
the queries. That would be great.

mcm1957:

- [ ] Readme.md is very short

  The README.md of this adapter is very minimalistic. Please add a short description about the functionality of the adpter,  add links the the manufacturer site (if applicable) and add a description of the configuration of the adapter. If you provide an external documentation - which should be located at doc/xx directory - please add links to at least the english documentation.

  Someone not familar with lorowan need more information. Please add links to the homepages of the supported networks or some place to get infomation about them

  Please add a short description describing the purpose of the paramaters and where to get them from.

- [] small typo in german header

image

- [ ] Password field twice?

  Why do you require the password twice? As far as I can guess this is an parameter used to get access to an external system. Duplicate password entry is used for setting up passwords for own servers usually. But that does not seem to be the case here. 

- [ ] check admin requirements 
 
  Admin 5.x.x does not provide accordion for sure. Please correct dependency. Check admin releasenotes to get correct value.

  This line https://github.com/BenAhrdt/ioBroker.lorawan/blob/6e3d9bae58e5a828a6d5bc5358833b61c6b16eb7/io-package.json#L155 logs complete config including password. Please filter password (and eventually other sensitive infomration.). As an alternative remove logginng oc config completly as this can be retrieved via ui or dump of system object anyway (without protected data)

- [ ] evaluate subsciptions

  You are subscribing to multipl wildcarded states. This requires js-controller to do some parsing. Maybe ist would be better to subscribe to all states and do filtering locally as this can be processes within the adapter code and process. Evnetually discuss this at telegramm to get feedback from Appollon77

```adapter.FORBIDDEN_CHARS```, i.e. by using the following snippet:

  ```
  function name2id(pName) {
  }
  ```
  
  It loks like object Ids are generated from date transmitted frome xternal systems and / or entered by user at UI. Please filter all ids to remove invalid characters. Filtering at UI is insufficient as config could be changed by direct access to system objet and i.e. by import of a configuration.

  setObject(Async) creates a new object whenever it is called. You should avoid using setObject unless you really want to create new objects. Creating a new object deletes any userConfig, i.e. configures history logging which is most likely not desired. Most likely either setObjectNotExists oder (even better in most cases) extendObject should be used. Extend Object ensures that attributes like role, read/write flag etc. are set as specified and hence extendObjet shul dbe preferred when called onece during startup of the adapter.

  Why do you recreate the object every time? Why not use extendObject (or setObjectNotExists)?
 https://github.com/BenAhrdt/ioBroker.lorawan/blob/6e3d9bae58e5a828a6d5bc5358833b61c6b16eb7/lib/modules/messagehandler.js#L101

- [ ] reevaluate state roles

  At least here https://github.com/BenAhrdt/ioBroker.lorawan/blob/6e3d9bae58e5a828a6d5bc5358833b61c6b16eb7/lib/modules/messagehandler.js#L101 all objects are created with write:true. Please check whther write:true is really needed / OK. Do you subscribe to all of them? 

McM1957

BenAhrdt:
Hello @mcm1957  & @Apollon77 

i have solved all the points of the last comment.

Here my comments to the points:

 _Readme.md is very short_

=> Solved with an first documentation

 _Password field twice?_

=> Solved with just one password

 _check admin requirements_

=> Solved with Setting Admin Version in dependencies to 6.13.15

 _sensitive information should not be logged into logfile_

=> Solved: Log just the accordion parameter

 _evaluate subsciptions_

=> Solved: Subscribe ("*"); and filter in onStateChange

 _invalid characters should be filtered from object ids_

=> Solved: Use the funktion replace(adapter.FORBIDDEN_CHARS;"_");

 _avoid usage of setObject_

=> Solved: Use extendObject 

 _reevaluate state roles
Only the values specified here (https://github.com/ioBroker/ioBroker.docs/blob/master/docs/en/dev/stateroles.md) may be used as state roles. Do not invent new names as this might disturb the functionalyity of other adapters or vis._

=> Solved:I use only this roles

In addition the roles MUST match the read/write functionality.

=> Solved: eg. button is read = false;

_Please avoid using general roles (state, value) whnever a dedicated role exists._

Solved: i dont use this

_At least here https://github.com/BenAhrdt/ioBroker.lorawan/blob/6e3d9bae58e5a828a6d5bc5358833b61c6b16eb7/lib/modules/messagehandler.js#L101 all objects are created with write:true. Please check whther write:true is really needed / OK. Do you subscribe to all of them?_

=> Solved: if i dont use wildcards i subscribe all... (fun). These states are all in folder "control" an where all subscribed and send topics i case of state change

mcm1957:
RE-CHECK!

mcm1957:

PR #3235, opened 2024-01-25
Bump iobroker.apg-info to 0.1.5

HGlab01:


mcm1957:
0.1.5 created 20.01.2024
12 user (10%
OK

PR #3252, opened 2024-01-29
Update ble to 0.14.0

mcm1957:
https://github.com/AlCalzone/ioBroker.ble/issues/828


mcm1957:
0.14.0 created 11.12.2023
260 user
OK

PR #3260, opened 2024-01-29
Update stable Frigate 

TA2k:


mcm1957:
1.0.2 created 29.01.2024

TA2k:
Hotfix worthy because of significant reducing memory consumption.
High memory consumption can lead to problem in systems

mcm1957:
ok

PR #3283, opened 2024-02-04
bump admin stable to 6.13.16

foxriver76:
because of the `too much recursion` bug which lead to crashes with some firefox addons


mcm1957:
6.13.16 created 30.01.2024
OK

PR #3285, opened 2024-02-05
Update webui to 1.0.37

jogibear9988:


mcm1957:
1.0.27created 31.01.2024,
OK

PR #3310, opened 2024-02-14
Added SolarEdge 1.2.2 to stable

GermanBluefox:
The SolarEdge adapter has been added to the stable sources list in the sources-dist-stable.json file. Relevant details such as meta data, icon link, type and version have been included.


mcm1957:
RE-CHECK!

PR #3313, opened 2024-02-15
Update ems-esp to 2.8.0

tp1de:
Please update my adapter ioBroker.ems-esp to version 2.8.0.


mcm1957:
@tp1de
Just as info, non blocking

https://github.com/tp1de/ioBroker.ems-esp/blob/main/index_m.html
seems to be misplaced at adapetr troot directory. Please check.

tp1de:
Even without node 16 within release script I get this warnings .... no idea where they come from

mcm1957:
> Even without node 16 within release script I get this warnings .... no idea where they come from

Ignore it for now.
I'll try to check whats causing the messages.
Thanks for reporting

PR #3338, opened 2024-02-20
Add smoothed to latest

BenAhrdt:
Please add my adapter ioBroker.smoothed to latest.

### DIFF
--- a/sources-dist.json
+++ b/sources-dist.json
@@ -2242,6 +2242,11 @@
     "icon": "https://raw.githubusercontent.com/TA2k/ioBroker.smartthings/master/admin/smartthings.png",
     "type": "household"
   },
+  "smoothed": {
+    "meta": "https://raw.githubusercontent.com/BenAhrdt/ioBroker.smoothed/main/io-package.json",
+    "icon": "https://raw.githubusercontent.com/BenAhrdt/ioBroker.smoothed/main/admin/smoothed.png",
+    "type": "logic"
+  },
   "snips": {
     "meta": "https://raw.githubusercontent.com/unltdnetworx/ioBroker.snips/master/io-package.json",
     "icon": "https://raw.githubusercontent.com/unltdnetworx/ioBroker.snips/master/admin/snips.png",


mcm1957:
reminder 27.2.2024

mcm1957:
RE-CHECK!

mcm1957:
RE-CHECK!

BenAhrdt:
? is there something to do?

mcm1957:
And sorry for the long time to wait for a review.

- [ ] typo in README.md ?

   ```smoothes selected signas according to a selected function```
   I guess signas should read as signa**l**s

- [ ] Readme.md is very short

  The README.md of this adapter is very minimalistic. If you have timme please add some more description an i.e. an example. I think (from teelgram discussions) this adapter could provide filtering using high / low pass but for sure m uch more functionality. This should be described at readme so users can decide to test the adapter.

```
function name2id(pName) {
}
```
  At least channelname is extracted from user config (https://github.com/BenAhrdt/ioBroker.smoothed/blob/0aa65bc897d143ea7a4b03d20ee94780be0b1c97/lib/modules/statehandling.js#L124). This must be filtered. Did not check other sources (like smotthed value). Please ensure that all ids retrieved from external sources (like user input) are filtered for illegal characters.

McM1957

mcm1957:
reminder 12.3.2024

BenAhrdt:
I will have a Look for your comments this week.
First of all i have to say that the Adapter Loks at values to Smooth them. So it is possible the values came out from orher adapters (EVER) with ack = true.
So the ack flag dont matters on this Adapter, because the adapter just work with the States and dont Write them.

mcm1957:
> So it is possible the values came out from orher adapters (EVER) with ack = true. So the ack flag dont matters on this Adapter, because the adapter just work with the States and dont Write them.

If you are reading values output by other adapters they should be acked by the writing adapter, see last line of the remark:
 
> When processing foreign states which are output of another adapter processing must only occure if the ack flag is true.

Do you really want to process unacknowledged values of other adapters ?

BenAhrdt:
Yes. Because in the past There were many questions about it with States from scripts which are not acked.
The Adapter Can work With it, cause of the Timestampe.

BenAhrdt:
Here my comments to the changes.
(All chanes are done and work)

 typo in README.md ?

smoothes selected signas according to a selected function
I guess signas should read as signals

 Readme.md is very short

The README.md of this adapter is very minimalistic. If you have timme please add some more description an i.e. an example. I think (from teelgram discussions) this adapter could provide filtering using high / low pass but for sure m uch more functionality. This should be described at readme so users can decide to test the adapter.


=> `Readme changes`

Some characters are not allowed to be used as part of an object id. If an object id is not hardcoded or even depending on user input you should ensure that such an object id does not contain an invalid character. Iobroker provides the constant adapter.FORBIDDEN_CHARS, i.e. by using the following snippet:

function name2id(pName) {
}

=> `Use replace(adapter.FORBIDDEN_CHARS, '_') and in case there was a forbidden char,
the adapter give a warning and dont create the state`

At least channelname is extracted from user config (https://github.com/BenAhrdt/ioBroker.smoothed/blob/0aa65bc897d143ea7a4b03d20ee94780be0b1c97/lib/modules/statehandling.js#L124). This must be filtered. Did not check other sources (like smotthed value). Please ensure that all ids retrieved from external sources (like user input) are filtered for illegal characters.

=> `see point before... checking the name and id. if the id is nadditionally not a number,
adapter will log a warning and dont create sate`

 onStateChange handler ignores ack flag

When processing foreign states which are output of another adapter processing must only occure if the ack flag is true.

McM1957

Please add a comment when you have reviewed and fixed the suggestionsor at least commented the suggestions and you think the adapter is ready for a re-review!

=> `This pointis discussed. the Adater smoothes acknowledged and not acknoledged values (in case of values out of some scripts the useres are asking me to smooth this not acknowledged values.
This is possible, because the Adapter dont write this values (ids), just read and safe timestamp.
If it will be writen from other sources with acknowledge = false and maybe 1ms later with false, there is no problem
caused of the timestamp it will be calculatet correctly`

mcm1957:
RE-CHECK!

PR #3396, opened 2024-03-11
Add waterkotte-easycon to latest

theknut:


mcm1957:

- [ ] onStateChange handler ignores ack flag (NOT BLOCKING - information for future use only)

  Hardcoding text inside jsonConfig blocks usage of weblate transaltion system. 

  NOTE: This is a suggestion only and nozt blocking.


  I suggest to change the paramater at ui to minutes or seconds. Users would not expect a ms parameter here.
 
  And in any case limit the interval ta UI and inside code (!) to a reasonable value. Polling too fast could result in starting a new polling cycle before the previous cycle has finished. This could result in a crash of the complet ioBroker installation due to resource problems. Please ensure that pollingIntervall cannot by smaller than the longest prosessing time of a poll loop or replace setInterval by a setTimeout started at the end of a poll. This would be the most stable solution anyway.


McM1957

mcm1957:
@theknut

How shall we continue?
Do you need any help?
Did you notice the last comment and the provided PR?

Please check andfix or at least comment the issues listed earlier 

reminder 8.4.2024

theknut:
Sorry, I was sick and didn't find the time to review it. I'm back now and will do it in the coming days.

theknut:
All tasks have been resolved as per https://github.com/theknut/ioBroker.waterkotte-easycon/commit/4932be865f0715a2c59f5c87a1ac6bfd4185ca2e

mcm1957:
RE-CHECK!

mcm1957:
@theknut 

Some more feedback - sorry

- [ ] onUnload registration missing

	```this.on("unload", this.onUnload.bind(this));```

  got lost at some time. So the onUnload handler will not be called. Please (re)add. See template for details: https://github.com/ioBroker/ioBroker.example/blob/55c3807672e11b281be472120043e8f2aa656bb9/TypeScript/src/main.ts#L23

Everything else looks good. Thanks for your effort.

reminder 19.4.2024

theknut:
- [x] bound onUnload via https://github.com/theknut/ioBroker.waterkotte-easycon/commit/acb311ebf1f78f08ffbb8ac2697a5de67f749c4e
- [x] updated `@iobroker/adapter-core` to `3.06` via https://github.com/theknut/ioBroker.waterkotte-easycon/commit/d9a0e2d43d51e0da0854ab2262e6f8428e486d80

mcm1957:
RE-CHECK!

mcm1957:

PR #3399, opened 2024-03-12
Update deconz to 1.4.0

Jey-Cee:
Please update my adapter ioBroker.deconz to version 1.4.0.


mcm1957:
Please consider using / adding standard test-and-release workflow with next / new releases. You can of course use your own tests too. This allows faster processing of checks.

As standard test-and.release workflow seems to missing currently, this test must be done manually which might cause some delay for release.

Jey-Cee:
Do you mean this one: https://github.com/Jey-Cee/ioBroker.deconz/blob/master/.github/workflows/test_and_release.yml
That is the standard but not up to date and with a small change that chooses the nodejs versions for testing from ioBroker installation.

mcm1957:
Yes I refer to this.
But the content is not standard nor is the name.

PR #3418, opened 2024-03-22
Update zendure-solarflow to 1.2.3 (or 1.2.6?)

nograx:
Please update my adapter ioBroker.zendure-solarflow to version 1.2.3.

### DIFF
--- a/sources-dist-stable.json
+++ b/sources-dist-stable.json
@@ -3192,7 +3192,7 @@
     "meta": "https://raw.githubusercontent.com/nograx/ioBroker.zendure-solarflow/main/io-package.json",
     "icon": "https://raw.githubusercontent.com/nograx/ioBroker.zendure-solarflow/main/admin/zendure-solarflow.png",
     "type": "energy",
-    "version": "1.0.7"
+    "version": "1.2.3"
   },
   "zigbee": {
     "meta": "https://raw.githubusercontent.com/ioBroker/ioBroker.zigbee/master/io-package.json",


mcm1957:
**I would suggest to wait a few days until 1.2.6 is tested at latest.**

1.2.3 was available at katest only 3 days, has been replaced with follow up releases very fast and never reached the app 7 days availability at latest.

In addiont please respect versionen in future. new functionality (Add states for pass) should result in a minor version increase. Its ok to leave it for now - but any  (future) breaking change requires a major version increase.

nograx:
1.2.3 was a good version. Currently I'm rapidly developing and releasing new features as the adapter lacks behind of the official Zendure app. 1.2.4-1.2.6 had a few small bugs I don't want to release to the stable channel. The next possible version is 1.2.7 if no other bugs occur.

mcm1957:
Ok, then lets wait for 1.2.7.

If you consider a release should go to stable, please avoid creating new latest releases as the goal of waiting app 1 week minimum is not to make the code older but to keep it available at latest and used for test installs. So if you i.e. create 1.2.4 3 days after 1.2.3 1.2.3 will only be availbale 3 days to be installed.

If you want to continue development I suggest to use -alpha.x releasea as these release can be installed via npm but are not listed at latest repository (at least if using the standard workflow)

mcm1957:
And anyway
As you added new features - as far as I read - the new release should be 1.3.x

PR #3433, opened 2024-03-28
Update net-tools to 1.0.11

Jey-Cee:
Hotfix for high cpu load.

Please update my adapter ioBroker.net-tools to version 1.0.11.


Jey-Cee:
Thank you

mcm1957:
Please check why your adapter still passes node 16 tests at GH. Looks like it is not using the standatd tests for integration.
As node 16 has been removed from current js-controller (pre 16) node 16 should fail since this week.

And please consider removing node 16 from testing and require node 18  with next normal (= non patch, non hot-fix) release.

PR #3455, opened 2024-04-01
Bump apg-info to 0.1.6

HGlab01:


mcm1957:
RE-CHECK!

mcm1957:
@HGlab01 

Every Adapter at STABLE reposittory MUST pass tests for Node 18 and 20 at all supported platforms (Linux, Windows, Mac).

Why did you remove Windows tests?
Please readd if there is no technical reason which prohibits the adapter to be used at windows.

mcm1957:
reminder 5.4.2024

HGlab01:
> Why did you remove Windows tests? Please readd if there is no technical reason which prohibits the adapter to be used at windows.

Because
1. no OS-based feature of nodejs is used
2. github windows instances are super slow in setup

But if it is a rule I will stick to it. Windows included again.

Apollon77:
@HGlab01 Windows is important because when native libs or dependencies are used or such there always might be build issues that arise, so it is better to check it and know. Thank you for your understanding

PR #3511, opened 2024-04-15
Add new adapter to manage Gree and C&H air conditioners.

XHunter74:


mcm1957:
This PR changes stable AND latest repository.
Besides it contains an unneeded number of intermediate commits.

Addion of gree-hvac will be processed with this PR:
https://github.com/ioBroker/ioBroker.repositories/pull/3512

If you really want to update release of apcups/ please create a dedicated PR for this. I would suggest to use www.iobroker.dev for this thus reducing the risk of unneded commits.

PR #3530, opened 2024-04-22
Update energiefluss-erweitert to version 0.4.1- Hotfix

SKB-CGN:
Hotfix: After uprading the Adapter, it could happen, that the first datasource was not updated properly


mcm1957:
merged
will trigger an extra update after checks have completed so stable is updates asap

SKB-CGN:
Thanks a lot!

PR #3537, opened 2024-04-25
Add opcua 1.0.3 to stable

mcm1957:
https://github.com/ioBroker/ioBroker.opcua/issues/31


mcm1957:
standard tests are missing

PR #3538, opened 2024-04-25
update innoxel to 0.4.1

matthsc:
Update innoxel adapter to 0.4.1
- fixes for bug reported by a user (private chat)
- drop support for node 16
- dependency updates


matthsc:
RE-CHECK!

mcm1957:
Please update adapter-core and @iobroker/testing to fix timeout at windows tests!

PR #3603, opened 2024-05-11
Update webui to 1.6.5

jogibear9988:


mcm1957:
No fix required.
Your adapter fails std tests due to reasons discussedxearluer.

mcm1957:
RE-CHECK!

PR #3615, opened 2024-05-14
Add google-home-nest to latest

danieldriessen:


mcm1957:

reminder 22.05.2024

mcm1957:
RE-CHECK!

mcm1957:
@danieldriessen 

- [ ] io-package.json should require up to date js-controller and admin version

  ```
        "dependencies": [
            {
                "js-controller": ">=5.0.19"
            }
        ],
        "globalDependencies": [
            {
                "admin": ">=6.13.16"
            }
        ]
  ```

  ```
  "encryptedNative": [
    "password",
    "data2"
  ],
  "protectedNative": [
    "password",
    "data2"
  ],
  ```
 
  At least 'client-secret' seems to be worth protecting, if other senstive information (cloud access data) exists this should be protected too. 

  relates to client secret and maybe other security information

- [ ] sensitive information should not be  stored at states

  You seem to store the client secret at a stae? Why do you do this? The secret will be accessible suing this.config whenever needed. And I do not expect that the client secret should be readable by anyone using the system neither that it should be changed by other adapters. State names and data are more or less public - to any securiyt relevant data should neither be part of an id or be stored within a state as other Adapters, any user of admin and depending on vis any vis user could access them.

  Do you really need the access tokens to be preserved when restarting the adapter? Otehrwise they should not be store at states too. 

```
function name2id(pName) {
}
```

  Looks like some parts of the object / state id are retrieved from external sources, i.e. the 'deviceId' Those parts (or the complete id) must be filtered to not contain any forbidden character. 
  
- [ ] onStateChange handler only for deletion?

  Using an onStateHandler only to detect the delition of an state and to recreate it looks like an overhead as every stateChange must be processed by js-controller. and every subscription required the js-controller to process one more regex if any adapetr changes any state.

  So why not simply create the state whenever it shoudl be wirtten?
  Or simply create the state during startup only?
  I do not know any adapter which tries to fix deleted states during runtime - besides you approach will not work 100% anyway as there is no guarantee that the onStateChange is called after deleting a state befor any other access of your adapter occures.

  I would suggest to remove the onSateChange handler and to create states during startup only.

- [ ] reevaluate state roles

  You have store translation directly at jsonConfig.json. Consider moving them into i18n/<lang>translation.json files.This would allow the use of weblate connection and hence adapting translations be native speakers withou direct access to repo files.

  This is only a suggestion and NOT blocking.

- [ ] do not mix await and .then

  While javascript allows mixing 'action .then .catch' and await I would strongly suggest to clean this up. Either use await (togerther with try/catch if required) OR use .then .catch constructs. Mixing could leada to some surprising and hard to diagnose behaviour.

McM1957

danieldriessen:
@mcm1957 

Thank you for the review and for the reply.

As I mentioned in the adapter request repo., I am not a professional programmer. Therefore I knew that my code would probably be far away from perfect. On top of that, I sadly have zero javascript experience (besides the code I wrote for this adapter). So thanks again for putting me in the right direction. You probably scratched your head a lot while reviewing my code and I want to apologize for the inconvenience.

I will try to fix & improve all mentioned aspects of my code in the following days.

Thanks again.

mcm1957:
> As I mentioned in the adapter request repo., I am not a professional programmer. 

No problem. Most of devs here do this at their free time and for fun.

>On top of that, I sadly have zero javascript experience (besides the code I wrote for this adapter). 

You code look very well. There are only seveal iobroker specific adaptions to be done. You will make this for sure. Please feel free to ask whenever you have a question. Please feel free to subscribe out telegram developer channels. Invites available at http://www.iobroker.dev.

mcm1957:
Any news?
Do you need help?

Please feel free to ask whenever you have a question. Please feel free to subscribe out telegram developer channels. Invites available at http://www.iobroker.dev/.

reminder 21.6.2024

danieldriessen:
Hey.
Thanks for offering help if needed.
I have started to work on your suggestions. Unfortunately I am not very far. But right now this has more to do with me having a lot of work currently. I hope that I will be able to work on the changes in the next week.
Afterwards I will post a comment and comment the suggestions.
Thanks again!

mcm1957:
reminder 13.7.2024

mcm1957:
Any news?
Could you give any estimated timeframe for a re-review?

Please add a comment when you have reviewed and fixed the suggestionsor at least commented the suggestions and you think the adapter is ready for a re-review!

reminder 31.7.2024

danieldriessen:
Hello, and sorry for the late reply.
Right now I am having a lot of things to do for our upcoming vacation. Additionally I am having to work a lot (company-wise) until the vacation. I will be back at the end of August and completing my adjustments on this adapter is my number 1 priority for September.
I hope the delay doe not cause too much trouble. If preferred you can also cancel the merge request until I reach out again.
Again. Sorry for the delay.
And thank you very much!

mcm1957:
Thanks for Info and try to relax.

mcm1957:
RE-CHECK!

mcm1957:
RE-CHECK!

mcm1957:
RE-CHECK!

mcm1957:
RE-CHECK!

mcm1957:
Any news?
Any timeframe for a re- review?

Please feel free to ask whenever you have a question. Please feel free to subscribe out telegram developer channels. Invites available at http://www.iobroker.dev/.

If you do not have time to continue I would suggest to close here and you open a new PR whenever you feel that the adapter is ready for a check.

danieldriessen:
Hello.
Sadly, I don't really have any time left for this the next couple of days. But I am still planing to continue very soon. I will close this request for now and open a new one once I am ready.
Thanks for the patience!

PR #3656, opened 2024-05-23
Update air-q version in stable to 1.0.3

pr0crstntr:
Added some requested changes concerning error messages and more logging info.


mcm1957:
RE-CHECK!

PR #3707, opened 2024-06-04
EMERGENCY - Bumb TibberLink stable to 3.2.1

hombach:
We have to release by fast path, because Tibber has blocked a version of the used tibber-api dependency.
So no one could connect anymore. 

Adapter is now updated to brand new version of the dependency + added a unique portion of the endpoint - so hopefully Tibber won't block again. 

see also https://github.com/hombach/ioBroker.tibberlink/issues/412 for some user tests

### DIFF
--- a/sources-dist-stable.json
+++ b/sources-dist-stable.json
@@ -2188,7 +2188,7 @@
   },
   "rssfeed": {
     "meta": "https://raw.githubusercontent.com/oweitman/ioBroker.rssfeed/master/io-package.json",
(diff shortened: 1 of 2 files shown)


hombach:
OOOPS empty PR was my  vault .... didn't notice that you already updated  - THANK YOU!!

mcm1957:
No problem. Normally I do not bypass dev without prio notice. In this case statements from forum and the changelog looked like a pending emergency release - and wore than no functionality could not happen.

Anyway - please look at and ev comment at forum (link above) as at least one user tells that 3.2.1 does NOT work for him.

PR #3709, opened 2024-06-05
Bump iot stable

foxriver76:


mcm1957:
Testing should be updated to current standards and be performed at ALL platforms.
@foxriver76 
@GermanBluefox 
@Apollon77 

see https://github.com/ioBroker/ioBroker.iot/issues/774

PR #3737, opened 2024-06-15
Update sonoff to 3.1.0

mattreim:
[ioBroker.sonoff#379](https://github.com/ioBroker/ioBroker.sonoff/issues/379)


mcm1957:
non std testing OK

PR #3760, opened 2024-06-22
Update rssfeed to 2.9.7

mcm1957:


mcm1957:
wait for clearification of https://github.com/ioBroker/ioBroker.admin/issues/2513

If this is not an issue at admin / js-controller this adapter must either load all required files fo github or disable github installation by setting common.nogit to true

PR #3789, opened 2024-07-07
Update growatt to 3.2.4

PLCHome:
Please update my adapter ioBroker.growatt to version 3.2.4.

### DIFF
--- a/sources-dist-stable.json
+++ b/sources-dist-stable.json
@@ -846,7 +846,7 @@
     "meta": "https://raw.githubusercontent.com/PLCHome/ioBroker.growatt/master/io-package.json",
     "icon": "https://raw.githubusercontent.com/PLCHome/ioBroker.growatt/master/admin/growatt.png",
     "type": "energy",
-    "version": "3.2.3"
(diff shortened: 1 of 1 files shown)


PLCHome:
Since Wednesday, it has no longer been possible to connect to the cloud with the previous version. In other words, adapters up to version 3.3.3 no longer work. The vendor has changed the login to password encryption, which is clearly better.
See: https://forum.iobroker.net/topic/36474/test-adapter-growatt-v3-2-0/636
The problems mentioned so far in the forum and Github issues were only user errors, misconfigurations and NPM issues that were not caused by version 3.2.4.

I ask that the adapter be put into stable as soon as possible so that it works again without beta installation steps.

PLCHome:
Why doesn't the adapter checker show the mishaps? 
Is there another one?

image

mcm1957:
Yes
Checker at repo is newer than webbased checker. Webbased checker is a little bit complicated to update. Will be updaten when vacklog of improved checks is done.

Fell free to run check using npx to get the most aktual checker.

PR #3802, opened 2024-07-10
Update radar-trap to v2.2.1 and vis-2-widgets-radar-trap to v2.2.0

Steiger04:
Thanks!

PR #3803, opened 2024-07-10
Update radar-trap to 2.2.1

mcm1957:
Splitted from https://github.com/ioBroker/ioBroker.repositories/pull/3802


mcm1957:
RE-CHECK!

PR #3804, opened 2024-07-10
Update vis-2-widgets-radar-trap to 2.2.0

mcm1957:
splitted from https://github.com/ioBroker/ioBroker.repositories/pull/3802


mcm1957:
Errors are false positive and can be ignored.

PR #3821, opened 2024-07-17
Add egigeozone2 v1.0.5 to stable

obakuhl:
Adding egigeozone2 v1.0.5 to stable

### DIFF
--- a/sources-dist-stable.json
+++ b/sources-dist-stable.json
@@ -542,6 +542,12 @@
     "type": "household",
     "version": "1.4.15"
   },
+  "egigeozone2": {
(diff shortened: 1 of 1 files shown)


obakuhl:
I don't understand issue E407.
In addition it would be preferred to jump directly to v1.0.6 as stable (minor changes), but v1.0.6 isnt available in latest yet.

mcm1957:
E407 is caused by an error in LATEST repo file.
I'll fix it.

If you want to wait for 1.0.6 it's no problem. Simple create a PR when 1.0.6 has been available at latest for app 2 weeks and has no major new problems. I will close this PR then to wait for 1.0.6.

PR #3844, opened 2024-07-23
Update yamaha to 0.5.4

Garfonso:
Please update my adapter ioBroker.yamaha to version 0.5.4.


Garfonso:
@mcm1957 you asked my to look into this adapter. I didn't find the time. But @foxriver76 updated adapter core and it now seems to run with the controller betas (it does on my system). So we should push this to stable. 

I'll see if I come around looking deeper into this adapter (and fixing all the checker warnings) in early autumn. Haven't forgotten about it. ;-)

PR #3876, opened 2024-08-03
REMOVE psa (latest)

mcm1957:


mcm1957:
see https://github.com/TA2k/ioBroker.psa/issues/23

PR #3889, opened 2024-08-06
Add adapter flexcharts to latest

MyHomeMyData:


MyHomeMyData:
RE-CHECK!

MyHomeMyData:
RE-CHECK!

mcm1957:

reminder 13.8.2024

mcm1957:

Please note that most likely @foxriver76 or @Bluefox might give some remarks too.

- [ ] errors and warning reported by adapter checker must be fixed

   Please fix errors and warnings reported by adapter checker (excpet W400 of course).

- [ ] remove github uinstall instructions

  Installation from GitHub is stringly discouraged. So please remove such instructions from README.md. (Tests know how to install a github version anyway and normal users should not be guided this way. In general ioBroker users should be familyr with installing an adapter, so normally no guide for installation is required. Thsi relates to installation only, configuratzion should be described as good as possible.

  ref: https://github.com/MyHomeMyData/ioBroker.flexcharts/tree/main#install-flexchart-adapter

- [ ] echarts.js seems to be an external copy

  Please verify that license requirements are fulfilled.

- [ ] check and adapt testing
  
  Tests for node 18 are mandatory. 
  
  Tests must be performed at all supported platforms (linux, windows, mac) unless special hardware or software restrictions prohibit this. Please readd mac.

**To be discussed with @foxriver76 and @apollon77**

- [ ] is the design to store the configuration of the chart AND the actual date within ONE object / state really the best solution? Normally one would expect to configure the grafik at one place and to provide the date at a second place as normally only data changes during operation.

  I refer to description here: https://github.com/MyHomeMyData/ioBroker.flexcharts/tree/main#basic-concept

- [ ]  it looks like that this adapter is only some sort of webserver. Wouldn't this better implemnted as a vis component (widget?) instead of an adapter and / or to use the standard webserver support of ioBroker?
  
- [ ] This adapter does not provide any refresh mechanism on itself. It seems to process the content of a state only when the html port get queried. So vis must do a cyclic refresh. Is this the way vis works? There will be no refresh if the content of the used state is changed.

McM1957

mcm1957:
reminder 15.9.2024

MyHomeMyData:
Thanks a lot for doing the review! Hopefully, I fixed the findings so far.

Regarding ECharts licence: It appies "Apache License, Version 2.0". To my understanding usage of local copy is allowed. I added a note in License part of Readme.

Regarding basic concept:
I'm curious to hear opinions of @foxriver76 and @Apollon77.

Yes, there is no refresh mechanism on itself.

It's quite possible that the current approach isn't optimal for this concept. Unfortunately, I don't know anything about programming widgets etc., so I can't assess these alternatives.
If we go in this direction, I'll need support to get started.

MyHomeMyData:
@mcm1957 Hi Martin, I would appreciate it if we could come to a decision on how to proceed here.

If it's not feasable to use an adapter I would need your advice for a better approach.
I try to summarize the present approach:
  * adapter creates a http server on a specific port. Port no. is presently the only configuration parameter.
  * a client (browser, iFrame, ...)  using the provided http-address can hand over a set of parameters
  * based on those parameters adapter decides to grab data for EChart from an ioBroker state or from an ioBroker script via sendMessage and callback
  * user provides complete definition of EChart via state or script
  * adapter hands over received definition of EChart to Apache echart-script

Finally an EChart is shown according to users definition. It's possible to use almost full feature set of Apache ECharts within ioBrokers eco system.

MyHomeMyData:
Hi Ingo, thank you for the quick and detailed answer. I'm glad that you also think the current implementation is a possible solution.
Unfortunately, my knowledge of web development is very limited and I need a few days to get an idea. I first have to find out whether I can manage to switch to web extension.

Viele Grüße
Jürgen

Apollon77:
Hi @MyHomeMyData ... web extension is no web development! It is just to "write the nodejs stuff slighly different so that the web adapter loads it as a plugin server side. SO it is still and 100% backend effort :-)

Lets look at simpleaoi adapter.
* the code is in a class here https://github.com/ioBroker/ioBroker.simple-api/blob/master/lib/simpleapi.js which gets a webserver instance which is created outside
* https://github.com/ioBroker/ioBroker.simple-api/blob/faaedaca284819ebf2db49b130e9d33f1263ce21/io-package.json#L126 defined this file as the "extenstionpoint" that the web adapter loads
* https://github.com/ioBroker/ioBroker.simple-api/blob/master/main.js allows it to be also staretd as server which is then "one use case" if it is not used as web extension.
* https://github.com/ioBroker/ioBroker.simple-api/blob/faaedaca284819ebf2db49b130e9d33f1263ce21/main.js#L41 basically checks if user wants it to run as web extensionthen it does nothing :-)) Else you see which parts are done - and then also tge same class is used and just the webserver is created "outside"
* using https://github.com/ioBroker/ioBroker.simple-api/blob/faaedaca284819ebf2db49b130e9d33f1263ce21/admin/jsonConfig.json#L9 the user casn select the web instance this instance should extend ... if not choosen it will craete stand alone server.

More info also on https://github.com/ioBroker/ioBroker.web/blob/master/WEB-EXTENSIONS-HOWTO.md

MyHomeMyData:
Hi @Apollon77 thanks again for all the help via telegram!

Adapter now works as web extension using mode:extension. Functionality - hopefully - is the same as before.

Please do a re-check.

mcm1957:
RE-CHECK!

mcm1957:
reminder 8.10.2024

MyHomeMyData:
RE-CHECK!

mcm1957:
@MyHomeMyData 
I'm still checking with @foxriver76 what combination of mode:extension / onlyWWW is valid. Currently it looks like onlyWWW should not be set - but I still do not know for sure. Sorry.

Please ignore checker errors until the recommended combination is clearified

mcm1957:
It's me again:

I checked oth aapters configures as mode:extension:

ioBroker.flexcharts.json
ioBroker.googleauth.json
ioBroker.habpanel.json
ioBroker.proxy.json
ioBroker.weblogin.json

The correct setup seems to be:
- onlyWWW: false (or not existing) at io-package.json
- main empty (main:"") or pointing to the extension file (i.e. lib/adapter,js)

I suggest to setup flexcharts this wai for now - if you have time. I cannot ensire that his setup will be correct but as the listed adpaters have such a setup (check yourself if interested) it should work. Foxriver will try to analyze the controlelr code and  respond within the next week - at least if nothing important arises in between.

MyHomeMyData:
@mcm1957 I changed configuration according to your proposal. Hope this fits for the moment.

Now instance is displayed as "running" and gray color. Fine from my point of view.

MyHomeMyData:
RE-CHECK!

mcm1957:
RE-CHECK!

mcm1957:

PR #3890, opened 2024-08-06
Add midas-aquatemp to latest

MiRo1310:


MiRo1310:
RE-CHECK!

MiRo1310:
RE-CHECK!

mcm1957:

remember 14.08.2024

mcm1957:

I would like to give some feedback based on my personal oppinion. @Apollon77 might have additional suggestions or even a different opinion to one or the other statement. Please feel free to contact him if you cannot follow my suggestions or want to discuss some special aspects.

  The README.md of any adapter related to device or multiple devices of a manufacturer should contain a link the manufacturers website and/or to a description of the device. This is to anable new users to easily check wether they associate the adapter with the correct device(s). (See https://github.com/ioBroker/ioBroker.repositories#requirements-for-adapter-to-get-added-to-the-latest-repository paragraph 4.)

- [ ] minimize logging

  Please  reduce logging to information important for user. I.e. I doubt that a user has much benefit to see that the token has been refreshed or that the objects have been created. Such logging is OK and valuable in case of problems but I suggest to use debug level for them. Some examples

  https://github.com/MiRo1310/ioBroker.midas-aquatemp/blob/97199701ccceda5488e41f6bb1cd387fa54d508d/src/lib/token.ts#L14

  https://github.com/MiRo1310/ioBroker.midas-aquatemp/blob/97199701ccceda5488e41f6bb1cd387fa54d508d/src/lib/createState.ts#L402

- [ ] sensitive information must not be logged into logfile

  To avoid publishing sensitive information when forwarding logs, those information should never be written into the logs but either omitted or at least partly masked so that the information cannot be misused. This relates at least to usernames, passwords, unique (user assigned) keys and tokens. In addition personal information like name, address, gegraphical coordinates should not be logged too.

  see https://github.com/MiRo1310/ioBroker.midas-aquatemp/blob/97199701ccceda5488e41f6bb1cd387fa54d508d/src/lib/token.ts#L28

- [ ] access to cloud.linked-go.com

  Adapter accesses cloud.linked-go.com site. Is this site related to midas? There is no Impressum at this site. It looks a little dubious. If it is related to midas please provide some proof / link. An in any case please document the access to the website differing from the manufacturer at README so that users are informed an later questions are avoided.

  https://github.com/MiRo1310/ioBroker.midas-aquatemp/blob/97199701ccceda5488e41f6bb1cd387fa54d508d/src/lib/endPoints.ts#L9

- [ ] please review object creation
  
  Here https://github.com/MiRo1310/ioBroker.midas-aquatemp/blob/97199701ccceda5488e41f6bb1cd387fa54d508d/src/lib/saveValue.ts#L30 you create a state if it does bot exist. But you set the state value oly if it already exists - not if it has been just created. (see return). Is this really desired?

- [ ] reevaluate state roles

  value.* is always a read only state. Changeable states must be level.*. See schema definition noted above.

  See as an example - but please check every role:

  https://github.com/MiRo1310/ioBroker.midas-aquatemp/blob/97199701ccceda5488e41f6bb1cd387fa54d508d/src/lib/createState.ts#L45

  https://github.com/MiRo1310/ioBroker.midas-aquatemp/blob/97199701ccceda5488e41f6bb1cd387fa54d508d/src/lib/createState.ts#L150

- [ ] i18n cleanup suggested

  option 1 / option 2 seem to be realted to template
  Using usernamen adn Username seems confusing (or one of them is obsolete)
  https://github.com/MiRo1310/ioBroker.midas-aquatemp/blob/97199701ccceda5488e41f6bb1cd387fa54d508d/admin/i18n/de/translations.json#L2

  see config.config refresh

  Tests for node 18 are mandatory. 
  
McM1957

mcm1957:
@MiRo1310 
Did you close this PR on purpose?
Why?
Or did you close on error i.e. by deleting a branch or fork?

In any case a new PR is welcone.

MiRo1310:
I have no information about which devices are supported and nothing about the API. I received a script from a user and built it into an adapter. There are only 8 installations, and so I won't invest much more time in it at the moment, as I have little time at the moment.

PR #3894, opened 2024-08-06
Update vis-2 to 2.9.64

mcm1957:
https://github.com/ioBroker/ioBroker.vis-2/issues/460


mcm1957:
2.10.7 already exists - wait for it

PR #3908, opened 2024-08-09
Update scenes to 3.2.1

mcm1957:
https://github.com/ioBroker/ioBroker.scenes/issues/344


mcm1957:
updated due to react adaptions / admin 7

Please consider fixing issues as far as possible with acceptable effort.

PR #3955, opened 2024-08-20
Change type of chargemaster adapter to energy (latest)

hombach:
I think this fits better than vehicle ??


hombach:
RE-CHECK!

mcm1957:
RE-CHECK!

PR #3976, opened 2024-08-27
update innoxel to 0.4.2

matthsc:
Updates the innoxel adapter from 0.4.1 to 0.4.2 in stable.

The update contains adapter checker feedback and dependency updates.


mcm1957:
RE-CHECK!

mcm1957:
E124 should not be raied if nogit is set to true.
The issue is still open: https://github.com/ioBroker/ioBroker.repochecker/issues/270

PR #4019, opened 2024-09-07
Update procon-ip to 1.5.5

ylabonte:
Please update my adapter ioBroker.procon-ip to version 1.5.5.


mcm1957:
@ylabonte 
1.5.5 requires node 18 according to package.json. BUT your tests are executed at node 120 and 22 only. Tests fail with node 18.
So either add node 18 to tests and fix whatever is required to make adapetr work at node 18 OR raise node requiredments at package.json to node 20 and create a new minor release. (A patch release is not ok if requirments are changed).

Please see:
https://github.com/ylabonte/ioBroker.procon-ip/issues/113

ylabonte:
Version 1.6.0 has been released. Waiting another 14+ days before declaring it as stable...

PR #4060, opened 2024-09-19
Moved admin@7.1.1 to stable and move kisshome-research to stable

GermanBluefox:


mcm1957:
splitted into #4061 and #4062

PR #4063, opened 2024-09-19
Add  kisshome-research 1.0.4 to stable

mcm1957:
see https://github.com/ioBroker/ioBroker.repositories/pull/4060


mcm1957:
RE-CHECK!

mcm1957:
merging anyway as this adpater is for special use only

PR #4069, opened 2024-09-21
Format documentation, migrate gulpfile.js to tasks.js

GermanBluefox:
Removed gulp from dependecies


mcm1957:
Update of kissathome-reasearch mioved to #4070

PR #4089, opened 2024-09-29
Add bluesound to latest

mcm1957:


mcm1957:
replaced by https://github.com/ioBroker/ioBroker.repositories/pull/4089

mcm1957:
@Uwe1958

reminder 6.10.2024

mcm1957:

- [ ] Readme.md cosmetics (non blocking)

  List of supported commands should be formatted as list (ev. add two spaces at EOL to force linebreak or use '-')
  
- [ ] Parameter timeout and pollinTime - suggest to use type number
  
  Any reason to use "text" for parameter timeout and Pollingtime instead of number? As the data must be entered as number, the parameters should be of type number.

- [ ] jsonConfig responsive design

  Please add missing sizes (xs, xl)  
  see https://github.com/ioBroker/ioBroker.admin/blob/master/packages/jsonConfig/SCHEMA.md#common-attributes-of-controls

- [ ] handle missing IP Address

  If no IP address is configured the adapter trys to initialize axios anyway. This will likely result in an exception. Adapter should terminate or do "nothing" if no IP is set

- [ ] subscribing to objects

  Why do you subscribe to all states? Subscribing makes sense only if you will react on changes to the object - so subscribing is useful for writeable objects only - at least in most cases.

  In addition states should be subscribed only once as multiple subscribes could cause use of unneded resources and result in multiple notifications.

- [ ] Object Ids

  setState(Async) should be called without instance prefix as it always addresses the own instance. Only calls to setForeignState(Async) must provide the instance information (namespace)

```
function name2id(pName) {
}
```
  Some state Ids seem to be constructed by preset names retrieved from external device. Those names must be filtered to remove illegal characters. Buest prxis is to limit to A-Za-z0-9, but using FORBIDDEN_HARS is OK too. Spaces and dots MUST be removed in addion as severla vis cannot handle spaces within stateIds and dot have special meaning (folder seperation)

- [ ] Code to create preset objects missing?

  I did not see code to create sobjects / states defined by presets. Please indicate or use setObjectNotexists/exztendObject calls.

- [ ] Consider removing example code (non blocking)

  I would suggest to remove example code from source to ease reading. But that's not blocking of course.

- [ ] reevaluate state roles

  Writeable states must not have role value use level instead (or i.e. level.volume).

- [ ] Timer must be stopped at onUnload

  All timer / intervals must be stopped at onUnload

- [ ] prefer setTimeout over setInterval

  When using setInterval please ensure that all oprations are completed before the next interval starts. This requires that the interval time must be lager than the sum of all timeout from multiple calls within a cycle. Otherwise a new interval could be started while the old one is still in progress. This might sum up, consume more and more resources and finally crash the complete ioBroker installation.
  
  Tests for node 18 are mandatory as this is the recommended node version unless adapter requires node 20 minimum.
  
McM1957

reminder 11.10.2024

Uwe1958:
@mcm1957 

Thanks for your feedback. I started fixing the issues as requested.

One problem though:

Regarding your request on setState() I always have to add the instance prefix (in fact it is always automatically entered when I type setState) .

Without the instance prefix I get the error

error: bluesound.0 (300351) unhandled promise rejection: setState is not defined
error: bluesound.0 (300351) ReferenceError: setState is not defined

Can you tell me how I can solve the issue raised.

Thanks.

mcm1957:
im referring tonthe state id.

rhis.setState is ok an required. 
To set the dtate control.volume its sufficient to use this.setStste('control.volume', ...). Rhe full pathvincludibg bluesound.0. would require setForeignStste othwise the adapter bame will be duplicated atcstste id.

Uwe1958:
@mcm1957 

Thanks got it now: ObjectId to omit the instance.

Uwe1958:
Code to create preset objects is included within onReady using the result from the /Presets call. It returns the 'id' used for the ObjectId always as a number like 1, 2 ... To be really sure I will include the replacement of FORBIDDEN_CHARS as requested.

The objects are then created dynamically using setObjectNotExistsAsync using the helper function (via utils.js). The names are hardcoded there and thus should not create any problems with special characters.

Uwe1958:
@mcm1957 
Changes are made as requested. Please advice whether I should create a new release after your review or how to proceed.
Thanks
Uwe

Uwe1958:
RE-CHECK!

mcm1957:
RE-CHECK!

Uwe1958:
RE-CHECK!

simatec:
The jsonConfig looks good and fulfills all resolutions

mcm1957:
** REVIEW CHECKLIST - WORK IN PROGRESS **

- [ ] Readme.md cosmetics (non blocking)

  List of supported commands should be formatted as list (ev. add two spaces at EOL to force linebreak or use '-')
  
- [ ] Parameter timeout and pollinTime - suggest to use type number
  
  Any reason to use "text" for parameter timeout and Pollingtime instead of number? As the data must be entered as number, the parameters should be of type number.

- [ ] jsonConfig responsive design

  Please add missing sizes (xs, xl)  
  see https://github.com/ioBroker/ioBroker.admin/blob/master/packages/jsonConfig/SCHEMA.md#common-attributes-of-controls

- [ ] handle missing IP Address

  If no IP address is configured the adapter trys to initialize axios anyway. This will likely result in an exception. Adapter should terminate or do "nothing" if no IP is set

- [ ] subscribing to objects

  Why do you subscribe to all states? Subscribing makes sense only if you will react on changes to the object - so subscribing is useful for writeable objects only - at least in most cases.

  In addition states should be subscribed only once as multiple subscribes could cause use of unneded resources and result in multiple notifications.

- [ ] Object Ids

  setState(Async) should be called without instance prefix as it always addresses the own instance. Only calls to setForeignState(Async) must provide the instance information (namespace)

```
function name2id(pName) {
}
```
  Some state Ids seem to be constructed by preset names retrieved from external device. Those names must be filtered to remove illegal characters. Buest prxis is to limit to A-Za-z0-9, but using FORBIDDEN_HARS is OK too. Spaces and dots MUST be removed in addion as severla vis cannot handle spaces within stateIds and dot have special meaning (folder seperation)

- [ ] Code to create preset objects missing?

  I did not see code to create sobjects / states defined by presets. Please indicate or use setObjectNotexists/exztendObject calls.

- [ ] Consider removing example code (non blocking)

  I would suggest to remove example code from source to ease reading. But that's not blocking of course.

- [ ] reevaluate state roles

  Writeable states must not have role value use level instead (or i.e. level.volume).

- [ ] Timer must be stopped at onUnload

  All timer / intervals must be stopped at onUnload

- [ ] prefer setTimeout over setInterval

  When using setInterval please ensure that all oprations are completed before the next interval starts. This requires that the interval time must be lager than the sum of all timeout from multiple calls within a cycle. Otherwise a new interval could be started while the old one is still in progress. This might sum up, consume more and more resources and finally crash the complete ioBroker installation.
  
  Tests for node 18 are mandatory as this is the recommended node version unless adapter requires node 20 minimum.
  
McM1957

mcm1957:
A hint for migration of EsLint (not blocking remark)

If you plan to migrate EsLint from 8.x.x to current release 9.x.x please migrate to use @iobroker/eslint-config. This is a new central configurarion of eslint-rules which should be used for all (new) adapters. But its not mandatory. A migration / installation guide is can be found here:

https://github.com/ioBroker/ioBroker.eslint-config/blob/main/MIGRATION.md

mcm1957:
@Uwe1958 

PR #4208, opened 2024-10-23
Update reolink to 1.1.2

oelison:
Please update my adapter ioBroker.reolink to version 1.1.2.


mcm1957:
testing with node.js 22 is missing.
please check / mrrge https://github.com/aendue/ioBroker.reolink/pull/117

oelison:
How to set as stable with a change? Make no sense to me.

oelison:
The test for node 22 is now integrated. But a release should create a new version. So there is no change and this could be integrated or not?

PR #4236, opened 2024-10-28
Add TeslaFi to latest

hombach:


hombach:
RE-CHECK!

simatec:
jsonConfig looks good...

mcm1957:
RE-CHECK!

hombach:
RE-CHECK!

hombach:
"xs" should specify a value of "12" at admin/jsonConfig.json/items/_StandardTab/_donate  ->  has to be 6, otherwise button is too big

I'm no friend of automated translations - so I don't use i18n

mcm1957:
RE-CHECK!

mcm1957:
Your TOO fast :-)
Size check has been disabled for divider an staticImage in the meantime. Checks are adjustes to results in real life. :-)

I18n is a warning only which will not be considered blocking.

hombach:
:)  -  so all this 5 times 12 for dividers could be removed?.... -> next release ;)

... so must be fixed label isn't relevant anymore? Or do I have to change something?

mcm1957:
Sorry, but due to travel / absence I will not able to do the review before ioBroker Meeting Solingen next weekend.

mcm1957:
@hombach 

- [ ] TeslaFi related to tibber?

  Is teslafi related to tibber or is this loga a copy/paste error?

  'pull of homes from Tibber-Server'

- [ ] unused onMessage handler

  onMessage handler has no obvious functionality. Please remove from code an config.

```
function name2id(pName) {
}
```

  setObject(Async) creates a new object whenever it is called. You should avoid using setObject unless you really want to create new objects. Creating a new object deletes any userConfig, i.e. configures history logging which is most likely not desired. Most likely either setObjectNotExists oder (even better in most cases) extendObject should be used. Extend Object ensures that attributes like role, read/write flag etc. are set as specified and hence extendObjet should be preferred when called once during startup of the adapter.

- [ ] reevaluate state roles

  Only the values specified here (https://github.com/ioBroker/ioBroker.docs/blob/master/docs/en/dev/stateroles.md) may be used as state roles. Do not invent new names as this might disturb the functionality of other adapters or vis.

  One example: role 'value' must not be writeable.

  Limit all intervals used to access cloud services to a reasonable value. If users configure an inapprobiate access rate this could lead to suspending the service by the provider for all ioBroker users.  
  
- [ ] linter setup

McM1957

**Please add a comment when you have reviewed and fixed the suggestions or at least commented the suggestions and you think the adapter is ready for a re-review!**

hombach:
TeslaFi related to tibber?
Is teslafi related to tibber or is this loga a copy/paste error?
-> yes, copy/paste error -> FIXED

unused onMessage handler
onMessage handler has no obvious functionality. Please remove from code an config.
-> REMOVED

-> NO USER DRIVEN object ids used

avoid usage of setObject
-> NOT USED

PART 1 - more to follow

hombach:
@mcm1957 Hi Martin, may I ask for a review this weekend... I'd need my fork of iobroker.repositories for my other adapters to get some things released... ;)

mcm1957:
Will check if I have time tomorrow. But cannot promise.
But what*s the problem? You can release differnet adapters independently. Simply use ba seperate branch per Adapter. If you use www.iobroker.dev or edit dierectly at www.github.com a new branch and the corresponding PR will be created automatically.

Example for tibberlink:
https://www.iobroker.dev/adapter/tibberlink/releases
And click to "set as stable" (if a new stable is possible)

If you got a reminder issue, those links are availbale there and need to be clicked only:
image

hombach:
LOL  -  you're right - I'm stupid ;)

mcm1957:
> LOL - you're right - I'm stupid ;)

thats defintstly wrong !

mcm1957:
REVIEW - WORK IN PROGRESS

- [ ] TeslaFi related to tibber?

  Is teslafi related to tibber or is this loga a copy/paste error?

  'pull of homes from Tibber-Server'

- [ ] unused onMessage handler

  onMessage handler has no obvious functionality. Please remove from code an config.

```
function name2id(pName) {
}
```

  setObject(Async) creates a new object whenever it is called. You should avoid using setObject unless you really want to create new objects. Creating a new object deletes any userConfig, i.e. configures history logging which is most likely not desired. Most likely either setObjectNotExists oder (even better in most cases) extendObject should be used. Extend Object ensures that attributes like role, read/write flag etc. are set as specified and hence extendObjet should be preferred when called once during startup of the adapter.

- [ ] reevaluate state roles

  Only the values specified here (https://github.com/ioBroker/ioBroker.docs/blob/master/docs/en/dev/stateroles.md) may be used as state roles. Do not invent new names as this might disturb the functionality of other adapters or vis.

  One example: role 'value' must not be writeable.

  Limit all intervals used to access cloud services to a reasonable value. If users configure an inapprobiate access rate this could lead to suspending the service by the provider for all ioBroker users.  
  
- [ ] linter setup

mcm1957:
Thanks for fixing most issues, The following issues still need attention:

setObject(Async) creates a new object whenever it is called. You should avoid using setObject unless you really want to create new objects. Creating a new object deletes any userConfig, i.e. configures history logging which is most likely not desired. Most likely either setObjectNotExists oder (even better in most cases) extendObject should be used. Extend Object ensures that attributes like role, read/write flag etc. are set as specified and hence extendObjet should be preferred when called once during startup of the adapter.

You are using setObject i.e. here:
https://github.com/hombach/ioBroker.teslafi/blob/60b533418f207cdfa9fd88a7d8c46ce3b2e82aa8/src/lib/projectUtils.ts#L179

Wouldn't it be sufficient to use extendObject? setObject will delete the old object and hence delete all user defined attribute, i.e. setting for history adapter.

- [ ] reevaluate state roles

Only the values specified here (https://github.com/ioBroker/ioBroker.docs/blob/master/docs/en/dev/stateroles.md) may be used as state roles. Do not invent new names as this might disturb the functionality of other adapters or vis.

  role 'value' must not be writeable.
https://github.com/hombach/ioBroker.teslafi/blob/60b533418f207cdfa9fd88a7d8c46ce3b2e82aa8/src/lib/projectUtils.ts#L218

  role 'indicator' must not be writeable.
https://github.com/hombach/ioBroker.teslafi/blob/60b533418f207cdfa9fd88a7d8c46ce3b2e82aa8/src/lib/projectUtils.ts#L269

  You're using setInterval here
https://github.com/hombach/ioBroker.teslafi/blob/60b533418f207cdfa9fd88a7d8c46ce3b2e82aa8/src/main.ts#L71

check and limit intervals used to access cloud services

- [ ] Limit all intervals used to access cloud services to a reasonable value. If users configure an inapprobiate access rate this could lead to suspending the service by the provider for all ioBroker users.
  
  Please add a suitable lower limit too:
  https://github.com/hombach/ioBroker.teslafi/blob/60b533418f207cdfa9fd88a7d8c46ce3b2e82aa8/src/main.ts#L71

hombach:
_**-> avoid usage of setObject**_

This is only used for adapter setup.... you have to read the total code - automatic scanning for key words dosn't work this time....
`				if (!forceMode) {
					await this.adapter.setObjectNotExistsAsync(stateName, {
						type: "state",
						common: commonObj,
						native: {},
					});
				} else {
					await this.adapter.setObjectAsync(stateName, {
						type: "state",
						common: commonObj,
						native: {},
					});
				}`

projectUtils.ts is a helper class I'm using in most adapters

**_-> reevaluate state roles_**

same, projectUtils.ts is a helper class I'm using in most adapters. The fact it's able to set up writable states doesn't tell it's used in the adapter. IUpo to now, all states are read only, as they are just states polled from the car. They have special background for the car but most of them no dedicated type in ioBroker. So I would prefer generall roles.

**_-> consider using adapter.setTimeout / this.setTimeout instead of (standard) setTimeout_**

 WiP
 
**_-> Limit all intervals used to access cloud services_**

allready done .... jsonConfiig.json:
`				"UpdateInterval": {
					"type": "number",
					"label": "LBL_UpdateInterval",
					"hidden": "!data.TeslaFiAPIToken",
					"min": 10,
					"max": 86400,
					"step": 1,`

hombach:
DONE

mcm1957:
-> avoid usage of setObject

Well extendObject would be the better solution anyway in my oppinion. But finally it's your decision as using setObject is strongly discouraged but not forbidden. So its non blocking.

-> reevaluate state roles
The mentioned roles must not be used with writeable statesat any time. So there is no need to allow the code to set writeable to true. I do not see why code creating incorrect states should be present. Please remove writeable option or pass role as parameter setting it correctly. As you use this code at several apaters the fix shpould be applied everywhere to avoid incorrect states.

hombach:
Hi Martin, adding the role as optional parameter in case of setting up writeable states is a good idea... implemented in 0.4.4.

hombach:
added a few more precise state roles in 0.4.5

mcm1957:
@hombach 

hombach:
Hi Martin, kann gerne was posten - darf man aber ja in der Catagory nicht.

mcm1957:
Als dev darfst du es.
Bitte ping HOMORAN via PN / Chat an dass er dich freischaltet

PR #4238, opened 2024-10-29
Add airquality to latest

raschy:
Please add my adapter ioBroker.airquality to latest.

### DIFF
--- a/sources-dist.json
+++ b/sources-dist.json
@@ -39,6 +39,11 @@
     "icon": "https://raw.githubusercontent.com/weggetor/ioBroker.airconwithme/main/admin/airconwithme.png",
     "type": "climate-control"
   },
+  "airquality": {
+    "meta": "https://raw.githubusercontent.com/raschy/ioBroker.airquality/main/io-package.json",
+    "icon": "https://raw.githubusercontent.com/raschy/ioBroker.airquality/main/admin/airquality.png",
+    "type": "weather"
+  },
   "airzone": {
     "meta": "https://raw.githubusercontent.com/SilentPhoenix11/ioBroker.airzone/master/io-package.json",
     "icon": "https://raw.githubusercontent.com/SilentPhoenix11/ioBroker.airzone/master/admin/Airzone.png",


simatec:
jsonConfig looks good. However, there is not really anything to configure. Is that the intention?

raschy:
Oh, yes. Under “Stations”, the user can/should enter the station(s) from which he wants to have measured values. That's all you need.

raschy:
Issue 1 [E254] and [W040] corrected, solved in version 0.0.4

mcm1957:
RE-CHECK!

mcm1957:
Sorry, but due to travel / absence I will not able to do the review before ioBroker Meeting Solingen next weekend.

mcm1957:
@raschy 

- [ ] package.json should require up to date js-controller and admin version

  ```
        "dependencies": [
            {
                "js-controller": ">=5.0.19"
            }
        ],
        "globalDependencies": [
            {
                "admin": ">=6.17.14"
            }
        ]
  ```

  As this adapter uses jsonConfig it should require admin v6 at least. Earlier releases have too many missing updates.

- [ ] directory 'logs'

  Does the directory "logs" really contain information needed for development or has it been commited by error?

- [ ] consider making this adapter a scheduled adapter

  As this adapter executes every 15 minutes only please evaluate whether it shouldn't be converted to a scheduled adapter. It looks like an overhead to let a process wait most of the time.

- [ ] object ids and name attribute of objects

  Object Ids must be named in english.

  The "name" attribute of all objects / states must be either a string in english or an i18n multilingual object with all supported languages. The only exception would be if the name attribute is directly delivered by the external system - which isn't he case here. I would suggest to use i18n 

- [ ] reevaluate state roles

image

  Role 'number' does not exist ("Anzahl Messtypen")
  Role "state" should only be used if no better role available. At least role value should be ok.
  "Letzte Messung" should be role 'date' (eventually convert to make it parseable by new(date). See https://github.com/ioBroker/ioBroker.docs/blob/master/docs/en/dev/stateroles.md

- [ ] linter setup

McM1957

reminder 10.12.2024

raschy:
Hopefully all bugs fixed and also migrated to ESLint 9.

mcm1957:
**RE-REVIEW - work in progress** 

- [ ] package.json should require up to date js-controller and admin version

  ```
        "dependencies": [
            {
                "js-controller": ">=5.0.19"
            }
        ],
        "globalDependencies": [
            {
                "admin": ">=6.17.14"
            }
        ]
  ```

  As this adapter uses jsonConfig it should require admin v6 at least. Earlier releases have too many missing updates.

- [ ] directory 'logs'

  Does the directory "logs" really contain information needed for development or has it been commited by error?

- [ ] consider making this adapter a scheduled adapter

  As this adapter executes every 15 minutes only please evaluate whether it shouldn't be converted to a scheduled adapter. It looks like an overhead to let a process wait most of the time.

- [ ] object ids and name attribute of objects

  Object Ids must be named in english.

  The "name" attribute of all objects / states must be either a string in english or an i18n multilingual object with all supported languages. The only exception would be if the name attribute is directly delivered by the external system - which isn't he case here. I would suggest to use i18n 

- [ ] reevaluate state roles

image

  Role 'number' does not exist ("Anzahl Messtypen")
  Role "state" should only be used if no better role available. At least role value should be ok.
  "Letzte Messung" should be role 'date' (eventually convert to make it parseable by new(date). See https://github.com/ioBroker/ioBroker.docs/blob/master/docs/en/dev/stateroles.md

- [ ] linter setup

mcm1957:
RE-REVIEW!

mcm1957:
RE-CHECK!

mcm1957:
**RE-REVIEW - work in progress worksheet only - please ignore for now** 

- [ ] package.json should require up to date js-controller and admin version

  ```
        "dependencies": [
            {
                "js-controller": ">=5.0.19"
            }
        ],
        "globalDependencies": [
            {
                "admin": ">=6.17.14"
            }
        ]
  ```

  As this adapter uses jsonConfig it should require admin v6 at least. Earlier releases have too many missing updates.

- [ ] directory 'logs'

  Does the directory "logs" really contain information needed for development or has it been commited by error?

- [ ] consider making this adapter a scheduled adapter

  As this adapter executes every 15 minutes only please evaluate whether it shouldn't be converted to a scheduled adapter. It looks like an overhead to let a process wait most of the time.

- [ ] object ids and name attribute of objects

  Object Ids must be named in english.

  The "name" attribute of all objects / states must be either a string in english or an i18n multilingual object with all supported languages. The only exception would be if the name attribute is directly delivered by the external system - which isn't he case here. I would suggest to use i18n 

- [ ] reevaluate state roles

image

  Role 'number' does not exist ("Anzahl Messtypen")
  Role "state" should only be used if no better role available. At least role value should be ok.
  "Letzte Messung" should be role 'date' (eventually convert to make it parseable by new(date). See https://github.com/ioBroker/ioBroker.docs/blob/master/docs/en/dev/stateroles.md

- [ ] linter setup

mcm1957:
@raschy 
Thanks for fixing most issues. The following issue still exists

- [ ] object ids and name attribute of objects

  Object Ids must be named in english.


The following issues are new:

- [ ] Adapter does not terminate if no stations are configured and causes "already running errors"

  see https://github.com/raschy/ioBroker.airquality/issues/19
  This problem must be fixed

- [ ] unneeded ObjectExists

  as setObjectNotExists already included the check whether an object esists an explicit check is not needed
https://github.com/raschy/ioBroker.airquality/blob/c1aa223da0d0fb51211e392f1d19b20194cdedef/src/main.ts#L352

mcm1957:
@raschy 

Did not get a feedback for my last comment. Whats the status from your side?
Are the two topic mentioned above fixed?
Please et me know ich a recheck is appropiate and / or if you need some help.

reminder 22.1.2025

raschy:
Sorry, I must have forgotten to reply to this. I think these problems have been fixed in the current version. I also made the adapter available for testing today.

mcm1957:
RE-CHECK!

mcm1957:
OK, looks goog now

mcm1957:

raschy:
Thank you. Tester request has already been created:
 https://forum.iobroker.net/topic/79286/test-adapter-airquality-v0-1-4-github-latest?_=1737410420962

PR #4251, opened 2024-11-02
Add waste-collection-erding to latest

Sickboy78:
Please add my adapter ioBroker.waste-collection-erding to latest.

### DIFF
--- a/sources-dist.json
+++ b/sources-dist.json
@@ -2929,6 +2929,11 @@
     "icon": "https://raw.githubusercontent.com/pottio/ioBroker.warp/main/admin/warp.png",
     "type": "vehicle"
   },
+  "waste-collection-erding": {
+    "meta": "https://raw.githubusercontent.com/Sickboy78/ioBroker.waste-collection-erding/main/io-package.json",
+    "icon": "https://raw.githubusercontent.com/Sickboy78/ioBroker.waste-collection-erding/main/admin/logo.png",
+    "type": "date-and-time"
+  },
   "waterkotte-easycon": {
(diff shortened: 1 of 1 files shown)


mcm1957:
@klein0r 
@Sickboy78 

Does the existing adapter **trashschedule** (https://github.com/klein0r/ioBroker.trashschedule) already support "erding"?
If  not, wouldn't it be an option to integrate this supplier into trashschedule? I assume that klein0r will accept a PR for this.
Do we need another waste adapter dedicated to a region?

@Apollon77 
Some general oppinion?

Sickboy78:
Imho Erding is currently not supported since LK Erding does not provide an ICAL calender (only ICS and CSV for download) and the API portal.awido.de is not integrated into trashschedule.
But it would totally make sence to integrate it, if this is possible. Need to take a closer look though ...

mcm1957:
I would suggest that you take a look if you have time and eventually discuss a solution with klein0r. Let's wait for the result. 
Thanks for you cooperation.

simatec:
If the adapter can be integrated independently and not in Trashschedule, I would ask you to migrate the adapter to jsonConfig.

Sickboy78:
@klein0r @mcm1957 

While digging into the Awido API for integration with trashschedule, I found that it is used by nearly 50 districts and cities, i.e. it would add hundreds of cities to be available in trashschedule.
So heres the PR:
https://github.com/klein0r/ioBroker.trashschedule/pull/285

Sickboy78:
Functionality was integrated with trashschedule, therfore canceling this PR.

PR #4256, opened 2024-11-03
Add controme to latest

MadErstam:
Added controme adapter to ioBroker latest repository.

### DIFF
--- a/sources-dist.json
+++ b/sources-dist.json
@@ -339,6 +339,11 @@
     "icon": "https://raw.githubusercontent.com/schmupu/ioBroker.contactid/master/admin/contactid.png",
     "type": "network"
   },
+  "controme": {
+    "meta": "https://raw.githubusercontent.com/MadErstam/ioBroker.controme/main/io-package.json",
+    "icon": "https://raw.githubusercontent.com/MadErstam/ioBroker.controme/main/admin/controme.png",
+    "type": "climate-control"
+  },
   "coronavirus-statistics": {
     "meta": "https://raw.githubusercontent.com/DrozmotiX/ioBroker.coronavirus-statistics/main/io-package.json",
     "icon": "https://raw.githubusercontent.com/DrozmotiX/ioBroker.coronavirus-statistics/main/admin/coronavirus-statistics.png",


mcm1957:
Due to outside activities (ioBroker meeting at solingen) this PR will most likely not be processed  before 12.11.2024.

MadErstam:
RE-CHECK!

MadErstam:
RE-CHECK!

MadErstam:
RE-CHECK!

simatec:
Except for the i18n translations, the jsonConfig looks good and fulfills all resolutions

MadErstam:
RE-CHECK!

MadErstam:
> Except for the i18n translations, the jsonConfig looks good and fulfills all resolutions

I18N translations were already directly included in the jsonConfig.json, but were missing the Ukrainian translations. I have added that now in the latest version. Unfortunately, the warning remains. If you could give me a hint how to fix the warning, I would be happy to make these changes.

mcm1957:
jsonConfig should use traslation stored at i18n files. (admin/i18n/{lang}.json or admin/i18n/{lang}/translation.json). This enables integration into weblate and usage of dev-tools npm run translate.  Direct translations inside jsonConfig is (still) supported but not recommended.

So the warning is correct.

MadErstam:
RE-CHECK!

MadErstam:
> jsonConfig should use traslation stored at i18n files. (admin/i18n/{lang}.json or admin/i18n/{lang}/translation.json). This enables integration into weblate and usage of dev-tools npm run translate. Direct translations inside jsonConfig is (still) supported but not recommended.
> 
> So the warning is correct. Please consider migrating to transaltions located at external files.

OK, understood and changed.

mcm1957:
Sorry, but due to travel / absence I will nozt able to do the review before ioBroker Meeting Solingen next weekend.

mcm1957:
@MadErstam 

I would like to give some feedback based on my personal opinion. @Apollon77 might have additional suggestions or even a different opinion to one or the other statement. Please feel free to contact him if you cannot follow my suggestions or want to discuss some special aspects.

- [ ] Readme.md needs formatting

  The README.md of this adapter needs formatting. Please check at https://github.com/MadErstam/ioBroker.controme?tab=readme-ov-file#iobrokercontrome. Especially the table of states available is corrupt.

- [ ] package.json should require up to date js-controller and admin version

  ```
        "dependencies": [
            {
                "js-controller": ">=5.0.19"
            }
        ],
        "globalDependencies": [
            {
                "admin": ">=6.17.14"
            }
        ]
  ```

- [ ] reevaluate state roles

  Role 'value' is reserved for read-only states. Consider using role 'level'if no beeter matching role exists.

- [ ] unused onMessage handler

  onMessage code seems to not provide any functionality. Please remove unused onMeassge handler from code and config (if enabled there).

- [ ] Why do you set "common.getHistory"

  io-package.json defines 'common.getHistory' as true. I cpould not see any message handler processing such a request. Please check and correct or comment. (https://github.com/MadErstam/ioBroker.controme/blob/f4a22fe1ced429deff16daabefc9e3c0e49960a6/io-package.json#L138)

- [ ] linter setup

McM1957

reminder 11.12.2024

MadErstam:
Hi @mcm1957!

Thank you for your comments! I think I fixed all of them. See one deviation to what you recommended regarding package.json below.

**I would consider the adapter ready for a re-review!**

* ☑️ Readme.md needs formatting

Done, was an error in the table formatting

* ☑️  package.json should require up to date js-controller and admin version

Shouldn't these go in io-package.json rather than to package.json? I did not find these two dependencies in any package.json of other adapters I have checked, but rather in io-package.json, where at least the js-controller dependency was already included. Anyway, I have included them now in io-package.json.

* ☑️ reevaluate state roles

Done and fixed. All state roles should now comply with the requirements. Read/write functionality also now matches to the role.

* ☑️ consider using adapter.setTimeout / this.setTimeout instead of (standard) setTimeout

Done.

* ☑️ check and limit configurable timeouts / intervals

Done. The timeout was already limited due to the variable defining the duration having a maximum value. For the interval, I included an upper limit now well below the max delay.

* ☑️ unused onMessage handler

Done. Was really unused and has been fully removed.

* ☑️ Why do you set "common.getHistory"

Was really not needed. Removed the setting.

* ☑️ linter setup

Done and fixed all new lint errors (prettier) and warnings.

mcm1957:
package.json should require up to date js-controller and admin version

Yes of course. Was a typo in my side. Sorry

mcm1957:
Thnaks for ultra fasct reaction and fixes

mcm1957:

PR #4286, opened 2024-11-14
Update roborock to 0.6.17

copystring:
Please update my adapter ioBroker.roborock to version 0.6.17.


mcm1957:
Please fix issues reported by adapter checker and add missing size attributes.

mcm1957:
responsive design is Ok according to https://github.com/copystring/ioBroker.roborock/issues/679#issuecomment-2420361373.

Adding size attributes should not be a problem as you use newline for every element anyway. But thats nothing to be discussed here.

PR #4292, opened 2024-11-16
Update discord stable to v2.5.2

crycode-de:
v2.5.2 is the same as 2.5.1 but with the latest fixes for sizes in jsonConfig.
Ref crycode-de/ioBroker.discord#76


mcm1957:
RE-CHECK!

PR #4293, opened 2024-11-16
Update ds18b20 stable to 3.0.1

crycode-de:
v3.0.1 is the same as v3.0.0 but with the latest fixes for sizes in jsonConfig.


crycode-de:
RE-CHECK!

mcm1957:
RE-CHECK!

PR #4294, opened 2024-11-16
Update odl stable to v4.0.2

crycode-de:
v4.0.2 is the same as v4.0.1 but with the latest fixes for sizes in jsonConfig.


mcm1957:
RE-CHECK!

PR #4319, opened 2024-11-25
panasonic-comfort-cloud 3.0.3

marc2016:


marc2016:
RE-CHECK!

marc2016:
RE-CHECK!

marc2016:
RE-CHECK!

marc2016:
problems are fixed now. Sorry for the delay.

mcm1957:
RE-CHECK!

marc2016:
Panasonic changed API again and adapter doesn't work with version 3.0.2 anymore.

mcm1957:
Thanks for Info.
Then lets hurry to get 3.0.3 out and running.

P.S: 63% of users still use 3.0.2 ???

marc2016:
I don't  why so many users still use version 3.0.2. 
I think they don't use the beta repo and don't follow the discussions on the GitHub repo of my adapter.

Thank you for your fast reaction! :-)

mcm1957:
No problem. I only wonder why so many user use a non function version withoput reporting problems ... 
Well "Never trust a statistic unless you have manipulated it yourself"
:-) :-) 

P.S: 3.0.3 should be available at stable already - at least after updating local cache ("Nach updates suchen")

PR #4381, opened 2024-12-17
Add  "syrconnect" to latest

eifel-tech:


mcm1957:
reminder 24.12.2024

mcm1957:

Thanks

eifel-tech:
> To verify the Object structure of this adapter during REVIEW please export the object structure of a working installation and attach the file i. You find a guide how to export the object struture here: https://github.com/ioBroker/ioBroker.repochecker/blob/master/OBJECTDUMP.md
> 
> Thanks

@mcm1957 [Here](https://github.com/user-attachments/files/18516119/syrconnect.json) you can have a look of the object structure.

mcm1957:
@eifel-tech 

- [ ] Readme.md is very short

   Thanks for providing as very good documentation in german and english. But please add one paragraph directly in README.md describing the purpose of the adapter. Most likely the first paragraph of the english docu will be sufficient.

- [ ] config parameter webPort at adminUI

  The config parameter webPort specifies a (tcp) port. So this paramater should be of type number. Do you have any special purpos to allow any test entry here?

- [ ] cache known devices / objects

  It looks like every request received causes a getObjectAsync to check if states have been created already. This is an avoidable load to the ioBroker database. Please cache objects already accessed during last start of adapetr to getObjectAsync is only called once per object after start of adapter. See https://github.com/eifel-tech/ioBroker.syrconnect/blob/703f981c6bb2a63bf8297de1936ecadc79d05093/main.js#L337

- [ ] do not subscribe one state more than once

  DO not subscribe identical stateIds nore than once. Especially do not subrive over an over again whenever a state is written. If you subscribe to all states of this adapter it should be sufficient to issue subscribe('*') once at onReady.

- [ ] All objects must be created when using hirarchy

  Your objects are structured hirarchical lex10.nnnnn.ABC. It looks like that neither an object for lex10 nor the serielnumber/id is created. Those objects must follwo the rule device -> channel > state. Device and or channel may be omitted but the must be a maximum of one device and one channel within one leave. Objects of type folder can be placed at any intermediate position. 

  But looking at the code the objects should be created. Please check code execution an chack at your object view if a type is listed (device and channel) 

Image

- [ ] object info is missing at io-package.json

  Definition and creation of info folder is missing at io.package.json.
  see https://github.com/ioBroker/ioBroker.example/blob/master/JavaScript/io-package.json
  Please add
  ```
  "instanceObjects": [
    {
      "_id": "info",
      "type": "channel",
      "common": {
        "name": {
          "en": "Information",
          "de": "Informationen",
          "ru": "Информация",
          "pt": "Informação",
          "nl": "Informatie",
          "fr": "Informations",
          "it": "Informazioni",
          "es": "Información",
          "pl": "Informacje",
          "uk": "Інформація",
          "zh-cn": "资料"
        }
      },
      "native": {}
    },
    ....
    ```

- [ ] reevaluate state roles

  Writeable state must not have a role value. Use level instead if not dedictated role is available

  All stateIds must be named using english wording unless the data is directly received from an external source. (Thats ok)
  All names must use either english wording or use i18n multilanguage setup. 

  So please change the german names to either english or use i18n objects containing at least english and german translations. You can generate such blocks i.e. by using https://translator-ui.iobroker.in/

- [ ] linter setup

  Please eveluate to use standard iobroker linter setup if acceptable. This is located at @iobroker/eslint-config. See migration guide at https://github.com/ioBroker/ioBroker.eslint-config/blob/main/MIGRATION.md. Note that this point will NOT be blocking - consider it an suggestion only.
  
  
McM1957

reminder 31.1.2025

mcm1957:
@eifel-tech 

Just another issue poped up_

The issue

https://github.com/eifel-tech/ioBroker.syrconnect/issues/218

has been closed by you. If the content of the issue is correct this adapter does NOT work with up to date firmware and you do not plan to enhance it. Is this correct?

So at least a BIG warning must be added to README.md that this adapter does no longer work with an up to date firmware.

In addition I would ask if you feel that really a good idea to add an adapter to the repositories which does not work with actual firmware? According to statistics currently app 17 installations exist. Please comment your oppinion and future plans.

Ref:
Issue https://github.com/eifel-tech/ioBroker.syrconnect/issues/21
Forum https://forum.iobroker.net/topic/79479/syrlex-plus-10-connect-sl-app-reverse-engineering

eifel-tech:
@mcm1957 

First and foremost, I developed the adapter for MY setup. Why should I change this so that it works for others but then maybe no longer works for me??? I can't just start developing anything without being able to test it. 
And if you think 17 installations aren't enough for the community, I'm fine with it - less work for me.

mcm1957:
Sorry, I did not mean that 17 installations are not enough.

The problem I saw was that the adapter seems to be outdated already as it does not support up to date firmware and you do not plan to investigate / develop it further.

PR #4416, opened 2024-12-28
Add dreame adapter to latest

TA2k:


TA2k:
RE-CHECK!

TA2k:
RE-CHECK!

mcm1957:
reminder 4.1.2025

mcm1957:

Thanks

TA2k:
[dreame.0.json](https://github.com/user-attachments/files/18539890/dreame.0.json)

mcm1957:
@TA2k 

- [ ] setObjectNotExists is not synchronous

  setObjectNotExists is an asynchronous function despite its name. Only the callback (not used by your code) is guaranteed to be executed after creation of the object. So please evalute wether ist really save to use setObjectNotExists without a callback. You may use setObjectNotExistsAsync with await or use await together with setObjectNotExists when reuiring js.controller 7.0.4 minimum (https://github.com/ioBroker/ioBroker.js-controller/blob/master/CHANGELOG.md#704-2024-12-04---lucy)

  Please adapt as most likely no other mechanismen than timings will gurantee that objects are really cerated befor first usage.

- [ ] setState is not synchronous
  
  setState is an asynchronous function too. The ste need not be store when the function returns. Please evaluate if an asynchronous operation is ok (setting an state fully asynchronous might be ok) or use await. see https://github.com/TA2k/ioBroker.dreame/blob/a7036514ce714d7ce40e0ded3337d382509818df/main.js#L825

  ```
  function name2id(pName) {
  }
  ```

  You retrieve the deviceId directly from the external cloud. This information is used as objectIds. So please filter the received information to ensure that no illegal characters ar contained.

- some objects are missing

  all intermediate objects must be created. 
   - `dreame.0.XXXXXXXX.general.deviceInfo.defaultVendors`
   - `dreame.0.XXXXXXXX.general.deviceInfo.extendScType` 
   - `dreame.0.XXXXXXXX.map.cleanset.2`
   - possibly more
   -  
  are not created. Please check / add creation.

- [ ] reevaluate state roles

  Some notes:

 - XXXXXXXX.map.cleanset.2.RoomOrder is writeable, so it should be level (value required write:false)
 - role 'array' does not exist. Please change. 

- [ ] linter setup

  Please use standard iobroker linter setup if acceptable. This is located at @iobroker/eslint-config. See migration guide at https://github.com/ioBroker/ioBroker.eslint-config/blob/main/MIGRATION.md.

  This is a suggestion only and not blocking. Decision is up to maintainer.
  
  
McM1957

reminder 1.2.2025

TA2k:
1. I will not add official links to vendors or photos to prevent copyright infringement this adapter is not related to the vendor I'm not allowed to use the website or photos
2. I increase min jscontroller to 7.0.4
3. I don't need the result or wait for the setState to not block the code I don't use await
4.  Added replace function for the cleanest part
5. This objects are created via json2iob or for all cleanser part from getType() function
6. Fixed
7. I use my own linter config

mcm1957:
@TA2k 
Thanks for the fast reaction.

Please see my comments and the 2 open issues:

1. I will not add official links to vendors or photos to prevent copyright infringement this adapter is not related to the vendor I'm not allowed to use the website or photos
    OK - accepted
    
2. I increase min jscontroller to 7.0.4
    see below
    
3. I don't need the result or wait for the setState to not block the code I don't use await
    OK - already indiceted at review that this is OK too
    
4. Added replace function for the cleanest part
    OK
    
5. This objects are created via json2iob or for all cleanser part from getType() function
    see below
    
6. Fixed
    OK
   
7. I use my own linter config
   Completly OK as already notes in review
   

The following issues need more / still attention:

- [ ] setObjectNotExists is not synchronous

  setObjectNotExists is an asynchronous function despite its name. Only the callback (not used by your code) is guaranteed to be executed after creation of the object. So please evalute wether ist really save to use setObjectNotExists without a callback. You may use setObjectNotExistsAsync with await or use await together with setObjectNotExists when reuiring js.controller 7.0.4 minimum (https://github.com/ioBroker/ioBroker.js-controller/blob/master/CHANGELOG.md#704-2024-12-04---lucy)

  Please adapt as most likely no other mechanismen than timings will gurantee that objects are really cerated befor first usage.

  Maybe I did not explain clearly. Even with js.controller 7.0.4 setObjectNotExists need an await. So please add await there to make objectCreation synchronous.
  
  Early js-controller versions did not return any useful value from setObjectNotExists but provided the callback as am entrypoint which guranteed that the object was created. New js-controller versions of setObjectNotExists return a promise if used without callback and are therfor identical to setObjectNotExistsAsync. This functionality has been added with js-controller 7.0.4 (see mentioned CHangelog of js.controller). setObjectNotExists is never synchronous by itself but requires the await if you want / must asure that the object is really created. (Of course you could collect all promises and wait at one final place too). 

- [ ] some objects are missing

  all intermediate objects must be created.

  - [ ] dreame.0.XXXXXXXX.general.deviceInfo.defaultVendors
  - [ ] dreame.0.XXXXXXXX.general.deviceInfo.extendScType
  - [ ] dreame.0.XXXXXXXX.map.cleanset.2
  - [ ] possibly more
 
  are not created. Please check / add creation.

  It's ok that hose object are created by json2iob. But there seems to be a bug at this code. At least within the dump provided the mentioned objects are not created. You can check you system and look for the type column of those (and other) objects. This column must not be empty and editing the the object should not result in "object does not exist".
  
  Maybe some ids received frome xternal source do contain a dot and the json2iob code does not eitehr replce the dot with ie. an _ or split the id and create the intermediate (folder type ?) object.

  Heres what I get when loading you json provide earlier:

  Type is empty:
image

  Edit returns:
image

  **Please check / verify and provide a new object json after corrects.**

TA2k:
fixed await for channel creation

Channel creation is fixed in json2iob
[dreame.0 (1).json](https://github.com/user-attachments/files/18565207/dreame.0.1.json)

mcm1957:
Thanks for fixing

mcm1957:

PR #4440, opened 2025-01-04
Add ai-toolbox to latest

ToGe3688:
Please add my adapter ioBroker.ai-toolbox to latest.

### DIFF
--- a/sources-dist.json
+++ b/sources-dist.json
@@ -40,6 +40,11 @@
     "icon": "https://raw.githubusercontent.com/ioBroker/ioBroker.admin/master/packages/admin/admin/admin.png",
     "type": "general"
   },
+  "ai-toolbox": {
+    "meta": "https://raw.githubusercontent.com/ToGe3688/ioBroker.ai-toolbox/main/io-package.json",
+    "icon": "https://raw.githubusercontent.com/ToGe3688/ioBroker.ai-toolbox/main/admin/ai-toolbox.png",
+    "type": "logic"
+  },
   "aio": {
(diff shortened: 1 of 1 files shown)


mcm1957:
reminder 11.1.2025

mcm1957:

Thanks

ToGe3688:
Here is the export of the object structure you requested!
[ai-toolbox.0.json](https://github.com/user-attachments/files/18559419/ai-toolbox.0.json)

mcm1957:
@ToGe3688 

- [ ] object types hirarchy is invalid

  Object types require the general hirarchy device -> channel -> state. Device and channel are optional. The must be at most one object with type device within a tree leave. Multiple channels ('subchannels') are OK. Channels mut be below device if a device is present, no device is allowed below a channel. At every place Objects of type folder can be present. States must not have any children.

  So you either can use folder for top level, device for the models and channel or folder for lower directories or use channel / folder only.

- [ ] reevaluate state roles

  i.e. role indicator requires a readonly boolean state, so it is not allowed for your usage. States containing json Data most likely should use role 'json'

McM1957

Please a new object.json dump after fixing roles / types.

reminder 10.2.2025

ToGe3688:
@mcm1957 Thank you for the feedback! I have worked through your suggested changes and think they should be fixed now. Attached is the new json export from the new version of the adapter.
[ai-toolbox.1.json](https://github.com/user-attachments/files/18589692/ai-toolbox.1.json)

mcm1957:
RE-CHECK!

mcm1957:
@ToGe3688 

PR #4464, opened 2025-01-11
Add oxxify-fan-control to latest

N-b-dy:
Please add my adapter ioBroker.oxxify-fan-control to latest.

### DIFF
--- a/sources-dist.json
+++ b/sources-dist.json
@@ -1861,6 +1861,11 @@
     "icon": "https://raw.githubusercontent.com/iobroker-community-adapters/ioBroker.owntracks/master/admin/owntracks.png",
     "type": "geoposition"
   },
+  "oxxify-fan-control": {
+    "meta": "https://raw.githubusercontent.com/n-b-dy/ioBroker.oxxify-fan-control/master/io-package.json",
+    "icon": "https://raw.githubusercontent.com/n-b-dy/ioBroker.oxxify-fan-control/master/admin/oxxify-fan-control.png",
+    "type": "climate-control"
+  },
   "palazzetti": {
     "meta": "https://raw.githubusercontent.com/inapsis/ioBroker.palazzetti/master/io-package.json",
     "icon": "https://raw.githubusercontent.com/inapsis/ioBroker.palazzetti/master/admin/palazzetti.png",


N-b-dy:
RE-CHECK!

N-b-dy:
RE-CHECK!

mcm1957:
reminder 18.1.2025

N-b-dy:
I'm not sure at all whats wrong with the copyright year, as 2025 is written in the Readme.md. The invitation for bluefox at npmjs.com is already created.

RE-CHECK!

mcm1957:
I'll check copyright during review

N-b-dy:
RE-CHECK!

N-b-dy:
Anything left from my side to do?

mcm1957:
No, please stand by until I hav edone the review.

mcm1957:

Thanks

N-b-dy:
[oxxify-fan-control.json](https://github.com/user-attachments/files/18526034/oxxify-fan-control.json)

The file contains the object tree for two configured fans - hope that fits.

Just for my understanding: Is this process here with "latest" necessary for each version? Or is it during the testing phase fine to create new versions within NPM and it is only necessary to create a new PR to get a specific version added to "stable"?

mcm1957:
> Just for my understanding: Is this process here with "latest" necessary for each version? Or is it during the testing phase fine to create new versions within NPM and it is only necessary to create a new PR to get a specific version added to "stable"?

NO, a review is only done for initial addition to latest. Later releases are automatically added to latest repository as soon as they are released. We believe that devs do not add malicios code on purpose :-) 
Additions to stable must be requested by dev but not checked in detail normally.

mcm1957:
RE-CHECK!

mcm1957:
@N-b-dy 

- [ ] README / copyright year

  Please move the copyright line behin ##license header before gnu text. This should fix the error. (And yes, this should be improved at checker)

- [ ]  do not delete all (!) objects

  Adapter must not delete all objects at every startup. This would also delet all userspecific setups (i.e. history settings) and could cause problems with running scripts. It's OK to remove devices no longer existing at config, but deleting all objects would cause unexpected side effects.

  see https://github.com/N-b-dy/ioBroker.oxxify-fan-control/blob/f1568a9c8dcaa04b8ce86fe11677245cee5796ca/src/main.ts#L61

  ```adapter.FORBIDDEN_CHARS```, i.e. by using the following snippet:

  ```
  function name2id(pName) {
  }
  ```

  StateId for devices is taken from user input. This input must be filtered and illegal character removed. Feel free to either replace invalid characters by i.e. '_' or to check config at startup, issue an error if illegal characters are detected and deny startup. You should add this check to jsonConfig to ('validator' attribute) but checking at code is required too.

- [ ] Object structure / device type

  I would suggest to change type for 'device' to folder and the folder named using deviceId to 'device'. So you would get a state hirarchie folder -> device -> channel ... -> state. Note that device - if present - is NOT allowed below channel and that only a maximum of one 'device' object is allowed at one tree leave.

  But consider this as an suggestion only. It looks useful as deviceId relate to real physical devices (if I'm correct). But its not mandatory. Using channel -> channel -> ... -> state with or wothout folders is acceptable too.

- [ ] consider to subscribe to all states insted aof multiple subscribes

  As you subscribe to a significant number of states, I would suggest to subscribe to all states. This reduces the work js-controller has during parsing every state change (of your adapter) and shifts the load to your adaptercode which anyway already checks which state is updated. This should provide better load balancing as js-controller and your adapter executed within different processes and could use seperated cpu cores.

- [ ] udp server error handling

  If I read the code correctly any error raised by udp server or a simple udp server close only log a message. There seems to be no retry to setup a new udp server or a requste to shutdown the adapter. So (if I'm correct) the adapter will be defunctional after an udp server error until the user issues a restart. This should be improved. You shoudl try to restart the udp server (a limited number of times) and / or shutdown to adapter (and requesting a restart) so that the sysdtem tries to fix any network problem without user intervention.

- [ ] do not execute sendQueue handling every 25ms

  Exeuting the sendQueue handling every 25ms could cause high but unneccessary lod to the system even if no data is to be sent. I would suggest that you i.e. change setIntervall to setTimeout and start a new timeout at the end of processing if more data is available. When adding data to queue you could start processing using setTimeout too. But of course other solutions are ok too, but do not force the adapter to be active (at such a short interval) if no data is processed. 

  see  https://github.com/N-b-dy/ioBroker.oxxify-fan-control/blob/f1568a9c8dcaa04b8ce86fe11677245cee5796ca/src/main.ts#L171

- [ ] intermediate objects must be created

  Some of your objects are not created, i.e.

  - [ ] fan
  - [ ] network
  - [ ] sensors
  - [ ] system
  
  do not have an object. Please ensure that all intermediate objects are created.

- [ ] reevaluate state roles

  Please review all roles according to rules mentioned above. i.e. (incomplete list)
  - [ ] boostModeFollowUpTime -> shouldn't this be a level?
  - [ ] fanSpeedFan1Rpm -> shouldn't this be value ?
  - [ ] ....
  
  Please evaluate all roles and use state only if no better matching role exists.

  pollingIntervall should be limited to reasonable lower and  upper limits. This should be done at jsonConfig AND within code.
 
  
McM1957

reminder 10.2.2025

N-b-dy:
@mcm1957

First of all, I want to thank you for the time and effort you spend to review this adapter and helping me to get a better understanding of the mechanisms, which are used behind the scenes of ioBroker. I've tried to apply all the review points you've suggested, each of them in a separate commit.

Regarding the point with "invalid characters should be filtered from object ids" I've struggled with the validator attribute in the jsonConfig. Can you provide me some links to documentation or example projects, where I can get an idea how this works within a JSON attribute (Regex, string comparison/replacement, ...)?

N-b-dy:
RE-CHECK!

N-b-dy:
RE-CHECK!

mcm1957:
> Regarding the point with "invalid characters should be filtered from object ids" I've struggled with the validator attribute in the jsonConfig. Can you provide me some links to documentation or example projects, where I can get an idea how this works within a JSON attribute (Regex, string comparison/replacement, ...)?

Take a look at https://github.com/mcm4iob/ioBroker.jsonconfig-demo. I hope some validations are done there. ioBroker.snmp also has vlidations - but this was one of my first attempts with jsonConfig so I'm not sure if these are good examples.

mcm1957:
@N-b-dy 

Please attach a new objectDump.json to check the fixes at object structure.
Thanks.

reminder 10.2.2025

N-b-dy:
@mcm1957 Please find attached an updated object structure export.

I've re-read the open points and also updated the input only buttons, so that they are not readable. From my manual tests within the dev-server, the input validation seems to work.

[oxxify-fan-control.json](https://github.com/user-attachments/files/18622658/oxxify-fan-control.json)

mcm1957:
Looks good now.

I've still some remarks (but their no longer blocking)

currentAnalogVoltage seesm to be a voltage value from name and role. But unit % is unusual in this context:

image

Date / Time states typically do not list hh:mm:ss or such as unit, check with other adapters.But thats most likely only an optical and cosmetic decision.
image

mcm1957:
@N-b-dy 

N-b-dy:
> Looks good now.
> 
> I've still some remarks (but their no longer blocking)
> 
> currentAnalogVoltage seesm to be a voltage value from name and role. But unit % is unusual in this context:
> 
> image
> 
> Date / Time states typically do not list hh:mm:ss or such as unit, check with other adapters.But thats most likely only an optical and cosmetic decision. image

Thats right, the analog voltage is a bit confusing. Unfortunately I can't really test this, as I do not have any sensor connected. The interface description in the project says that the unit is “%” and not “V”. The range goes from 0-10V, but the value is reported as percent.

I'll remove the units for date and time, thats fine from my side.

Many thanks for your review and support.

PR #4504, opened 2025-01-25
Add Matter for Beta Start

Apollon77:
Adds Matter Adapter, I will merge myself when we start the Beta

### DIFF
--- a/sources-dist.json
+++ b/sources-dist.json
@@ -1411,6 +1411,11 @@
     "icon": "https://raw.githubusercontent.com/oelison/ioBroker.matrix-org/main/admin/matrix-logo.png",
     "type": "messaging"
   },
+  "matter": {
+    "meta": "https://raw.githubusercontent.com/ioBroker/ioBroker.matter/main/io-package.json",
+    "icon": "https://raw.githubusercontent.com/ioBroker/ioBroker.matter/main/admin/matter.png",
+    "type": "iot-systems"
+  },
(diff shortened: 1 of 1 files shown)


mcm1957:
OK to merge yourself of course :-)

Apollon77:
Notes:
* We decided against Compact mode for now because Matter is really "heavy" code and functionality wise and uses some central functionality on the Node.Js process and we do not know if really anything can be stopped on a shutdown in a way compact mode requires it. Matter also requires a bit more RAM, so the benefits running that inside compact is much lower than for other adapters
* Exact dependency pinning is intentional for matter.js (to ensure 3 dependencies always have the same version), and for the others this will be changed as soon as current issues are solved there
* Testing running special for Matter and also includes a test that the UI gets loaded and such and that's why it uses legacy-testing package intentionally
* singletonHost is intentionally because of the heave usage of special Ports (MDNS and 5540+) where multiple instajces would collide.

mcm1957:
All fine

- Compact mode is already a suggestion only. 
- dependency pinning is notes as in most cases its not necessary and we had troubles at last js-controller update. But I will never consider it blocking.
- testing is to be improved at checker, I'll note it. Please ignore for now
- singleton* is primary as an aid for review. I'll search for a solution to "confirm" it. Please ignore for now.

Apollon77:
RE-CHECK!

PR #4512, opened 2025-01-29
Update bshb to 0.4.1

holomekc:
Please update my adapter ioBroker.bshb to version 0.4.1


mcm1957:
nonstandard tests, manual check required
reminder 31.1.2025

holomekc:
Hi @mcm1957 
I assume this is an automated message. What do I need to do to make it a standard test? I changed the workflow name. The only thing that is different is that I do not release with the workflow and that I use yarn instead of npm. I think at least.

mcm1957:
Well,
You are using yarn and not the "normal" workflow as it is provided by i.e. iobroker.examples based on npm directly.

**But its ok** - the tests themself are executed and thats the important pass. I do not want to force you to use anything you like to do other. Its more importent that you feel well suporteing an ioBroker adapter. I only have to lokk into the workflow runs manually. No problem.

PR #4524, opened 2025-02-01
Bump TeslaFi stable to 1.1.0

hombach:


hombach:
is it usefull that adapter checker isworking against GIT?  It should check the Version in NPM to be released.

mcm1957:
Could be disussed but the primary goal is to check the repository BEFORE a release is deployed. Extending to get data from 2 sources could be an option but is not implemented currently.

PR #4535, opened 2025-02-06
Fixed typo and wrong branch for oxxify-fan-control adapter

N-b-dy:
Casing updated and wrong branch name fixed, which came up from adapter checker after adding to latest within #4464

### DIFF
--- a/sources-dist.json
+++ b/sources-dist.json
@@ -1861,6 +1861,11 @@
     "icon": "https://raw.githubusercontent.com/iobroker-community-adapters/ioBroker.owntracks/master/admin/owntracks.png",
     "type": "geoposition"
   },
+  "oxxify-fan-control": {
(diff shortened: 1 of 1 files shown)


N-b-dy:
Not sure what exactly went wrong? If just synced my fork of this repo, but somehow a lot more adapters are checked here. Do I need to delete my fork and recreate it every time, when I want to create a new PR?

mcm1957:
> Not sure what exactly went wrong? If just synced my fork of this repo, but somehow a lot more adapters are checked here. Do I need to delete my fork and recreate it every time, when I want to create a new PR?

Your PR contains multiple commits including the initial commiz adding the adapter. So it seems that you local fork was not 100% in sync. Maybe your branch was not based on current head of repository or some commits have not been pushed back to official repo.

Normally editing files at local fork can always cause such problems. Best way to update is 

- using iobroker.dev to set a new stable release (not possible in this case)
- using github webeditor to edit sources-dist-*.json directly. This crates a PR bases an a unique branch automatically.

PR #4584, opened 2025-02-22
Add bayernluft to latest

mcm1957:
Please add my adapter ioBroker.bayernluft to latest.

### DIFF
--- a/sources-dist.json
+++ b/sources-dist.json
@@ -195,6 +195,11 @@
     "icon": "https://raw.githubusercontent.com/bembelstemmer/ioBroker.batrium-bms/main/admin/batrium-bms.png",
     "type": "energy"
   },
+  "bayernluft": {
+    "meta": "https://raw.githubusercontent.com/iobroker-community-adapters/ioBroker.bayernluft/main/io-package.json",
+    "icon": "https://raw.githubusercontent.com/iobroker-community-adapters/ioBroker.bayernluft/main/admin/bayernluft.png",
+    "type": "climate-control"
+  },
   "beckhoff": {
     "meta": "https://raw.githubusercontent.com/iobroker-community-adapters/ioBroker.beckhoff/master/io-package.json",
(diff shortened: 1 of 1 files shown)


mcm1957:
@boriswerner @Negalein 
FYI - This PR will be used to trak review and eventually changes required by review.

reminder 29.2.2025

mcm1957:
Object Dump
[bayernluft-v3.0.0-0.json](https://github.com/user-attachments/files/18921663/bayernluft-v3.0.0-0.json)

mcm1957:

```
function name2id(pName) {
}
```

  Base device names are created from data entered by user at adminUI. This data must be filtered so that only valid characters are used to create states. See https://github.com/iobroker-community-adapters/ioBroker.bayernluft/blob/41a4003f6c29ef0f7dfca831713faa8203871f7c/main.js#L413

  Best attempt wouöd be to read config names once and store them at an internal object after filtering / converting the name. This would nsure unique access to objectIds.

  All stateIds must be named using english wording unless the data is directly received from an external source. (Thats ok)
  All names must use either english wording or use i18n multilanguage setup with all supported languges. (to be done) 

- [ ] Why do you create objects only for reachable devices?

  As devices are specified by the user I do not see any reason to delay object creation until a device is reachable. Thios would simplify the code in my oppinion.
  
- [ ]  device table should get an active checkbox

  This would allow disabling a system without removing it if it is out of order for any reason. Without this possibility ioBroker will log an eroor / warning everytime the systems are scanned.

- [ ] avoid calls to getState if not required

  Device status should be stored locally to avoid accessing the ioBroker database only to check if a device is reachable.

- [ ] handling of received data
 
  - some device data is check to be present (i.e. (deviceInfo.temperatureIn !== undefined)) That should be added for ALL data as interfaces could change or transmissionproblems could break assumtions
  - Data not returend (== undefined) should set the state to null / set qs flag instead of keeping an possible incorrect old value. Any comments ?


McM1957

**I will start to fix the obvious points directly and create a PR for easy check / comment**

PR #4611, opened 2025-03-05
REMOVE nibeuplink

sebilm:
Nibeuplink does not exist anymore: https://github.com/sebilm/ioBroker.nibeuplink/issues/303


mcm1957:
Thanks for info and PR

PR #4613, opened 2025-03-06
Add mhi-wfrac to latest

hacki11:


hacki11:
RE-CHECK!

mcm1957:
Please enable issues at your repository

mcm1957:
reminder 13.3.2024

hacki11:
RE-CHECK1

hacki11:
Object Dump [mhi-wfrac.0.json](https://github.com/user-attachments/files/19132576/mhi-wfrac.0.json)
Issues enabled in GitHub repository

hacki11:
RE-CHECK!

hacki11:
Testsreports:
https://forum.iobroker.net/post/1256531
https://forum.iobroker.net/post/1256538

hacki11:
RE-CHECK!

hacki11:
Users are quite happy with the current featue set and stability. Anything I need to do or await to get approved or is it more a capacity topic?
https://forum.iobroker.net/post/1258014

mcm1957:
Review still pending due to missing spare time. Sorry.

mcm1957:
RE-CHECK!

mcm1957:
@hacki11 
And sorry for the long delay - I'm short of free time currently.

- [ ] axios dependencies seems to be missing

  Your code requires axios (https://github.com/hacki11/ioBroker.mhi-wfrac/blob/317ef14e52abe816d0503295b5d7edf380ec7900/main.js#L9). But no admin dependency is included into package.json. Please add. (You adapter works, as most likely any other adapter and the core system install axios anyway. But the dependency should be present at adapters package.json)

- [ ] consider using this.delay

  Adapterclass provides a 'delay' function which allows to let code 'sleep* for a specific time. Private sleep / delay functions (delayInms) are mots likely not needed. But note that this is only a hint and nozt blocking.

  ```adapter.FORBIDDEN_CHARS```, i.e. by using the following snippet:
  ```  
  function name2id(pName) {
  }
  ```

  You retrieve the airconId from the device identified by its IP. Please ensure that the returned airconId does not contain any invlid characters. This could happen for newer firmware / newer devices or by simply an incorrect transmission. See https://github.com/hacki11/ioBroker.mhi-wfrac/blob/317ef14e52abe816d0503295b5d7edf380ec7900/main.js#L651


- [ ] reevaluate state roles

  Some small issues:
  - account is a number but role indoicator requires a boolean. So role should be value most likely
  - same applies to several other states (i.e. apMode, autoHeating, ...) - change type to boolean or role whatever applies better.
  
- [ ] Apply units where appropiate

  Several states provide values which typically hav a unit associated, i.e. °C for temperaturs. I suggest to add those units. (but will not consoider it blocking)

McM1957

**Please add a new objects export after fixing roles.**

reminder 13.4.2025

hacki11:
- [x] axios dependencies seems to be missing => added to package.json
- [x] consider using this.delay => eliminated use of delays
- [x] invalid characters should be filtered from object ids => airconId for using as channel sanitized
- [x] reevaluate state roles
- [x] Apply units where appropiate => all values with units had them already
- [x] consider using adapter.setTimeout / this.setTimeout instead of (standard) setTimeout => used this.setX
- [x] check and limit configurable timeouts / intervals => N/A as using this.setX now
[mhi-wfrac.json](https://github.com/user-attachments/files/19526119/mhi-wfrac.json)

mcm1957:
RE-CHECK!

mcm1957:
I still suggest to check polling intervall limits. A value of 0 could put a big load onto the ioBroker system an potenzially overlaod the airconditioner interface, But as it is a timeout (and not an interval) I do not consider it blocking. Take it as a suggestion only.

Else lgtm.

mcm1957:

hacki11:
> I still suggest to check polling intervall limits. A value of 0 could put a big load onto the ioBroker system an potenzially overlaod the airconditioner interface, But as it is a timeout (and not an interval) I do not consider it blocking. Take it as a suggestion only.
> 
> Else lgtm.

I now check and replace for None (replace with 20s), lower 5s (replace with 5s), higher 3600s (replace with 3600s). Interval is also validated due to jsonConfig min/max but I added it as a fallback and replace it automatically for a good user experience. Thanks for your work!

PR #4614, opened 2025-03-06
Add skiinfo to latest

oweitman:
Please add my adapter ioBroker.skiinfo to latest.

### DIFF
--- a/sources-dist.json
+++ b/sources-dist.json
@@ -2331,6 +2331,11 @@
     "icon": "https://raw.githubusercontent.com/ioBroker/ioBroker.simple-api/master/admin/simple-api.png",
     "type": "communication"
   },
+  "skiinfo": {
+    "meta": "https://raw.githubusercontent.com/oweitman/ioBroker.skiinfo/main/io-package.json",
+    "icon": "https://raw.githubusercontent.com/oweitman/ioBroker.skiinfo/main/admin/skiinfo.png",
+    "type": "misc-data"
+  },
   "slideshow": {
     "meta": "https://raw.githubusercontent.com/gaudes/ioBroker.slideshow/main/io-package.json",
     "icon": "https://raw.githubusercontent.com/gaudes/ioBroker.slideshow/main/admin/slideshow.png",


mcm1957:
reminder 13.3.2024

oweitman:
[skiinfo.0.json](https://github.com/user-attachments/files/19126698/skiinfo.0.json)

mcm1957:
@oweitman 

- [ ] object dump does only contain object "config"

  Either your adapter does not create more than this single object - then subscribing to any state is not needed as the config object is not writeable and cannot (should not) be written to by users. Or objects are only created when selecting skiing areas. In this case please select some skiing area and provide an object dump containing realistic amount of states.

Image

  Will be checked with realistic object dump if more states than the config state exist. Please ignore for now.

- [ ] reevaluate state roles
  
  Will be checked with realistic object dump if more states than the config state exist. Please ignore for now.

  Will be checked with realistic object dump if more states than the config state exist. Please ignore for now.

- [ ] onStateChange handler calls `skiinfoserver.doStateChange(id, state);`
  https://github.com/oweitman/ioBroker.skiinfo/blob/bb70cf3b2c16e87fff07fa9f087af643d46b4b2f/main.js#L75
  But I could not detect this method at skiinfoserver class. Please advise.

- [ ] consider using ioBroker delay

  Iobroker adapter class provides the function delay to "sleep" for  aspecific amount of time. So a provate sleep / delay should not be necessary. This is a hint only and is of course not considered blocking.

- [ ] check and adapt testing
  
  Please use standard iobroker test-and-release.yml workflow like 
https://github.com/ioBroker/ioBroker.example/blob/master/JavaScript/.github/workflows/test-and-release.yml
  
  
McM1957

reminder 13.4.2025

oweitman:
- object dump does only contain object "config"
added with more content. but it only contains the favorite array
- onStateChange handler ignores ack flag
- onStateChange handler calls skiinfoserver.doStateChange(id, state);
function removed
- Add timeout to axios calls
timeout added
- consider using ioBroker delay
functions switched
- check and adapt testing
its here
https://github.com/oweitman/ioBroker.skiinfo/blob/main/.github/workflows/test-and-release.yml

oweitman:
new [skiinfo.json](https://github.com/user-attachments/files/19527877/skiinfo.json)

mcm1957:
RE-CHECK!

mcm1957:

PR #4639, opened 2025-03-14
add plenticore-g3 to latest

FernetMenta:


FernetMenta:
Here is the object structure as requested (my prod instance at home). Note that this structure depends on the optional settings and process data selected by the user. The states of processdata-available and settings-available hold all possible IDs queried by the adapter after having connected to the device. Only a very small number of data oints are preselected.

[plenticore-g3.0.json](https://github.com/user-attachments/files/19251027/plenticore-g3.0.json)

mcm1957:
@FernetMenta 

  ```
  function name2id(pName) {
  }
  ```

  As far as I could see objectIds are created based on data read from the external device. But I could not see any explizit filtering (besides replacing ':' by '.') at one place.

  Please filter external data before using it as stateIds. Filtering based on FORBIDDEN_CHARS is the required minimum. But in addition the best solution is to limit stateIds to [A-Za-z0-9-_] Especially spaces are known to create problems with other adapters and VISes. But also & annd other charactes having special meaning i.e. at html are critical.

  And please note that the dot (.) has a special meaning. It seperates objects as a hirarchie. But you must create objects for every hirarchy seperatly and explicitly.

- [ ] some objects are missing / not created

   There is not object created i.e. for 'sbc', 'sbc.statistic', 'processdata', 'processdata.devices' and eventually others. Th column 'type' must not be empty. Every level of the object hirarchie must be created. A valid structure is  device -> channel -> state. There must not be more than one device withon a leaf , but device can be missing totally too. device must not be below channel. States must not have children. Folders can be places anywhere in between.

Image

- [ ] reevaluate state roles

  As you seem o know at least some states reported by the adapter (as you provide a description an dunit) please use the best matching role for known states.

  It seems that some labels are not translated into all supported languages, i.e. 
  - 'Allgemeine Einstellungen' -> 'Language'
  - some 'Beschreibungen'
  
Image

Image

  But also check lower limit. A timeout of i.e. 0 will put a high load at ioBroker host and continuos polling could overload the inverter too.

  
McM1957

**Please attach a new object dump after fixing hirarchie and ev. roles**

reminder 13.4.2025

FernetMenta:
@mcm1957 Thank you very much for the review. At the moment I am looking for missing translations. Can you help me with that one?

> some 'Beschreibungen'

What are you referring to?

mcm1957:
example at screenshot - 'Power of AC side of inverter' is english althou langauge for admin UI is german

FernetMenta:
> example at screenshot - 'Power of AC side of inverter' is english althou langauge for admin UI is german

The screenshots are not visible to me. I get 404 when I click on them.

mcm1957:
Never heard that Github block some pictures ... 

Anyway: I refer to Instance config (adminUI). At tab "Prozessdaten" some states / data-points have a description. These descriptions are not translated yet. Exmple 'Power of AC side of inverter'

In general a description should be added to all states / data-point which you know their purpose

FernetMenta:
Stange, the datapoint you are referring to is translated here. 
image

Did you try latest version 0.1.0?

FernetMenta:
> In general a description should be added to all states / data-point which you know their purpose

Well, the idea is to have only a very small number of predefined datapoints to query and leave the optionals to the user. Did you notice that users can add their own descriptions to the datapoints they select in addition the the presets?

mcm1957:
Well the idea that users have small list of preactivated / recommended datapoints is fully ok. But I would have suppoerted the user with a description so that he cn more easily check what to select. But it's up to you - I will not consider the empty descriptions as blocking.

Do you take this description form teh state description?

It looks like you do not use an i18n arra for state descriptiosn too. So the state description is not adapting to selected UI language.
See   description = this.#I18n.translate(found.description);

You should eventually use I18n.getTranslatedObject which returns an multilanguage object and enebales admin to show the text depending on the selected lanugare at run time (and not fixed to language at installation time).

FernetMenta:
> Do you take this description form teh state description?

Yes, I do. Changing the language triggers a restart of the adapter, which will update the translations in the JSON used by the admin UI. Works here but you need a Kostal Inverter of course. The descriptions are updated after a couple of seconds the adapter needs to restart.

> Empty fields can be edited. But I cannot change existing descriptions. Is this desired this way?

Yes, there are appproximately 400 data points and I wanted to give users the chance to document the data points they intend to use as optional. Maybe some future version can provide a kind of description repo where those can be shared between users.

FernetMenta:
Requested fixes and changes by review:

- fix object hierarchy [38ba8022ad84be71bb00a75ac107f21c4f0510d3](https://github.com/FernetMenta/ioBroker.plenticore-g3/commit/38ba8022ad84be71bb00a75ac107f21c4f0510d3)
  addresses "invalid characters should be filtered from object ids" and "some objects are missing / not created"
  Now the api endpoints "processdata" and "settings" map to channels. The module IDs map to folders and the data points to states. Only [A-Za-z0-9-_] are allowd in IDs.

- fix state roles: [dbd06dd30231acae5226b3419dfb9915151d10db](https://github.com/FernetMenta/ioBroker.plenticore-g3/commit/dbd06dd30231acae5226b3419dfb9915151d10db)
addresses  "reevaluate state roles"
Preset data points have a best matching role. For optional processdata there is a function that tries to guess the role from the name: for example data id that ends with "_P" will get "value.power" as role.

- fix missing translation: [d6cd20902cb7f39bdab8b956543d8d9a57021333](https://github.com/FernetMenta/ioBroker.plenticore-g3/commit/d6cd20902cb7f39bdab8b956543d8d9a57021333)

- fix setTimeout and check user input: [541145453ff3fcd3ce9a9bfadecb5ccce0d41cd8](https://github.com/FernetMenta/ioBroker.plenticore-g3/commit/541145453ff3fcd3ce9a9bfadecb5ccce0d41cd8)
addresses "consider using adapter.setTimeout / this.setTimeout instead of (standard) setTimeout" and "check and limit configurable timeouts / intervals"

New object dump:
[plenticore-g3.json](https://github.com/user-attachments/files/19551239/plenticore-g3.json)

mcm1957:
Refering to https://github.com/ioBroker/ioBroker.repositories/pull/4639#issuecomment-2768859049:

Changing the language at system settings does NOT trigger an restart of the adapter. Only changing the languge at adapter setting will restart the adapter. Admin UI languiage should be controlled by adminUI (= system language) and not by any extra setting in adminUI. So if you do note have any language use for i.e. vis elements or messages displayed at the device I even would suggest to remove the langauge field from the configuration (but that is a suggestion only as long as the adapter honours the system languge for configuration if this is selected).

Normally an adapter restart would not change descritions too as setObjectNotExists will NOT update descriptions.

But I just noticed that you seem to use setObject during EVERY start of the adapter. **Please replace setObject by setObjectNotExists or extendObject.** setObject should not (or better must not) be used on a regular base as setObject

- always creates noew objects and thus deletes any userspecific setting (i.e. history settings)
- resets i.e. name changes performed by users at object. Use option '{ preserve: { common: ['name'] } },' with extendObject to keep name changes.

Usage of setObject is disouraged expect for special cases where you really want to delet all existing config of the state.

Anyway - I will review the fixes already done and summarize if still something is open.

FernetMenta:
@mcm1957 With setObjectNotExists the descriptions of the states stay at the language set at the time they were created, right? Is this intended behaviour?
What is best practice for the case that a translation is required by object descriptions and by the admin ui?

FernetMenta:
> I18n returns such an object using I18n.getTranslatedObject function. So a simple replace of I18n.translate with I18n.getTranslatedObject should do the trick. (Description of function see adapter-core).

Ahh, nice. I will look into this. I have no strong feelings about the language parameter. I saw it in some other adapter project I learned from and thought it is a good idea. Probably better to drop it.

FernetMenta:
> Best practise would be to use language object for name, description, ... (thinks using i18n multilanguage support)
> 
> `description: { en: 'This is a state', de: 'Das ist ein Zustand', .... },`
> 
> I18n returns such an object using I18n.getTranslatedObject function. So a simple replace of I18n.translate with I18n.getTranslatedObject should do the trick. (Description of function see adapter-core).
> 

I updated post: https://github.com/ioBroker/ioBroker.repositories/pull/4639#issuecomment-2769294620 with new fix based on your suggestion

FernetMenta:
> Please replace setObject by setObjectNotExists or extendObject. setObject should not (or better must not) be used on a regular base as setObject

done by https://github.com/FernetMenta/ioBroker.plenticore-g3/commit/f163f13d5c544dfe93132f922be61159ef3165f2

mcm1957:
lgtm

mcm1957:

PR #4644, opened 2025-03-16
Add dnscope to latest

simatec:


simatec:
RE-CHECK!

mcm1957:
@simatec 
Do you really think its ready :-)

image

(But I will not have time the next days anyway - sorry)

simatec:
> @simatec Do you really think its ready :-)
> 
> image
> 
> (But I will not have time the next days anyway - sorry)
> 
> reminder 23.3.2025

Readme added

simatec:
RE-CHECK!

mcm1957:
@simatec 

- [ ] reevaluate state roles

  data.currentIpV4 must not have role indicator as indicator requires type boolean. I suggest value or info.ip.
https://github.com/simatec/ioBroker.dnscope/blob/096754fa00f94daa5dbc393159b7939b3d2f2a92/src/main.ts#L73

- [ ] eventually use delay 

  adapter class provides 'delay' for holding processing for a specified typ. This could be used insteat of sleep in many cases.

  see https://github.com/ioBroker/ioBroker.js-controller/blob/bc4ae0b67453a8e2b9b351c12720bcaafad99e9b/packages/adapter/src/lib/adapter/adapter.ts#L2737

- [ ] scheduled adapters should have random delays
  
- [ ] do not access cloud services at fixed timestamps
  
McM1957

reminder 5.4.2025

simatec:
@mcm1957 all fixed

mcm1957:
spreading cloud access seems to be still open. I've created Issue https://github.com/simatec/ioBroker.dnscope/issues/16 to track it.

Otherwise lgtm.

mcm1957:

PR #4649, opened 2025-03-18
Update wireless-mbus stable to 0.10.0

lvogt:


mcm1957:
@simatec 
As https://github.com/lvogt/ioBroker.wireless-mbus/issues/228 is still open, could you check if responsive design is acceptable for stable update or need an prior update?

PR #4672, opened 2025-03-30
Update sainlogic to 0.12.2

phifogg:
Please update my adapter ioBroker.sainlogic to version 0.12.2.


mcm1957:
In addition standard github tests must pass. They are currently failing as your package.json decelares a dependency admin:^7.4.10. Admin is no npm pacakge and hence must not be included in dependencies at package(!).json.
 
image

PR #4683, opened 2025-04-02
Add wiim to latest

KaiIOB:
Please add my adapter ioBroker.wiim to latest.


mcm1957:
reminder 9.4.2025

KaiIOB:
Here comes the object structure:

[wiim.0.json](https://github.com/user-attachments/files/19619877/wiim.0.json)

mcm1957:
@KaiIOB 

- [ ] Readme.md changelog

  Please rmove '<' from changelog headers. It sould read '###' to indicate H3 headers.
    
  ```
  function name2id(pName) {
  }
  ```

  You retriev the servername (first level object) from the bonjour esponce. This might conatin invalid characters and hence must be filtered. At least FORBIDDEN_CHARS maust be removed. Spaces should be removed in additon to. Optimal solution is to allow only characters [0-9A-za-z_-]. Dots must be handled seperatly as they are used as seperator between levels. 

  Sepcials chars can cause problesm with several VIS implementations (i.e. space is know to break VISes).

- [ ] Please minimize non-debug logging

  Please minimize logging to such info that a normal user will use. Feel free to log as much information as useful at DEBUG level. And please remove ornaments ect (like '******* xxx*****') and reduce to pure information. See https://github.com/KaiIOB/ioBroker.wiim/blob/eaaf2c23cd8017e9e45a4ff89495ac3bdc3337c1/wiim.js#L35

- [ ] do not log errors as info

  Errors shoudl be logged as error or wanring (depending ob effect on opertion. If an error is normal and does not effect normal use, log it as debug. see ie. https://github.com/KaiIOB/ioBroker.wiim/blob/eaaf2c23cd8017e9e45a4ff89495ac3bdc3337c1/wiim.js#L86
- [ ] log messages should / must be in english

  As ioBroker is an international software, please use only english wording for all logging messages - unless you translate logging depending on system language. Logging using the system langugae is ok of course - buts its your decision if you want toe invest the effort. German only logs are not desired. Thios refers to all text defined by you - error messages reurned from aservice (external text) can be kepogt as is of course.

- [ ] all objects must be created

  You do not cretate intermediate objects. i.e. 'wintergarten' does not have a type assigned. Object hiorarchie must be
'device' -> 'channel' -> 'state'. Device and / or channel might be omitted. chennels might have subchannels if required. device is NOT allowed below channel. States must not have any children. In addition 'folders' are allowed at every level.

- [ ] reevaluate state roles

  - role indoicator requires a state of type boolean
  
- [ ] check r/w of states

  Are you sure that tracklength is a writeable state?

- [ ] remove unused admin files

  You adapter uses jsonConfig.json. So words.js seems to be outdated. I did not see any configuration for custom attributes or admin tab. So custom_m.html and tab_m.html seem to be obsolete too.

  wim.PNG is (most likely) not used t0o ans should be removed.

- [ ] remove unused config paramaters from io-package.json

  IP_Address and Request-type are not used at configuration. So they should be removed from io-package.json. See   https://github.com/KaiIOB/ioBroker.wiim/blob/eaaf2c23cd8017e9e45a4ff89495ac3bdc3337c1/io-package.json#L87

- [ ] remove unused dependencies

  tcp-ping and xml2js are listed at dependencies but there is no require of them insoide code. So this dependencies seem to be obsolete. Please remove.
 
- [ ] use setTimeout or delay instead of one-time intervals

  During startup you use ruther complicated setInterval solutions which seem to do only one thing - delay operation. Please eiterh use setTimout or (more likely to do the jiob there) use this.delay so delay operation as desired. In addtio  it's at least unusal to loop until some state is reached (i.e. bonjourFinished). You can simply call the next routine at the point whre you set bonjourFinished - either directly or by using a setTimout. 
  
- [ ]  general remark

  Some remarks:
  - The adapter seems to scan for devices only during startup. Whats the behavior when some device gets switched off while the adapter is running? The adpater should NOT liog errors over an over while the device is sitched off.

  - How does the adapter recognize a devide sitched off during adapter startup? The adapetr should - in my oppinion - scan from tie to time to detect new devices. It's not hande for users if the must restart the adapter whenever a new device is switched on. Consider running bounjour descovery at background .
  
McM1957

** Please attach a new object dump whith corrected object creation and roles **

reminder 2.5.2025

mcm1957:
@KaiIOB 
Did you have time to review the issues and fix them?

reminder 8.5.2025

mcm1957:
@KaiIOB
Did you have time to review the issues and fix them?

I do not have time to review before 26.5.2025 but please drop a note until then stating if an within what timeframe you plan to process the issues.

reminder 26.5.2025

KaiIOB:
Hello Martin,
thank you for the review!
Sorry for not getting back to you earlier, work kept me busy.
To be honest, as there doesn't seem to be any "customers" for the adapter, I think it does not make any sense to put more work into it. I'm pretty sure you have other adapters to work on which are  much more interesting for the community, right?
Rgds
Kai

16.05.2025 18:33:46 Martin M. ***@***.***>:

mcm1957:
reminder none

mcm1957:
RE-CHECK!

PR #4749, opened 2025-04-27
Update mhi-wfrac to 2.1.3

hacki11:
Please update my adapter ioBroker.mhi-wfrac to version 2.1.3.

*This pull request was created by https://www.iobroker.dev 20f0116.*

### DIFF
--- a/sources-dist-stable.json
+++ b/sources-dist-stable.json
@@ -1629,6 +1629,12 @@
     "type": "weather",
     "version": "3.0.3"
   },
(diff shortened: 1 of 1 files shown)


hacki11:
I guess, the adapter checker is just using the latest release as it shows 2.1.4. But I want to use 2.1.3 here. The commit is correct, just what the adapter checker writes is wrong. 2.1.4 is too new to get into stable.

PR #4786, opened 2025-05-10
Add onlycat to latest

Sickboy78:
Please add my adapter ioBroker.onlycat to latest.

*This pull request was created by https://www.iobroker.dev 20f0116.*

### DIFF
--- a/sources-dist.json
+++ b/sources-dist.json
@@ -1841,6 +1841,11 @@
     "icon": "https://raw.githubusercontent.com/ioBroker/ioBroker.onkyo/master/admin/onkyo.png",
     "type": "multimedia"
   },
+  "onlycat": {
+    "meta": "https://raw.githubusercontent.com/Sickboy78/ioBroker.onlycat/main/io-package.json",
+    "icon": "https://raw.githubusercontent.com/Sickboy78/ioBroker.onlycat/main/admin/onlycat.png",
+    "type": "iot-systems"
+  },
   "onvif": {
     "meta": "https://raw.githubusercontent.com/iobroker-community-adapters/ioBroker.onvif/main/io-package.json",
     "icon": "https://raw.githubusercontent.com/iobroker-community-adapters/ioBroker.onvif/main/admin/onvif.png",


Sickboy78:
[onlycat.0.json](https://github.com/user-attachments/files/20140825/onlycat.0.json)

mcm1957:
reminder 17.5.2025

mcm1957:
Sorry, due to lack of time and holidays I will not be able to process this before 26.5.2025.

Sickboy78:
@mcm1957 
Thanks for taking the time to review!

- review info logging -> I reduced most of the info logs to debug and some debug to silly.
- invalid characters should be filtered from object ids -> I already do, see https://github.com/Sickboy78/ioBroker.onlycat/blob/a82322f8a8705f21d3082b04da59127305d9ca1f/main.js#L1279. I use /\W/ig which is [^a-zA-Z0-9_].
- promisifictaion of setObjectNotExists -> Old habit, replaced it with setObjectNotExistsAsync.
- some labels seem to have no translation into all languages -> done
- consider using adapter.setTimeout -> done
- Please add node24.x to testing matrix -> done

Best Regards

mcm1957:
**INTERNAL COPY for review purposes only - Please ignore**

- [ ] review info logging

  Please review logs of level info. Those should provide real information to users. Many are fully ok, but several other seem to be better places at debug level (i.e. event received, updateing events, nothing changed, nothing to do etc.) Adapter should not fill log. Just review which info logs are really useful for users and which are more like of debugging. Users will be happy if log is not too big. [non blocking - users will repond anyway if too much noise]

```
function name2id(pName) {
}
```
  You retrieve the pet names from external sources. Please verify that pet names used as Objetc-IDs are filtered. Best attemt ois to restrict to A-Za-z0-9_-, but filtering using FORBIDDENS_CHARS is ok if you remove spaces and dots too.

- [ ] promisifictaion of setObjectNotExists
  https://github.com/Sickboy78/ioBroker.onlycat/blob/3c4832ab0a0c4e6273fbb0820ef94cb8b9d392ff/main.js#L83
  Why do you use this construct? steObjectNotExistsAsync is already an asynchronose / promise returning version.
  https://github.com/ioBroker/ioBroker.js-controller/blob/6ade4045339082a9408e17e0871ad26dc41206ca/packages/adapter/src/lib/adapter/adapter.ts#L1059

  It seems that some labels are not translated into all supported languages

  Please add node24.x to testing matrix.

  Tests for node 22 are mandatory as this is the current LTS node version. 
  
  So the recommended testmatrix is [20.x, 22.x,24.x] depending on engines requirments setting at package.json 

mcm1957:
RE-CHECK!

PR #4824, opened 2025-05-22
Add wbec to latest

jb-io:
Please add my adapter ioBroker.wbec to latest.

### DIFF
--- a/sources-dist.json
+++ b/sources-dist.json
@@ -3090,6 +3090,11 @@
     "icon": "https://raw.githubusercontent.com/theknut/ioBroker.waterkotte-easycon/main/admin/waterkotte-easycon.png",
     "type": "climate-control"
   },
+  "wbec": {
+    "meta": "https://raw.githubusercontent.com/jb-io/ioBroker.wbec/master/io-package.json",
+    "icon": "https://raw.githubusercontent.com/jb-io/ioBroker.wbec/master/admin/wbec.png",
+    "type": "iot-systems"
+  },
   "weather-warnings": {
     "meta": "https://raw.githubusercontent.com/ticaki/ioBroker.weather-warnings/main/io-package.json",
     "icon": "https://raw.githubusercontent.com/ticaki/ioBroker.weather-warnings/main/admin/weather-warnings.png",


mcm1957:
reminder 30.5.2025

jb-io:
Here the missing obj-json

[wbec.1.json](https://github.com/user-attachments/files/20432038/wbec.1.json)

jb-io:
Just released a new version with translated states:
[wbec.0.json](https://github.com/user-attachments/files/20432559/wbec.0.json)

mcm1957:

SORRY FOR THE VERY LONG PROCESSING TIME

- [ ] invalid email

  Please add an email address which allows you to connect. We cannot reach out to you using a "no-reply" github address in case some contact is required. 
 
```
function name2id(pName) {
}
```
  
  ObjectIds seem to be based on response from wbed.cleint api which in return seems to read informatuon from wallbos / device. So this ids must be filtered using FORBIDDEN_CHARS at least. Best praxis is to limit ids to A-Za-z0-9_- but filteriung FORBIDDEN_CHARS is sufficient if you remove dot and spaces too.

- [ ] reevaluate state roles and object types

  - Your 'directories' cgf, info, modbus ect. (except boxn) do not really correlate to some devide. So I would suggest to change the type to channel or folder
  - some states do not ahve a valid role assigned, i.e. box0.busid, cfg.*
  - indicators MUST be boolean
  - writeable states MUST NOT have a value role, use LEVEL in this case (i.e currLim)
  
  Please add node.js 24 tests

  Tests for node 22 are mandatory as this is the current LTS node version. 
  
McM1957

**Please add a new object export too***

reminder 18.6.2025

mcm1957:
@jb-io 
Did you notice my review comments?

mcm1957:
@jb-io 
I've dropped some review comments at Jun 4. Please provide some feedback if you already processed / fixed those remarks so that a re-review should be started.

If you did not yet have time to review / process please let me know some timeframe when to expect this to be done.

Thanks

mcm1957:
reminder 20.7.2025

mcm1957:
@jb-io
**Just a kind reminder**

Please comment to let me know if you still plan to finish this adapter.

**If there is no reaction at all (not even some 'iI'll will process after x.y.z') until 21.8.2025 this PR will be closed.**

reminder 13.8.2025

PR #4898, opened 2025-06-29
Add pi-hole2 to latest

oweitman:
Please add my adapter ioBroker.pi-hole2 to latest.

### DIFF
--- a/sources-dist.json
+++ b/sources-dist.json
@@ -1986,6 +1986,11 @@
     "icon": "https://raw.githubusercontent.com/unltdnetworx/ioBroker.pi-hole/master/admin/pi-hole.png",
     "type": "network"
   },
+  "pi-hole2": {
+    "meta": "https://raw.githubusercontent.com/oweitman/ioBroker.pi-hole2/main/io-package.json",
+    "icon": "https://raw.githubusercontent.com/oweitman/ioBroker.pi-hole2/main/admin/pi-hole2.png",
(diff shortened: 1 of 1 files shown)


mcm1957:
reminder 15.7.2025

mcm1957:
RE-CHECK!

mcm1957:
@oweitman 

- [ ] onUnload does not contain code

  All existing connections and especially all running timers should be closed / stopped during adapter shutdown. At a firszt look, somehow ioUtils.closeConnections should be called - but I did not dig into details. Please check.

- [ ] all Objects must be created

  Object 'Data' seems to be used withput explicit creation. No type for this obejct is defined - most likely it should be a 'folder')

  Image

  Please check / correct.

- [ ] reevaluate state roles

  Role 'value' is used at almost all states but requires read only states with type number. Please use i.e.

  - json common.type = string for states with json content
  - level for writeable states containung numbers
  - text for states containing strings
  - ...

  or any role matching better.
    
  Minimum and maximum values for refresh timers should be set at jsonConfig (a negative value does not make much sense i.e.) but checks should be done at (startup) code too as paramaters could be set outside of jsonConfig too

  
McM1957

**please attach a new object dump after fixing roles**

oweitman:
[pi-hole2.json](https://github.com/user-attachments/files/21245598/pi-hole2.json)

mcm1957:

PR #4906, opened 2025-07-02
Add weathersense to latest

ltspicer:
Please add my adapter ioBroker.weathersense to latest.

### DIFF
--- a/sources-dist.json
+++ b/sources-dist.json
@@ -3105,6 +3105,11 @@
     "icon": "https://raw.githubusercontent.com/woessmich/ioBroker.weatherflow_udp/master/admin/weatherflow_udp.png",
     "type": "weather"
   },
+  "weathersense": {
+    "meta": "https://raw.githubusercontent.com/ltspicer/ioBroker.weathersense/main/io-package.json",
+    "icon": "https://raw.githubusercontent.com/ltspicer/ioBroker.weathersense/main/admin/weathersense.png",
+    "type": "metering"
+  },
   "weatherunderground": {
     "meta": "https://raw.githubusercontent.com/iobroker-community-adapters/ioBroker.weatherunderground/master/io-package.json",
     "icon": "https://raw.githubusercontent.com/iobroker-community-adapters/ioBroker.weatherunderground/master/admin/wu.png",


mcm1957:
reminder 15.7.2025

ltspicer:
RE-CHECK!

mcm1957:

- [ ] consider using '@iobroker/eslint-config'

  Please sonsider using standard iobroker eslint config. While this is not absolutly required and can be finally decided by the dev its highly recommended so that standard checks / formats within ioBroker adapters are as common as applicable.same name.

  (This issue will not be considered blocking)


- [ ] do not write to /home/iobroker

  Adapter should not write into iobroker driectories except the adapter specific data directory or some subdirectory of it (iobroker-data/<adaptername>). Please use routines provided to retrieve the name of the adapter data directory.
  If the adapter really should be able to write to any directory, please do not default to the iobroker user home directoy (/home/iobroker).


- [ ] really use hardcoded ip?
  https://github.com/ltspicer/ioBroker.weathersense/blob/6165d454135709803ab3dd789d206aa39299be00/main.js#L339
  Normally domain names should be used instead of fixed IPs as IPs might change in future due to changes / location of manufacturer cloud.


- [ ] all objects must be created

  Top level objects ('1', sensor) are not explicitly created. Most likely these should be of type device.

  Image

- [ ] reevaluate state roles

  Several states containing strings are of type 'value'. But 'value' requires the object type 'number'. https://github.com/ioBroker/ioBroker.docs/blob/master/docs/en/dev/stateroles.md#values-numbers-read-only. Most ikely 'test' or'date' will be the correct role.

- [ ] label from heizoel24-max 
  
  https://github.com/ltspicer/ioBroker.weathersense/blob/6165d454135709803ab3dd789d206aa39299be00/admin/i18n/en/translations.json#L2
  
    
McM1957

**Please attach a new object dump after fixing creation / roles.**

reminder 31.7.2024

ltspicer:
Object structure:
[weathersense.0.json](https://github.com/user-attachments/files/21300746/weathersense.0.json)

mcm1957:

PR #4909, opened 2025-07-02
Update shelly to 10.0.0

mcm1957:
Please update my adapter ioBroker.shelly to version 10.0.0.


klein0r:
We should set 10.1.0 as stable ASAP (fixes in BLE Script and Encryption Key handling). See #4923

PR #4916, opened 2025-07-08
Add openmediavault to latest

Scrounger:


Scrounger:
RE-CHECK!

Scrounger:
> This should ensure that the structure is visible, but no sensitive content is exported.

Excluding the states is not the solution, object structure it self can contain sensitive data (e.g. mac's in id).
The hierarchy is predefined according to the dev docs channel -> device -> channel. All roles are set to states.
More details can be found through code review.

Sorry but I don't have the time to do the effort to bring the structure into a clean structure.
It is your decision whether the adapter should go into the repository or not.

mcm1957:
reminder none

@Apollon77 please make the review if you have time

Scrounger:
"RE-CHECK!"

Scrounger:
@Apollon77: here also i added a function to auto assign roles - as for the freeair adapter:
https://github.com/Scrounger/ioBroker.openmediavault/blob/11670b31e6e91d60635bdfe68595020525fcf024/src/lib/myIob.ts#L457

both adapter shares now the same file for all the object stuff of iob

mcm1957:
@Apollon77 **just a kind reminder.**
Please make the review for this adapter and merge if OK.
As this is no "0815" adpter and I cannot determine if it matches guidelines as it is or which adaptions you would suggest.
**So a review by you is required.**

mcm1957:
Please discuss with Apollon77. Review of this Adapter (and eventually release if he is satisfied) has been assigned to him.
I do not release an adapter which has not been reviewed by me.

Scrounger:
@mcm1957 

@Apollon77 has done a review above (https://github.com/ioBroker/ioBroker.repositories/pull/4916#issuecomment-3310939410). The points of criticism were the unnecessary roles, which I have implemented. Let's see if something will come of it this year...

mcm1957:
As we could not find a solution at the first discussions (i.e. you still think roles are not necessary) - I stopped review on this adapter and passed it to Apollon77. So he is your contact now. He can decide how to process, accept the adapter as it is or request changes and can release it to repositories.

Scrounger:
@mcm1957 

> As we could not find a solution at the first discussions (i.e. you still think roles are not necessary)

Now it's getting really ridiculous.
As I already wrote, the roles are unimportant to me, that's my opinion and I'm not changing it. 
You have a different one, which is perfectly fine....

We've been messing around here for almost 4 months.
This is a well-written adapter that meets all the requirements according to the documentation, including the requirements for roles, because, quote: “if you don't know which role the state has, use this one,” which leaves you with the option of using ‘state’ as a role.
I have implemented all the points from your review and after reviewing @apollon77 comment, I decided to implement the roles after all.

So what's the problem?

PR #4926, opened 2025-07-13
Update wireless-mbus stable to 0.10.1

lvogt:


mcm1957:
Please fix
 ❗ [E160] "admin" is not allowed in common.dependencies
and create a new release.

admin must be listed at globalDependencies. See https://github.com/iobroker-community-adapters/ioBroker.shelly/blob/bf44ddf812804cbdbcc32790a8423a6f68d73fad/io-package.json#L185 as an example.

This should fix E190 too

PR #4938, opened 2025-07-17
Update alexa2 to 3.27.3

Apollon77:
Please fast lane because of Amazon changes. Was tested in Forum and GitHUb


mcm1957:
@Apollon77 
I assume fast lane is desired :-)

PR #4979, opened 2025-08-04
Add xsense to latest

arteck:


mcm1957:
RE-CHECK!

arteck:
@mcm1957  check again

mcm1957:
RE-CHECK!

mcm1957:
RE-REVIEW 

- [ ] errors reported by adapter checker must be fixed

  Please fix errors and (as far as possible) warnings reported by adapter checker.

 - ❗ [E140] width and height of logo (specified at "common.extIcon") are not equal
 - ❗ [E184] "common.installedFrom" is invalid. Please remove from [io-package.json](https://github.com/arteck/ioBroker.xsense/blob/master/io-package.json).
 - 👀 [W168] "common.keywords" should not contain "iobroker, adapter, smart home" [io-package.json](https://github.com/arteck/ioBroker.xsense/blob/master/io-package.json)

  - you are using this.clearInterval (OK) but for setInterval the std node version is still used. Mixing is not optimal. I would suggest to use this.setInterval. (https://github.com/arteck/ioBroker.xsense/blob/fd8c424f1c0b8eb91fb6b347f2c6f9460e405ce2/main.js#L85)
  
  - this.config.pollintervall is not checked / limited to reasonable values - if I'm correct. Especially a value of 0 would cause high load. Please limit to reasonable values within code.
 
  
McM1957

**Please add a comment when you have reviewed and fixed the suggestions or at least commented the suggestions and you think the adapter is ready for a re-review!**

reminder 28.8.2025

mcm1957:
@arteck 
Just a kind reminder / question. Did you have time to fix last issues?
Or did my last comment not reach you?

reminder 4.9.2025

arteck:
RE-CHECK!

mcm1957:
RE-CHECK!

mcm1957:

PR #5003, opened 2025-08-15
Add sofarcloud to latest

ltspicer:
Please add my adapter ioBroker.sofarcloud to latest.

### DIFF
--- a/sources-dist.json
+++ b/sources-dist.json
@@ -2441,6 +2441,11 @@
     "icon": "https://raw.githubusercontent.com/ioBroker/ioBroker.socketio/master/admin/socketio.png",
     "type": "communication"
   },
+  "sofarcloud": {
+    "meta": "https://raw.githubusercontent.com/ltspicer/ioBroker.sofarcloud/main/io-package.json",
+    "icon": "https://raw.githubusercontent.com/ltspicer/ioBroker.sofarcloud/main/admin/sofarcloud.jpg",
+    "type": "energy"
+  },
   "solaredge": {
     "meta": "https://raw.githubusercontent.com/iobroker-community-adapters/ioBroker.solaredge/master/io-package.json",
     "icon": "https://raw.githubusercontent.com/iobroker-community-adapters/ioBroker.solaredge/master/admin/solaredge.png",


ltspicer:
RE-CHECK!

mcm1957:
reminder 23.8.2025

ltspicer:
[sofarcloud.0.json](https://github.com/user-attachments/files/21812293/sofarcloud.0.json)

mcm1957:
@ltspicer 

- [ ] io-package translations are dummies only

  Please translate
  ```
    "desc": {
      "en": "Reads data from the SofarCloud server",
      "de": "Reads data from the SofarCloud server",
      "ru": "Reads data from the SofarCloud server",
      "pt": "Reads data from the SofarCloud server",
      "nl": "Reads data from the SofarCloud server",
      "fr": "Reads data from the SofarCloud server",
      "it": "Reads data from the SofarCloud server",
      "es": "Reads data from the SofarCloud server",
      "pl": "Reads data from the SofarCloud server",
      "uk": "Reads data from the SofarCloud server",
      "zh-cn": "Reads data from the SofarCloud server"
  ```

- [ ] i18n translations are fake

  Please translate. iobroker-dev npm run translate will do for you.

  
  ```
  "encryptedNative": [
    "password",
    "data2"
  ],
  "protectedNative": [
    "password",
    "data2"
  ],
  ```

- [ ] io-package config parameters not used at config

  https://github.com/ltspicer/ioBroker.sofarcloud/blob/293ad07618145e0a3f539d1209378d47ef29c390/io-package.json#L101

  Native section must contain all configuration parameters. 
  But  your native section contains config paramaters not uses at adminUI and in addtion id must not contain spaces. Please fix.

- [ ] modification of process.env

  Adapter must not modify process environment - at least if it is enabling compact mode as this would alter the environment of other adapters using the same process. 

  https://github.com/ltspicer/ioBroker.sofarcloud/blob/293ad07618145e0a3f539d1209378d47ef29c390/main.js#L10C1-L10C48

- [ ] redundant delay

  function dely is redundant as adapter code provides identical functionality (this.delay)

  (This is a suggestion only and not blocking)
 
- [ ] invalid characters must be filtered from object ids
  
```
function name2id(pName) {
}
```

  Object-Ids are not hardcoded within the adapter but retrieved from the cloud. Those Ids must be filtered so that the caontain only valid characters. Minimum check is against FORBIDDEN_CHARS + eliminating dots ans spaces. 

- [ ] reevaluate state roles

  All states are created using role value. value must only be used for readonly numeric states. Please see documentation.

  It seems that some labels are not translated into all supported languages, i.e. ...
  
  
McM1957

**Please attach a new object dump after fixing roles**

reminder 10.9.2024

ltspicer:
I hope, all issues are resolved ;)
[sofarcloud.0.json](https://github.com/user-attachments/files/22034417/sofarcloud.0.json)

mcm1957:

- [ ] i18n translations are still fake

  Please translate. iobroker-dev npm run translate will do for you.

  I.e. take a look into admin/i18n/fr (https://github.com/ltspicer/ioBroker.sofarcloud/blob/main/admin/i18n/fr/translations.json). The text is for sure not french.

  Please do a translation using npm run translate. You will have to delete existing incorrect translations as npm run translate does not change existing translations.
  

@ltspicer 
Please fix translation problem. Rest lgtm

reminder 9.9.2025

ltspicer:
Ok thanks.
I resolved the issues ;)

PR #5006, opened 2025-08-17
Update yahka to 1.1.5

tarikweiss:
In favor to issue [#536](https://github.com/jensweigele/ioBroker.yahka/issues/536) in YAHKA.


mcm1957:
And jsut noticed:
Standard tests will likely be failing as iobroker/testing is not declared as devDependency. 
Changing dependencies / package.json will require a new (patch) release.

PR #5049, opened 2025-09-05
fix vis-2-widgets-icontwo extIcon url

skvarel:
fix vis-2-widgets-icontwo extIcon url


skvarel:
This error will be solced with this PR

grafik

PR #5061, opened 2025-09-08
Pls add vis-2-widgets-icontwo to stable

skvarel:
Icon adapter for vis-2


skvarel:
Non-code adapter - they are just visualization icons

PR #5074, opened 2025-09-12
Add sbms to latest

buffoletti:


buffoletti:
new release v0.2.0 with modified object tree
[sbms.0 (1).json](https://github.com/user-attachments/files/22310555/sbms.0.1.json)

buffoletti:
RE-CHECK!

mcm1957:

- [ ] errors and warnings reported by adapter checker must be fixed

  Please add missing translation. iobroker-dev (npm run translate) can do the job for you.

  If yo have any question how to fix an error or warning, please let us know.

- [ ] @iobroker/testing and additional deps

  As you are using @iobroker testing you should be able to drop most testing related dependencies from dev_Dependencies:
    "@types/chai": "^4.3.20",
    "@types/chai-as-promised": "^7.1.8",
    "@types/mocha": "^10.0.10",
    "@types/proxyquire": "^1.3.31",
    "@types/sinon": "^17.0.4",
    "@types/sinon-chai": "^3.2.12",
    "chai": "^4.5.0",
    "chai-as-promised": "^7.1.2",

     Those modules are already provided by @iobroker/testing 5.1.0 and newer

- [ ] check dependencies

  All packages required by your code must be listed as dependencies.

  at least axios seems to be missing.

- [ ] evaluate writeState function

  Why do you use a write State function? setState changes states within the own adapter namespace. If you want to change foreign states you need to use steForeignState. So constructing a fully qualified id to use with setState is at least not required. setState blabla will change the state smbs.0.blabla.

  In addtion please note that setState without a callback is an async function. If its important to wait for the state to be changed you should use await setState(...)
  
- [ ] why do you use explicit debug flag?
   https://github.com/buffoletti/ioBroker.sbms/blob/4eace5d011c7dd6df411554dace2e1b9ec41e55a/lib/handler/html.js#L62
   Any reason why you do not use adapter.log.debug to log debug messages?

  Processing foreign states which are output of another adapter  should only be processed if the ack flag is true if there is no special reason to process unconfirmed changes too.

  Intermediate objects (device, channels, folders) are not created. I.e no object exists for balancing, cells., ... Those objects are most likely channels or folders. 

  cells.min, "xsbms.type", ... and others seem to be missing a creation of cells., xsbms objects too.

- [ ] hirarchy of objects is invalid

 The general hirarchy of objects is device - channel - state. There must be a maximum of one device within a tree, an adapter can have multiple device objects or even none.

  A device must not be below a channel. Channels can have channels als children if this really makes sense.

  Folders can be located at any level

  device, channel and folder object do NOT store any values. Values are stored with state objects.

  States are always at the end of a leave. States MUST NOT have any children.  This rule is definitly violated. Please restructure you object tree.

  It seems that some labels are not translated into all supported languages, i.e. ...
  
  Your axios timeout is set to 5000. So the minimum interval value msut be 5000 or greater. Otherwiese request could be queued faster than the timeout clears request from an unresponsive device. This will lead to an overlaod of the complet hist and possible crash the complet iobBroker host.

  Tests for node 20 are mandatory if this node version is allowed at engines clause. 
  Tests for node 22 are mandatory as this is the recommended node version. 
  
- [ ] linter setup

  Please use standard iobroker linter setup if acceptable. This is located at @iobroker/eslint-config. See migration guide at https://github.com/ioBroker/ioBroker.eslint-config/blob/main/MIGRATION.md.
   
  
McM1957

** Please attach a new object dump after fixing the structure**

buffoletti:
RE-CHECK!

buffoletti:
[sbms.0.json](https://github.com/user-attachments/files/22521950/sbms.0.json)
thanks for reviewing the adapter. i hope i got them all. 

If possible the adapter creator should be updated. the eslint thing was a harder than expected for me as beginner...

mcm1957:
RE-CHECK!

PR #5100, opened 2025-09-17
Add docker-manager to latest

GermanBluefox:


mcm1957:
please fix issues reported by checker
Thanks

mcm1957:
RE-CHECK!

mcm1957:
RE-CHECK!

mcm1957:
@Apollon77 **just a kind reminder.**
Please make the review for this adapter / project and merge if OK.
As this is no "0815" adapter - I cannot determine if it matches guidelines as it is or which adaptions you would suggest.
**So a review by you is required.**

PR #5188, opened 2025-10-12
Bumb TibberLink stable to 6.0.1

hombach:
Bumb TibberLink stable to 6.0.1 for support of 15 minutes prices.
6.0.0 is used about one week now - 6.0.1 just fixes a wrong number constant. 
And as this should be usable for all quite soon ..... ;)


mcm1957:
Well - might be.
I've dropped a note at checker to investigate

PR #5371, opened 2025-11-27
Update energy-tracker to 0.3.1

StefaniOSApps:
Please update my adapter ioBroker.energy-tracker to version 0.3.1.

*This pull request was created by https://www.iobroker.dev 509607d.*


StefaniOSApps:
I know this is a very short timeframe (only about one day), but the holiday season is coming up and I will be on vacation soon. I would like to wrap this up beforehand if possible.

This update only includes dependency bumps and a small API path adjustment (from v1 to v2). I can confirm that there are no issues expected when promoting it to *stable*. It’s essentially an end-of-year cleanup.

If promoting it right now is not acceptable, please feel free to let it sit for about a week and then merge it.

Thanks a lot!

PR #5421, opened 2025-12-23
Add autodarts to latest

skvarel:
Please add my adapter ioBroker.autodarts to latest.

*This pull request was created by https://www.iobroker.dev 974ec1c.*

PR #5427, opened 2025-12-23
Add autodarts entry to latest

mcm1957:
@skvarel

This PR replaces PR #5421.

Original PR was based on a damaged sources-dist.json file. This fact confused checker.
Please watch this PR for further information / feddback.

Sorry for the inconvinience.

object-dump see original PR #5421
---> SEE BELOW for new dump


skvarel:
[autodarts.0.json](https://github.com/user-attachments/files/24386126/autodarts.0.json)

mcm1957:
@skvarel 

- [ ] reevaluate state roles

  Only a small request: config states must either be RO or have role level. Role value requires a RO state.
      
McM1957

skvarel:
reevaluate state roles

will be solved with next update

PR #5437, opened 2025-12-28
Add smartfriends to latest

Black-Thunder:
Please add my adapter ioBroker.smartfriends to latest.

*This pull request was created by https://www.iobroker.dev 974ec1c.*


mcm1957:
RE-CHECK!

Black-Thunder:
@mcm1957 Thanks for the review.

>  creation of intermediate objects missing
> Intermediate objects (device, channels, folders) are not created. I.e no object exists for 'details', 'info', ...
> There is no object created for 'devices' (sse empty device type)

Thanks for the heads-up. I totally missed that. I've already released a new version that adds the "devices" folder.

> is the duplicated 14326 really desired?

Yes, this depends on the devices a user has connected to the gateway. Usually, one (master) device consists of several sub-devices. In my case, the master device (14326) only contains itself, which is why it seems duplicated. Typically, there are many more different devices under one master device.

PR #5486, opened 2026-01-19
Add solectrus-influxdb to latest

patricknitsch:


mcm1957:
@patricknitsch 

- [ ] Readme.md is very short

  The README.md of this adapter is very minimalistic. Please add a short description about the functionality of the adpter,  add links the the manufacturer site (if applicable) and add a description of the configuration of the adapter. If you provide an external documentation - which should be located at doc/xx directory - please add links to at least the english documentation.

  It's completly OK to provide an extended documentatione externally. But the basic information about the purpose of the adapter including naming the relevant manufactorer or device is required. The content of README.md is includd into ioBroker webpages - so this information must be presnt at README.md even it is duplicated.

  Please keep README.md written susing english language only.

  ```
  "encryptedNative": [
    "password",
    "data2"
  ],
  "protectedNative": [
    "password",
    "data2"
  ],
  ```

  When processing foreign states which are output of another adapter  processing should normally only occure if the ack flag is true.

  Buttons have attribute read:false. So you should not rely on the value returned by the button. Writing to a button with ANY value and ACK=FALSE should trigger the action. (But I will not consider this blocking as many adapter implement buttons not 100% correctly. Typical user write to a button with value true and ack=false. User must not expect tha the adapter resets the value of an button.)

  Intermediate objects (device, channels, folders) are not created. I.e no object exists for ''buffer', 'sensors'

McM1957

reminder 13.2.2026

patricknitsch:
@mcm1957 : Thank you for your feedback. I have checked all open points and fixed them. Please feel free to do a re-view. Thanks in advance.

PS: Tried to activate all checkboxes.

patricknitsch:
Understood. Thank you in advance.

mcm1957:
lgtm

PR #5497, opened 2026-01-24
Add foxesscloud to latest

skvarel:
Please add my adapter ioBroker.foxesscloud to latest.

*This pull request was created by https://www.iobroker.dev c94083c.*

### DIFF
--- a/sources-dist.json
+++ b/sources-dist.json
@@ -795,6 +795,11 @@
     "icon": "https://raw.githubusercontent.com/TA2k/ioBroker.ford/master/admin/ford.png",
     "type": "vehicle"
   },
+  "foxesscloud": {
+    "meta": "https://raw.githubusercontent.com/inventwo/ioBroker.foxesscloud/main/io-package.json",
+    "icon": "https://raw.githubusercontent.com/inventwo/ioBroker.foxesscloud/main/admin/foxesscloud.png",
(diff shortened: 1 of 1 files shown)


mcm1957:
@skvarel 

- [ ] possible unneeded dependencies

  As you use @iobroker/eslint-config no local includes of eslint / prettier packages are required. They should be removed
  Please see migration guide: 

  Why do you need those devDependencies?

    "dotenv": "^17.2.3",
    "eslint": "^9.0.0",
    "should": "^13.2.3"

  eslint is definitly included into standard iobroker eslint config and shoul not be included explecitly.

- [ ] Unusual file reference

  Why did you modify test command? "test:js": "node node_modules/@iobroker/testing/node_modules/mocha/bin/mocha.js ... refers to a module located as node_moduls of a dependency. It's not guranteed that it is located there as npm might choose to optimize and use shared locations, I suggest to use the standard definition (mocha .... without any special setup).
 
- [ ] when using hardcoded translations all languages should be provided
  see https://github.com/inventwo/ioBroker.foxesscloud/blob/9b7fca5ff653a7df28ea5560ef5437955c95e797/main.js#L67
  You can use iobroker translator to generat a ready to copy code block.
  
McM1957

reminder 13.2.2026

PR #5544, opened 2026-02-18
Add schlueter-thermostat to latest

mcm1957:
splitted from #5486 
@[patricknitsch](https://github.com/patricknitsch)

### DIFF
--- a/sources-dist.json
+++ b/sources-dist.json
@@ -2361,6 +2361,11 @@
     "icon": "https://raw.githubusercontent.com/ioBroker/ioBroker.scheduler/master/admin/scheduler.png",
     "type": "logic"
   },
+  "schlueter-thermostat": {
+    "meta": "https://raw.githubusercontent.com/patricknitsch/ioBroker.schlueter-thermostat/main/io-package.json",
+    "icon": "https://raw.githubusercontent.com/patricknitsch/ioBroker.schlueter-thermostat/main/admin/schlueter-thermostat.png",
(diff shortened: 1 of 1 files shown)


mcm1957:
@patricknitsch 

  Intermediate objects (device, channels, folders) are not created. I.e no object exists for 'temperature', 'setpoint', ...

- [ ] some writeable states are not subscribed

   Several writeable states are not subscribed. Why do you use the ruther unusal construct of using a state to hold a value and an extra apply button? Normally writing a state value causes the adapter to react on the entered value. Please reevaluate this concept. In addtion values entered should be acked by adapter to signal it received the data.

- [ ] object types do not meet requirments

  Objects must be structured by type device - channel - state. device and channel are optional, each of them or both might be missing. States never must have children. Within one tree a maximum ob one device might exist. devices are not allowed to be located below channels. Multiple channels within one tree are allowed if they really make sense.

  Folders can be located anywhere between devices / channels and states.

  So your devices below a channels and multple devices within one tree are invalid. This will for sure cause problems with device detector. Reduce to one device per tree and eventually use top level folders if anything besides 'groups' might exist or remove groups completly if there is no parallel tree planned. 
  
McM1957

reminder 15.3.2026

mcm1957:
@patricknitsch 

**Please add a new object dump t**o check fixed object structure

Sorry I missed to ask for it at the first pass.

mcm1957:
@patricknitsch 

PR #5577, opened 2026-03-03
Add ioBroker.simple-proxy-manager to latest

lubepi:
## Add new adapter: simple-proxy-manager

**Type:** network

**Description:** Simple HTTPS reverse proxy manager with virtual hosts, IP filtering and automatic certificate reload

**Links:**
- Repository: https://github.com/lubepi/ioBroker.simple-proxy-manager
- npm: https://www.npmjs.com/package/iobroker.simple-proxy-manager

**Checklist:**
- [x] The adapter has a unique name
- [x] Published on npm (v0.1.1)
- [x] Uses @iobroker/adapter-core
- [x] Has io-package.json with all required fields (titleLang, desc, licenseInformation, tier, etc.)
- [x] Has README.md
- [x] bluefox added as npm collaborator
- [x] Adapter checker passed (https://adapter-check.iobroker.in)

### DIFF
--- a/sources-dist.json
+++ b/sources-dist.json
@@ -2466,6 +2466,11 @@
     "icon": "https://raw.githubusercontent.com/ioBroker/ioBroker.simple-api/master/admin/simple-api.png",
     "type": "communication"
   },
+  "simple-proxy-manager": {
+    "meta": "https://raw.githubusercontent.com/lubepi/ioBroker.simple-proxy-manager/main/io-package.json",
+    "icon": "https://raw.githubusercontent.com/lubepi/ioBroker.simple-proxy-manager/main/admin/simple-proxy-manager.png",
+    "type": "network"
+  },
   "skiinfo": {
     "meta": "https://raw.githubusercontent.com/oweitman/ioBroker.skiinfo/main/io-package.json",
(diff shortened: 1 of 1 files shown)


lubepi:
[simple-proxy-manager.0.json](https://github.com/user-attachments/files/25728286/simple-proxy-manager.0.json)

mcm1957:
@lubepi 

- [x] consider using 'node:package' format

- [x] reevaluate state roles

  Certificate date strings should most like have role 'date'. See https://github.com/ioBroker/ioBroker.docs/blob/master/docs/en/dev/stateroles.md#common

 
- [x] check and limit configurable timeouts / intervals
  
  Suggestion to check checkIntervalMs to be positive and to be not too large.

 
- [x] adapt testing to supported node releases

  Tests for node 20 are mandatory as long as engines clause does not require a higher nod.js release. 
  Tests for node 22 are mandatory as this is the recommended node version. 
  
McM1957

reminder 8.4.2026

mcm1957:
Sorry, but @Apollon77 seems to be out of free time currently.
I'll release the adapter so that you do not have to wait any longer. If there are any concersn, @Apollon77 will contact you.

mcm1957:

PR #5664, opened 2026-03-22
Add influxdb-prologger to latest

simpros:
Please add my adapter ioBroker.influxdb-prologger to latest.

*This pull request was created by https://www.iobroker.dev a59a21f.*

### DIFF
--- a/sources-dist.json
+++ b/sources-dist.json
@@ -1335,6 +1335,11 @@
     "icon": "https://raw.githubusercontent.com/ioBroker/ioBroker.influxdb/master/admin/influxdb.png",
     "type": "storage"
   },
+  "influxdb-prologger": {
+    "meta": "https://raw.githubusercontent.com/simpros/ioBroker.influxdb-prologger/main/io-package.json",
+    "icon": "https://raw.githubusercontent.com/simpros/ioBroker.influxdb-prologger/main/admin/influxdb-prologger.png",
+    "type": "storage"
+  },
   "innogy-smarthome": {
     "meta": "https://raw.githubusercontent.com/PArns/ioBroker.innogy-smarthome/master/io-package.json",
     "icon": "https://raw.githubusercontent.com/PArns/ioBroker.innogy-smarthome/master/admin/innogy-smarthome.png",


mcm1957:
@simpros 
Ans sorry for the long time to provide some feedback.

- [ ] check dependencies

  Why do you install those packages as dependency? 
  Wouldn't it be sufficient to have them as devDependency?
  But see request to use jsonConfig below first ...

    "@mui/icons-material": "^6.5.0",
    "@mui/material": "^6.5.0",
    "react": "^18.3.1",
    "react-dom": "^18.3.1"

- [ ] Why do you use react for adminUI?

  Why do you use react for adminUI. It looks like all functionality of admin UI could be implemented using jsonConfig which is the standard configuration environment. jsonConfig does not require any special knowlegde and does not need to be compiled. In addtion it supports a common look and feel.

  Please migrate to jsonConfig or explain which functionaliyt you need which is not provided by jsonConfig.
 
- [ ] Why do you use this.config.enableDebugLogs ?

  ```
	if (this.config.enableDebugLogs) {
		this.log.debug(`On-change write for "${id}" -> bucket "${group.bucket}": ${line}`);
	}
  ```

  Why do you encapsulate oll debug logs into an extra enableDebugLogs clause? That looks pretty redundant. If there is no very special reson to do tghis, please remove enableDebugLogs as log.debug can be enabled / disabled anyway at instance level.

- [ ] Adapter does not meet responsive design requirements

  AdminUI on small screens does not adjust and is cut of. 
  image
  
  Note: jsonConfig would solve this automatically.

  When processing foreign states which are output of another adapter  processing should only occure if the ack flag is true normally.


- [ ] labels seem to have no translation into all languages
  
  It seems that some labels are not translated into all supported languages. 
  Please consider adding missing translations for other languages than en and de.

  You may use adapter-dev / npm run translate to translate all languages within i18n.
  
  
  
McM1957

reminder 25.4.2026

simpros:
Hey @mcm1957

thank you very much for the feedback!

----

> Why do you use react for adminUI?

I started with the jsonConfig but I was not able to get the UX that I wanted / needed to make the setup easy.
The creation of logging groups would proabably still work with the jsonConfig.
image

But where it gets tricky, is that this collection interactivally connects to the datapoints, and I was not able to get this smoothly done with the jsonConfig.



I don't think linking that data of two tabs is feasable with jsonConfig.

------

> check dependencies

I'll update the dependencies 👍 

------

> Why do you use this.config.enableDebugLogs ?

Good catch, in the typescript that I currently use I've just commented out the debug logs when not needed and during the migration (with AI) it got implemented it in the same way. What I'm not so sure about, is the potential **massive** amounts of logs that this could produce (could easily get 100+ logs per second).

I think it's still a valid option in the adapter to make it either debug or not 🤔 


-----

> Adapter does not meet responsive design requirements

Will work on that! Thank you for pointing that out. 👍 

-----

> onStateChange handler ignores ack flag

I'll also fix that! 👍 

-----

>  labels seem to have no translation into all languages

I'll add them 👍

simpros:
I've released version 1.0.1, where the low-hanging fruits were fixed (responsive-design, ack-flag recognition, translations)

mcm1957:
This is a intermediate RE-REVIEW result only. Waiting for Feedbacl from @Apollon77 / @GermanBluefox in respect to react usage.

- [x] check dependencies

  Why do you install those packages as dependency? 
  Wouldn't it be sufficient to have them as devDependency?
  But see request to use jsonConfig below first ...

    "@mui/icons-material": "^6.5.0",
    "@mui/material": "^6.5.0",
    "react": "^18.3.1",
    "react-dom": "^18.3.1"

- [ ] Why do you use react for adminUI?

  Why do you use react for adminUI. It looks like all functionality of admin UI could be implemented using jsonConfig which is the standard configuration environment. jsonConfig does not require any special knowlegde and does not need to be compiled. In addtion it supports a common look and feel.

  Please migrate to jsonConfig or explain which functionaliyt you need which is not provided by jsonConfig.

  OK, I got your explanation. Will check with @GermanBluefox and ask for his oppinion / review.
 
- [x] Why do you use this.config.enableDebugLogs ?

  ```
	if (this.config.enableDebugLogs) {
		this.log.debug(`On-change write for "${id}" -> bucket "${group.bucket}": ${line}`);
	}
  ```

  Why do you encapsulate oll debug logs into an extra enableDebugLogs clause? That looks pretty redundant. If there is no very special reson to do tghis, please remove enableDebugLogs as log.debug can be enabled / disabled anyway at instance level.

   OK, don't understand the benefit anyway. But I do not consider it blocking.

- [ ] Adapter does not meet responsive design requirements

  AdminUI on small screens does not adjust and is cut of. 

  Will be verified together with question about react. 

- [x] onStateChange handler ignores ack flag 

  When processing foreign states which are output of another adapter  processing should only occure if the ack flag is true normally.


- [x] labels seem to have no translation into all languages
  
  It seems that some labels are not translated into all supported languages. 
  Please consider adding missing translations for other languages than en and de.

  You may use adapter-dev / npm run translate to translate all languages within i18n.
  
  
reminder 28.4.2026

Apollon77:
@simpros Thanks for your contribution. For me the main question is: why do we need another influxdb adapter when we already have one? When I see it right it adds more flexibility in the data logging configuration "into the DB", but getting the data out again to use in ioBroker is nit the goal, right?! SO it is not following the usual "history adapter featuresets" - From that regard would be interesting that this gets clear in the readme because else users migth have expectations for that

simpros:
hey @Apollon77 👋 

You are right, it's another way to write data FROM ioBroker to influxdb v2. It's not planned to be used to get data into ioBroker.
It might get some value calculation functions in the future (e.g. calculating a median/average/max/min value within the cron period of the values) but that also just affects values written to the influxdb.

I'll modify the README, thank you for the feedback 👍

mcm1957:
Internal Ref: TG Poll @GermanBluefox  at 21.4.2026
https://t.me/c/2020938082/2020

simpros:
Which telegram-group would I need to join to see the result? 🙂

mcm1957:
Sorry - should have hidden the remakr and / or make it more clear. This reference is only a note for me so that I known when and where I triggered Bluefox.

You will get feedback here from Bluefox (or me)  as soon as he has responded.

mcm1957:
@simpros 
Please fix errors reported by adapter checker.
Dependency updates are you decision when to perform best.

reminder 13.6.2026

mcm1957:
@simpros
Please fix errors reported by adapter checker.
Dependency updates are you decision when to perform best.

reminder 28.6.2026

simpros:
Heyhey 👋 

I've added version 1.1.0, sorry for the long delay. That version addressed the noted Readme adjustments and added a debounce the instant writes to not cause unneccessary write pressure to the influxdb.

mcm1957:
No problem. The long dely is mainly caused by me.

mcm1957:
In the meantime I've some assistent aquired - here's it's output:

# Adapter Review Report

| Field | Value |
|---|---|
| **Adapter Name** | ioBroker.influxdb-prologger |
| **Adapter Version** | 1.1.0 |
| **Review Checklist Version** | V2.0 |
| **Review Date** | 2026-06-26 |
| **Reviewer** | Claude (automated review) |

---

## Summary

All source files under `src/`, `src-admin/src/`, `io-package.json`, `package.json`, and `README.md` were reviewed.  
Test files (`*.test.ts`, `test/`) were excluded as per checklist instructions.

---

## Results

### ✅ Passed Checks

- [x] ✅ **README language** — `README.md` is written entirely in English. No mixed-language content found. A `README.de.md` does not exist, which is acceptable (not required).

- [x] ✅ **Sensitive information not logged** — The `token` field is declared in both `encryptedNative` and `protectedNative` in `io-package.json`. A review of all log calls (`this.log.error`, `this.log.warn`, `this.log.info`, `this.log.debug`) in `src/main.ts` and `src/lib/influx-client.ts` confirms that `this.config.token` is never passed to any log statement. The token is only used in HTTP `Authorization` headers.

- [x] ✅ **Object ID character filtering** — This adapter does not construct any ioBroker object IDs from user input or external data. The only state IDs created by the adapter are the hardcoded `info` and `info.connection`, which use safe characters. Object IDs used for subscriptions come from user configuration (`dp.objectId`) but are used as-is for lookups (not as new object IDs to create), so filtering is not required here.

- [x] ✅ **State IDs use English wording** — All state IDs created by the adapter (`info`, `info.connection`) use English identifiers.

- [x] ✅ **`common.name` and `common.desc` use English or i18n** — In `io-package.json`, `common.desc` uses a full i18n object covering all 11 supported languages. `common.name` is `"influxdb-prologger"` (a product name, not a natural-language label). Instance object names (`"Information"`, `"Device or service connected"`) use English, which is acceptable for technical channel/state names. `titleLang` uses an i18n object with all 11 languages.

- [x] ✅ **Log messages in English, no translation mechanism** — All log messages in `src/main.ts` and `src/lib/influx-client.ts` are plain English string literals. No calls to `I18n.t()` or similar are used in log messages.

- [x] ✅ **Admin UI uses i18n translations** — All visible UI text in the admin tabs (`ConnectionTab.tsx`, `GroupsTab.tsx`, `DatapointsTab.tsx`, `AdvancedTab.tsx`, `App.tsx`) uses `I18n.t('key')` exclusively. No hardcoded English strings are displayed to the user.

- [x] ✅ **All 11 supported languages present in i18n files** — Translation files exist for all required languages: `en`, `de`, `ru`, `pt`, `nl`, `fr`, `it`, `es`, `pl`, `uk`, `zh-cn`. All 11 files were verified to be present under `src-admin/src/i18n/`.

- [x] ✅ **i18n files contain all supported languages and no extra keys** — Each i18n JSON file contains exactly the same 47 keys as `en.json`. No extra or missing keys were detected across any of the 11 language files.

- [x] ✅ **No unused i18n keys** — All keys in the i18n files are referenced by `I18n.t('...')` calls in the admin UI source files. Cross-referencing keys in `en.json` against usage in `App.tsx`, `ConnectionTab.tsx`, `GroupsTab.tsx`, `DatapointsTab.tsx`, and `AdvancedTab.tsx` shows full coverage. *(Note: `"datapoints"` and `"loggingGroups"` are defined but not directly found in a visible `I18n.t()` call in the tab files — these may be used by `GenericApp` internals or reserved. No action required.)*

- [x] ✅ **`onObjectChange` handler not registered** — The adapter does not register an `onObjectChange` handler, which is correct since it does not need to react to object changes.

- [x] ✅ **`onStateChange` handler honors the `ack` flag** — The `shouldProcessOnChangeState()` function in `src/lib/state-change.ts` correctly distinguishes own states (ignore if `ack === true`) from foreign states (process only if `ack === true`). This logic is applied in `main.ts` `onStateChange()` before any processing occurs.

- [x] ✅ **HTTP/fetch calls use a timeout** — All `fetch()` calls in `src/lib/influx-client.ts` use `AbortSignal.timeout(this.config.writeTimeout || 5000)` (write endpoint) or `AbortSignal.timeout(5000)` (health/test endpoints). No unguarded fetch calls exist.

- [x] ✅ **No fixed-timestamp cloud access** — The adapter does not schedule writes at specific wall-clock timestamps (e.g., every full hour or midnight). Cron groups fire on user-configured cron expressions.

- [x] ✅ **No code modification of npm-delivered directories** — The adapter does not write to or modify any files within its own install directory or npm package directories at runtime.

- [x] ✅ **README does not contain GitHub-direct or raw npm install instructions** — The `README.md` contains no `npm install` commands or GitHub-URL-based installation instructions for end users.

- [x] ✅ **`package.json` scripts — no install-time execution scripts** — The `package.json` does not contain `preinstall`, `postinstall`, or `install` scripts. All defined scripts (`build`, `watch`, `test`, `lint`, etc.) are standard development-time scripts not executed during `npm install` by end users.

- [x] ✅ **No dead code detected** — All exported functions and classes in the lib files are used by `main.ts`. All config parameters defined in `io-package.json` native/encryptedNative are referenced in the adapter code.

- [x] ✅ **`onStateChange` handler is used and justified** — The adapter subscribes to foreign states for on-change groups and processes them in `onStateChange`. The handler is properly used and not empty.

---

### ⚠️ Checks Requiring Action

- [ ] ❌ **Configurable timeout (`writeTimeout`) not validated/clamped in code** — The `writeTimeout` config value is passed directly to `AbortSignal.timeout(this.config.writeTimeout || 5000)` without any bounds check. Node.js `AbortSignal.timeout` (and underlying `setTimeout`) accepts a maximum of 2,147,483,647 ms. If a user enters `0`, a negative value, or an extremely large value via direct config edit, the behavior is undefined or may silently fail. The Admin UI uses `min: 1000, max: 60000` as a hint only — this is not enforced in code.

  > **References:**
  > - `src-admin/src/tabs/AdvancedTab.tsx`, line ~31: UI min/max are hints only.
  > - `src/lib/influx-client.ts`, line ~83: `AbortSignal.timeout(this.config.writeTimeout || 5000)` — no upper/lower bound enforcement.
  > - `src/lib/influx-client.ts`, line ~132: `AbortSignal.timeout(this.config.writeTimeout || 5000)` — same issue in `testConnection()`.

- [ ] ❌ **Configurable `flushWindowMs` not validated/clamped in code** — The `flushWindowMs` value per group is passed directly to `setTimeout()` inside `SpontaneousWriteBuffer`. Node.js `setTimeout` has a maximum of 2,147,483,647 ms; values of `0` or negative numbers cause immediate or erratic firing. The admin UI enforces `min: 100` at the UI level only — not in code.

  > **References:**
  > - `src-admin/src/tabs/GroupsTab.tsx`, line ~145: `slotProps={{ htmlInput: { min: 100 } }}` — UI hint only.
  > - `src/lib/spontaneous-write-buffer.ts`, line ~72: `slot.timerId = setTimeout(() => { void this.flushSlot(key); }, this.windowMs)` — `this.windowMs` is not validated.
  > - `src/main.ts`, line ~148: `const windowMs = group.flushWindowMs ?? 5000;` — no clamping before passing to `SpontaneousWriteBuffer`.

- [ ] ❌ **Configurable `maxRetries` not validated in code** — The `maxRetries` value controls the retry loop count `const maxAttempts = this.config.retryOnError ? (this.config.maxRetries || 3) + 1 : 1`. A user-supplied value of `-1` would result in `maxAttempts = 0`, silently disabling all writes (the for-loop would never execute). The Admin UI shows `min: 0, max: 10` but this is not enforced in code.

  > **References:**
  > - `src-admin/src/tabs/AdvancedTab.tsx`, line ~46: UI `min: 0, max: 10` — hint only.
  > - `src/lib/influx-client.ts`, line ~72: `const maxAttempts = this.config.retryOnError ? (this.config.maxRetries || 3) + 1 : 1` — no lower-bound check.

- [ ] ❌ **`SpontaneousWriteBuffer` uses bare global `setTimeout` instead of adapter-managed `this.setTimeout`** — The adapter correctly passes `this.setTimeout.bind(this)` to `InfluxClient` for retry delays, but `SpontaneousWriteBuffer` uses the native global `setTimeout` directly. This means the buffer's flush timer is not registered with ioBroker's adapter lifecycle and will not be automatically cleared on `onUnload`. While `flushAll()` calls `clearTimeout` on pending timers, a timer that fires concurrently during unload could attempt a write after the adapter is torn down.

  > **References:**
  > - `src/lib/spontaneous-write-buffer.ts`, line ~72: `slot.timerId = setTimeout(...)` — uses global `setTimeout`.
  > - `src/main.ts`, line ~148–155: `new SpontaneousWriteBuffer(...)` — adapter's `this.setTimeout` is not passed to the buffer constructor.

---

## Notes

- The adapter's overall code quality is high: well-structured TypeScript, good separation of concerns across lib files, clear naming, and solid retry/backoff logic.
- The i18n coverage is excellent — all 11 required languages are fully translated with consistent key sets.
- The ack-flag handling is correctly implemented.
- The `onUnload` shutdown sequence (stop crons → flush buffers → signal disconnected) is logically correct.


But don't be confused. I do not consider the issue as blocking. But I would recommend to review the, Check / limiting paramater values within code is good practice as the could be manipulated outside of adminUI to (using object access).

mcm1957:

PR #5708, opened 2026-03-26
Updated air-q to 1.1.0

Sibgatulin:
Long overdue update. Stable is at 1.0.5 while 1.1.0 was released on Feb 18.

Skipped releases contain important fixes:

    1.0.6: night mode support (skip polling when device WiFi is off)
    1.0.7: fix missing sensor handling, incorrect translations
    1.1.0: 19 new sensors (mold, refrigerants, gas sensors)

Fixed the issues as @mcm1957 requested in #5682


mcm1957:
There are two solutions;

- commit build/ tree to github (most used way)
or
- set flag common.nogit to true at io-package.json to indicate that this adapter cannot be installed from github (as build is missing in giuthub)

Bith variants are valid ans should solve the reported issue.

PR #5736, opened 2026-03-31
Add flowers to latest

sadam6752-tech:
Please add my adapter ioBroker.flowers to latest.

*This pull request was created by https://www.iobroker.dev e435dad.*

### DIFF
--- a/sources-dist.json
+++ b/sources-dist.json
@@ -790,6 +790,11 @@
     "icon": "https://raw.githubusercontent.com/ioBroker/ioBroker.flot/master/admin/flot.png",
     "type": "visualization"
   },
+  "flowers": {
+    "meta": "https://raw.githubusercontent.com/sadam6752-tech/ioBroker.flowers/main/io-package.json",
+    "icon": "https://raw.githubusercontent.com/sadam6752-tech/ioBroker.flowers/main/admin/flowers.png",
+    "type": "climate-control"
+  },
   "followthesun": {
     "meta": "https://raw.githubusercontent.com/HGlab01/ioBroker.followthesun/main/io-package.json",
     "icon": "https://raw.githubusercontent.com/HGlab01/ioBroker.followthesun/main/admin/followthesun.png",


sadam6752-tech:
[flowers.0.json](https://github.com/user-attachments/files/26374192/flowers.0.json)
RE-CHECK!

mcm1957:
@sadam6752-tech 

- [x] errors reported by adapter checker must be fixed

  Please fix 
  [E8915] Dependabot npm entry (directory: "/") has no "cooldown" configured. A cooldown of at least 7 days (e.g. "cooldown: { default: 7 }") is recommended to reduce supply chain risk.

  A PR has been provided

- [x] possible unneeded dependencies

  As you use @iobroker/eslint-config no local includes of eslint / prettier packages are required. They should be removed
  Please see migration guide: 

  "eslint": "^9.17.0"

  Intermediate objects (device, channels, folders) are not created. I.e no object exists for 'plants', 'ficus', ...

  Image  

  NOTE: Code looks like intermediate Objects are already created. Maybe the attached object dump is / was old? Please check and attach new dump.

- [ ] reevaluate state roles

  NOTE: Code looks like roles are created currently. Maybe the attached object dump is / was old? Please check and attach new dump.

  
McM1957

**Please attach new object dump after fixing**

reminder 14.5.2026

sadam6752-tech:
[flowers.0.json](https://github.com/user-attachments/files/27250803/flowers.0.json)
Other issues have been corrected in the latest release.

sadam6752-tech:
All issues have been fixed in v0.3.7 – v0.3.9:

E8915 – Fixed: added cooldown: { default-days: 7 } to dependabot npm entry. Also added separate entry for test/package directory (W8905).

eslint in devDependencies – Fixed: removed "eslint": "^9.17.0" from devDependencies (v0.3.7).

button read=false – Fixed: sendDailyReport and sendWeeklyReport now have read: false as required by ioBroker role spec (v0.3.9).

intermediate objects – The code already creates plants (channel) and plants.<name> (device) via extendObjectAsync since v0.3.3. The previously attached dump was from an older version. New dump attached below.

object cleanup – Added: objects for plants removed from config are now automatically deleted on adapter start (v0.3.8).

CHANGELOG_OLD.md – Added (v0.3.7).

[flowers.0.json](https://github.com/user-attachments/files/27252027/flowers.0.json)

RE-CHECK!

mcm1957:
reminder 11.5.2026

mcm1957:
Thanks for fixing things. And sorry for long delay.

I'm soryy, but I must raise an additonal issue:

image

The structure of objects is invalid.

The order must be DEVICE - CHANNEL - STATE
states must not have any children (ok)
device and channel are optional.
You can add folder at any position (but not below state)

BUT there must not be a second device below a device within the same tree (ok)
AND there must not be a device below a channel (THATS INCORRECT)

There might be a channel below a channel if this makes sense.

**So garten_links as device below the channel plants is invalid.**

You can change plants to type folder and everthing would be ok i.e.

**To speed things up I will release the adapetr anyway, but please fix the object types as the incorrect order will cause problems with iot and otehr parts using type detector.**

mcm1957:

PR #5739, opened 2026-04-01
Add enpal to latest

skvarel:
Please add my adapter ioBroker.enpal to latest.

*This pull request was created by https://www.iobroker.dev e435dad.*


mcm1957:

- [ ] reevaluate state roles

  Currently all states use "value". This is Ok as all values are numeric. But I would suggest to evaluate whether some table mapping the states to useful roles could be implemented. If the output of the adapter should be used in conjunction with iot or devicemanger useful roles are required for the teype detector to guess the device type.

  This is a suggestion and NOT blocking

  
McM1957

mcm1957:

PR #5752, opened 2026-04-05
Update volvo to 2.0.1

Vinnedinho:


Vinnedinho:
vccapikey is in protectedNative. having it in encryptedNative cased problems

mcm1957:
@Vinnedinho 
As 2.0.0 is already available since app. a week and used by > 5% of users I would suggest to raise stable directly to 2.0.1 instead of making two update withion a short timeframe.

Please comment.

EDIT: I can adjust the PR if it's ok for you.

Vinnedinho:
i changed the version to 2.0.1.

PR #5756, opened 2026-04-06
Added linky to sources-dist.json

raintonr:

PR #5757, opened 2026-04-06
Add linky to latest

mcm1957:
@raintonr 
This PR replaces #5756.

Please watch this PR and react on feedback

### DIFF
--- a/sources-dist.json
+++ b/sources-dist.json
@@ -1451,6 +1451,11 @@
     "icon": "https://raw.githubusercontent.com/Smart-Gang/ioBroker.linktap/master/admin/LinkTap_Logo.png",
     "type": "garden"
   },
+  "linky": {
+    "meta": "https://raw.githubusercontent.com/iobroker-community-adapters/ioBroker.linky/main/io-package.json",
+    "icon": "https://raw.githubusercontent.com/iobroker-community-adapters/ioBroker.linky/main/admin/linky.png",
+    "type": "energy"
+  },
   "linux-control": {
     "meta": "https://raw.githubusercontent.com/Scrounger/ioBroker.linux-control/master/io-package.json",
     "icon": "https://raw.githubusercontent.com/Scrounger/ioBroker.linux-control/master/admin/linux-control.png",


raintonr:
It's now...

```json
{
    "i18n": true,
    "type": "panel",
    "items": {
        "serialPort": {
            "type": "text",
            "label": "Serial port device",
            "xs": 12,
            "sm": 12,
            "md": 12,
            "lg": 9,
            "xl": 6
        }
    }
}
```

raintonr:
> You do not miss anythink. It's all right.

So can we merge this?

mcm1957:
image

Object Dump still missing
And I did not do a review yet.

raintonr:
[linky.json](https://github.com/user-attachments/files/26999247/linky.json)

Object dump attached.

mcm1957:
reminder 7.5.2026

mcm1957:
@raintonr 

- [x] object id casing

  Any special reason that all object ids are uppercase? Normally objectIds are lower or mixed Case.

  Thats only a suggetion. If theres a special reason, it's OK to keep this.

- [x] writeable states

  All sattes are writeable. But as no onStateChange handler is present and the adapter shows data received using seriel interface only this is most likely wrong. Please correct.

- [x] reevaluate state roles

  Several states are missing roles completly.
  Source contains wrog roles as Wh is a unit to be used for energy, W is the unit used for power.

  Please check and fix roles.  
  
McM1957

reminder 14.5.2026

raintonr:
> * [ ]  object id casing
>   Any special reason that all object ids are uppercase? Normally objectIds are lower or mixed Case.
>   Thats only a suggetion. If theres a special reason, it's OK to keep this.

The casing is like this to match names of the TiC specification so think should be left as is.

> * [x]  writeable states

Fixed.

> * [x]  reevaluate state roles

Fixed. I note some changes to the roles doco since this adapter was created.

mcm1957:
reminder 21.5.2026

mcm1957:
ltgm

mcm1957:

PR #5831, opened 2026-04-12
Add life360ng to latest

skvarel:
Please add my adapter ioBroker.life360ng to latest.

### DIFF
--- a/sources-dist.json
+++ b/sources-dist.json
@@ -1431,6 +1431,11 @@
     "icon": "https://raw.githubusercontent.com/TA2k/ioBroker.libre/main/admin/libre.png",
     "type": "health"
   },
+  "life360ng": {
+    "meta": "https://raw.githubusercontent.com/inventwo/ioBroker.life360ng/main/io-package.json",
+    "icon": "https://raw.githubusercontent.com/inventwo/ioBroker.life360ng/main/admin/Life360_s.svg",
+    "type": "geoposition"
+  },
   "lifx": {
     "meta": "https://raw.githubusercontent.com/foxthefox/ioBroker.lifx/master/io-package.json",
     "icon": "https://raw.githubusercontent.com/foxthefox/ioBroker.lifx/master/admin/lifx_logo.png",


skvarel:
[life360ng.0.json](https://github.com/user-attachments/files/26658862/life360ng.0.json)

mcm1957:
reminder 26.4.2026

mcm1957:
@skvarel 

- [x] process.env usage

   Please evaluate if using process.env is really needed. If this is needed, common.compact must be set to false in most cases as when using compact mode  multiple instances share one process and hence share the process environment. In addition note that the process environment is readable by all instances running within the same process when using compact mode. As far as I know the process environment is readyble by standard unix tools too as long as one has access to the process. So storing an unencrypted token in process env makes it more likely that it can be read by foreign code.

  Best solution would be to try to eleminate process.env usage totally. If this is too complex pleas eblock compact mode at least.

  Looking at the code I cannot see any place where the env varaibles are set. So simply removing them should do the reuired cleanup.
  see https://github.com/inventwo/ioBroker.life360ng/blob/c17da17b031536762282d3719209afb9e0c0d7b8/lib/life360CloudConnector.js#L112

- [x] dead code / example code commented out

  Please consider to remove unused / comment out code and examples at main.js. see https://github.com/inventwo/ioBroker.life360ng/blob/c17da17b031536762282d3719209afb9e0c0d7b8/main.js#L45

- [x] consider using standard function this.delay() provided by adapter package instead of private sleep().

  This is a remark only and not blocking. It's your decision how to proceed.

- [x] invalid characters should be filtered from object ids
  
  ```
    function name2id(pName) {
  }
  ```
   There's already a function getSantinizedObjectName (https://github.com/inventwo/ioBroker.life360ng/blob/c17da17b031536762282d3719209afb9e0c0d7b8/lib/life360DbConnector.js#L90) but this function does not remove all suggested characters. At least spaces should be remove too as they are known to cause problems with vis implementations. But german Umlaute cause troibles within some scenaroiuos too. The safest way is to limit object ids to [a-zA-Z0-9_-]. Please that at least all characters listed at FORBIDDEN_CHARS, ". " and " " are removed.

- [x] unused onStateChange handler

- [x] unused onObjectChange handler

  You configred an onObjectChange handler but it looks like this handler does not do any important tasks. Please remove the handler if the adapter does not react on object changes.

- [x] invalid object hirarchy

  Objects must be sorted as device - channel - state. 
  Devices must not be located below any other device or channel.
  Channels should nornally have no children of type channel, but nested channels afre allowed if this makes sense.
  Folders are allowed anywhere in between (except as children of states)
  Objects of type state must not have any children (OK)

  'Places' is a channel. But it's child 'Refugium...' is a device. This is invalid.
  
- [x] consider using adapter.setTimeout / this.setTimeout instead of (standard) setTimeout
  
- [x] consider replacing setIntervall by setTimeout

  When using setIntervall you must ensure that every processing cycle is shorter than the interval. This is difficult when using external request (i.e http requests) unless you can set a timeout fpor the external operation which in total (if more than one external operation occures during an interval) is shorter than the interval value.

  A safe way is to start the interval and at the end of the porvessing loop restart the processing using setTimeout. Especially if th intervall doe not be very exact this is a safe way.

  If some external processing lasts to long when using setIntervall multiple processings could be started in parallel (a new one is started befor the last one has finished) which in many cases slows down processing and hence more an more paralle proessing loops are started in parallel. This can cause serious memeory problems and finally crash the compe ioBroker host. 

  So ensuring that interval processing never lasts to long is important.

McM1957

**Please attach a new object dump after fixig the object structure.**

reminder 5.5.2026

skvarel:
Pls find att. the new object hierarchy:

[life360ng.0(new).json](https://github.com/user-attachments/files/26959733/life360ng.0.new.json)

mcm1957:
lgtm
Thanks for fixing and revamp the adapter

mcm1957:

PR #5855, opened 2026-04-16
Update openligadb to 1.9.3

oweitman:
Please update my adapter ioBroker.openligadb to version 1.9.3.


mcm1957:
@oweitman 
Any reason why windows and mac-os are missing from testing matrix?
I cannot imagine that this adapetr could be platform dependent and I did not notice such a not at readme.
Please let me know of there is any reason for this.

oweitman:
Okay, it seems to have been lost in one of the recent adjustments.
workflow changed

PR #5860, opened 2026-04-18
Update squeezeboxrpc to 1.6.4

oweitman:
Please update my adapter ioBroker.squeezeboxrpc to version 1.6.4.

### DIFF
--- a/sources-dist-stable.json
+++ b/sources-dist-stable.json
@@ -2797,7 +2797,7 @@
     "meta": "https://raw.githubusercontent.com/oweitman/ioBroker.squeezeboxrpc/main/io-package.json",
     "icon": "https://raw.githubusercontent.com/oweitman/ioBroker.squeezeboxrpc/main/admin/squeezeboxrpc.png",
     "type": "multimedia",
-    "version": "1.6.2"
+    "version": "1.6.4"
   },
   "srm": {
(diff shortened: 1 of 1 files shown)


mcm1957:
@oweitman 
Are you sure those packages are not used?

👀 [W5042] Package "aws-sdk" is used in source files but not found in dependencies of [package.json](https://github.com/oweitman/ioBroker.squeezeboxrpc/blob/master/package.json). Dependency might be missing.
 👀 [W5042] Package "tunnel-ssh" is used in source files but not found in dependencies of [package.json](https://github.com/oweitman/ioBroker.squeezeboxrpc/blob/master/package.json). Dependency might be missing.

oweitman:
Yes, I've been using the Squeezenode package since the adapter's launch (2019). Due to dependency issues, I've included the package's source code directly.

The AWS and SSH components have never been actively used, and there's no user-configuration option to select them.

I'd prefer not to modify the original source code. Another alternative would be to include the packages in the package.json file to satisfy the rule.

mcm1957:
OK - I understand. Its related to an EXTERNAL package nd I can understand that you do not want to modify an external code.

I'll evalute if I can improve. As its a warning only, lets ignore it for now. Thans for fast response.

PR #5864, opened 2026-04-19
Update bmw  to 4.3.5

hombach:


mcm1957:
@hombach 
Did you deploy the npm package manually? The package is not signed by npm like i.e. shelly packages:

image

Seems you do not use trusted deploy yet. Please consider using it in future to ensiute that the npm package really relates rto the corresponding githuib repo and tags.

PR #5908, opened 2026-04-25
Update Openknx stable 1.1.16

TA2k:

PR #5910, opened 2026-04-26
Update openknx to 1.1.16

mcm1957:
replace #5908


TA2k:
The old PR has a wrong title 1.1.16. That's lead to you Manual change to 1.1.16 but the correct version is 1.1.6
Please change this as a emergency change

PR #5938, opened 2026-05-03
Fix incorrect icon URL for nextcloudtalk

Rello:
correct master->main


Rello:
This PR is fixing E4005

PR #5943, opened 2026-05-05
Move admin to stable and remove chai from deps

GermanBluefox:


mcm1957:
Please do not mix adapter versionnupdates and changescto code at iobroker.repositories at ONE Pr.

create ONE seperste PR for every Adapter which should be added or whoch version should be bumped at stable. 

Do not update more than one Adapter within one PR.

Code changes need a seperste PR in all cases.

PR #5944, opened 2026-05-05
Remove chai from deps

GermanBluefox:


mcm1957:
reminder 8.5.2026
Will be nerged and checked after my holidays

PR #5950, opened 2026-05-07
Add tidy to latest

skvarel:
Please add my adapter ioBroker.tidy to latest.

*This pull request was created by https://www.iobroker.dev 14a3068.*

### DIFF
--- a/sources-dist.json
+++ b/sources-dist.json
@@ -1105,11 +1105,6 @@
     "icon": "https://raw.githubusercontent.com/Apollon77/ioBroker.homee/master/admin/homee.png",
     "type": "iot-systems"
   },
-  "hoymiles": {
-    "meta": "https://raw.githubusercontent.com/Eistee82/ioBroker.hoymiles/main/io-package.json",
-    "icon": "https://raw.githubusercontent.com/Eistee82/ioBroker.hoymiles/main/admin/hoymiles.png",
-    "type": "energy"
-  },
   "homekit-controller": {
     "meta": "https://raw.githubusercontent.com/Apollon77/ioBroker.homekit-controller/main/io-package.json",
     "icon": "https://raw.githubusercontent.com/Apollon77/ioBroker.homekit-controller/main/admin/homekit-controller.png",
@@ -1129,6 +1124,11 @@
     "meta": "https://raw.githubusercontent.com/TA2k/ioBroker.hoover/main/io-package.json",
     "icon": "https://raw.githubusercontent.com/TA2k/ioBroker.hoover/main/admin/hoover.png",
     "type": "household"
+  },
+  "hoymiles": {
+    "meta": "https://raw.githubusercontent.com/Eistee82/ioBroker.hoymiles/main/io-package.json",
+    "icon": "https://raw.githubusercontent.com/Eistee82/ioBroker.hoymiles/main/admin/hoymiles.png",
+    "type": "energy"
   },
   "hoymiles-ms": {
     "meta": "https://raw.githubusercontent.com/mcm4iob/ioBroker.hoymiles-ms/main/io-package.json",
@@ -2886,6 +2886,11 @@
     "icon": "https://raw.githubusercontent.com/hombach/ioBroker.tibberlink/master/admin/tibberlink.png",
     "type": "energy"
   },
+  "tidy": {
+    "meta": "https://raw.githubusercontent.com/inventwo/ioBroker.tidy/main/io-package.json",
+    "icon": "https://raw.githubusercontent.com/inventwo/ioBroker.tidy/main/admin/tidy.svg",
+    "type": "storage"
+  },
   "tileboard": {
     "meta": "https://raw.githubusercontent.com/ioBroker/ioBroker.tileboard/master/io-package.json",
     "icon": "https://raw.githubusercontent.com/ioBroker/ioBroker.tileboard/master/admin/tileboard.png",
@@ -3060,6 +3065,11 @@
     "icon": "https://raw.githubusercontent.com/TA2k/ioBroker.vesync/main/admin/vesync.png",
     "type": "climate-control"
   },
+  "victron-cerbo": {
+    "meta": "https://raw.githubusercontent.com/ioBroker/ioBroker.victron-cerbo/main/io-package.json",
+    "icon": "https://raw.githubusercontent.com/ioBroker/ioBroker.victron-cerbo/main/admin/victron-cerbo.svg",
+    "type": "energy"
+  },
   "viessmann": {
     "meta": "https://raw.githubusercontent.com/misanorot/ioBroker.viessmann/master/io-package.json",
     "icon": "https://raw.githubusercontent.com/misanorot/ioBroker.viessmann/master/admin/viessmann.png",
@@ -3073,11 +3083,6 @@
   "virtualpowermeter": {
     "meta": "https://raw.githubusercontent.com/Omega236/ioBroker.virtualpowermeter/master/io-package.json",
     "icon": "https://raw.githubusercontent.com/Omega236/ioBroker.virtualpowermeter/master/admin/virtualpowermeter.png",
-    "type": "energy"
-  },
-  "victron-cerbo": {
-    "meta": "https://raw.githubusercontent.com/ioBroker/ioBroker.victron-cerbo/main/io-package.json",
-    "icon": "https://raw.githubusercontent.com/ioBroker/ioBroker.victron-cerbo/main/admin/victron-cerbo.svg",
     "type": "energy"
   },
   "vis": {


skvarel:
[tidy.0.json](https://github.com/user-attachments/files/27470150/tidy.0.json)

mcm1957:
reminder 21.5.2026

skvarel:
https://github.com/inventwo/ioBroker.tidy/issues/11

skvarel:
RE-CHECK!

mcm1957:
Checklist Version:** V1.3  
**Review Date:** 2026-06-10  
**Reviewed Files:** main.js, io-package.json, package.json, README.md, admin/jsonConfig.json, admin/i18n/*.json (11 languages)

---

## ✅ Passed Checks

- [x] 🟢 **README language** — README.md is written in pure English. No mixed-language content detected. No German-only README variant exists (not required but acceptable if present).

- [x] 🟢 **Sensitive information not logged** — `io-package.json` defines no `protectedNative` or `encryptedNative` entries. No sensitive config values exist that could be inadvertently logged.

- [x] 🟢 **Object ID sanitization** — Channel IDs derived from user-configured path names are run through `sanitizeName()` in `main.js` which strips all characters except `[a-zA-Z0-9_-]` and lowercases the result. This is stricter than the minimum `FORBIDDEN_CHARS` requirement and is compliant.

- [x] 🟢 **State IDs use English wording** — All state IDs created by the adapter (`trigger`, `result`, `lastScan`, `count`, `deadCount`, `staleCount`, `orphanedCount`, `exceptionCount`, `complete`) use English-only terms.

- [x] 🟢 **Log messages in English** — All `this.log.*` calls in `main.js` use English text with no translation mechanism applied. Examples: `"Starting Tidy adapter..."`, `"Scanning path: ..."`, `"Manual scan triggered for ..."`.

- [x] 🟢 **No axios/HTTP calls** — The adapter does not make any HTTP/HTTPS calls; no timeout check is required.

- [x] 🟢 **No access to cloud services** — The adapter only reads from the local ioBroker object/state database. No cloud polling, no fixed-timestamp cloud access, and no random-delay requirement applies.

- [x] 🟢 **setInterval used via adapter wrapper** — The adapter uses `this.setInterval(...)` (the ioBroker wrapper), not the raw Node.js `setInterval`. The wrapper is safe with respect to unload handling.

- [x] 🟢 **Unload handler clears interval** — `onUnload` correctly clears `this.scanInterval` and calls the callback in all branches, including a try/catch guard.

- [x] 🟢 **onStateChange honors ack flag** — The handler in `main.js` checks `!state.ack` before processing trigger button presses (`id.endsWith('.trigger')`). Own states triggered by the adapter itself (written with `ack: true`) are therefore ignored correctly.

- [x] 🟢 **i18n files cover all required languages** — All eleven required languages (`en`, `de`, `ru`, `pt`, `nl`, `fr`, `it`, `es`, `pl`, `uk`, `zh-cn`) are present as separate files under `admin/i18n/`.

- [x] 🟢 **All names and descriptions use i18n or English** — `common.desc` in `io-package.json` uses a full i18n object covering all 11 languages. `common.titleLang` covers all 11 languages. `instanceObjects[0].common.name` also uses a full i18n object.

- [x] 🟢 **No code modifies npm package directories** — No file writes to `node_modules` or other package-managed paths are present.

- [x] 🟢 **No dead code detected in core logic** — The primary execution paths (`onReady`, `onStateChange`, `scanPath`, `scanComplete`, `analyzeDatapoint`) are all reachable and used. The commented-out `onMessage` block is intentionally scaffolded and not dead code.

- [x] 🟢 **configurable scanInterval limited in UI** — `jsonConfig.json` sets `min: 1, max: 168` on `scanInterval`. `daysUntilStale` and `daysUntilDead` both have `min: 1, max: 3650`.

---

## ❌ Issues Requiring Action

- [x] 🔴 **`buildExceptionSets()` is dead code / duplicate of `loadExceptionSets()`**

  `main.js` contains two methods that serve the same purpose: `loadExceptionSets()` (async, calls `getForeignObjectAsync` for type detection) and `buildExceptionSets()` (sync, no object lookup). `buildExceptionSets()` is only called from `isExcluded()` as a lazy fallback when `_exceptionExact` is `null`. In normal execution, `loadExceptionSets()` is always called before `isExcluded()`, so `buildExceptionSets()` will never actually be invoked. If `isExcluded()` is ever called before `loadExceptionSets()` (e.g. in an unexpected code path), `buildExceptionSets()` will silently produce incomplete exception sets (missing the async object-type lookup). The dead/fallback method should either be removed or the guard in `isExcluded()` should be replaced with a clear error/warning.

  - Reference: `main.js`, `buildExceptionSets()` method (approx. line 430–455) and `isExcluded()` guard (approx. line 462–464).

- [x] 🔴 **`scanInterval` value is not validated/clamped in code before use with `setInterval`**

  `jsonConfig.json` sets UI limits (`min: 1, max: 168`) for `scanInterval`, but the code in `onReady` does not validate or clamp the value before computing `intervalMs = this.config.scanInterval * 60 * 60 * 1000`. If a value is provided outside of this range (e.g. via direct API write to the instance object), a value of 0 or negative would result in `setInterval` being called with an invalid delay, and a value exceeding ~596 hours would overflow the 32-bit setTimeout limit (2,147,483,647 ms). The checklist explicitly states that UI-level limits are not sufficient.

  **Recommended fix:** Add a runtime guard before computing `intervalMs`:
  ```js
  const clampedHours = Math.max(1, Math.min(168, this.config.scanInterval));
  const intervalMs = clampedHours * 60 * 60 * 1000;
  ```
  - Reference: `main.js`, `onReady()` (approx. line 57–65).

- [x] 🔴 **`daysUntilStale` and `daysUntilDead` are not validated in code**

  Similar to `scanInterval`, both values are used directly from `this.config` in `analyzeDatapoint()` without any runtime range check. A value of 0 would cause every datapoint to appear dead or stale. A very large value (> ~24 days) multiplied through `now - state.ts` could theoretically overflow floating-point comparisons, though JavaScript handles large floats gracefully. Regardless, a zero or negative value is semantically invalid. The minimum useful check is:
  ```js
  const daysUntilStale = Math.max(1, this.config.daysUntilStale || 90);
  const daysUntilDead  = Math.max(1, this.config.daysUntilDead  || 365);
  ```
  - Reference: `main.js`, `analyzeDatapoint()` (approx. line 490–491).

- [x] 🟡 **Unused i18n translation keys detected in `admin/i18n/en.json` (and other language files)**

  The `en.json` i18n file contains several keys that do not appear to match any string currently used as a label, help text, or tooltip in `jsonConfig.json`. These appear to be leftover keys from previous UI iterations. Examples of likely orphaned keys:

  - `"Pick an adapter instance or subfolder (e.g. 0_userdata.0, alias.0). The path stays editable — remove the instance number (alias.0 → alias) to scan all instances."` — not referenced in current `jsonConfig.json`
  - `"Pick a folder or namespace (e.g. alias, 0_userdata.0, alias.0.Heizungen). Individual datapoints are not shown. The path stays editable — remove the instance number (alias.0 → alias) to scan all instances."` — not referenced
  - `"Datapoints listed on the Exceptions tab are omitted ... Pick a single state to exclude one datapoint, or pick a folder/channel..."` (the version ending with `\n\n`) — appears to be a superseded variant
  - `"For German tables: Use status_de and issue_de columns\nFor English tables: Use status and issue columns..."` (the version ending with `\n\n`) — appears superseded

  These orphaned keys exist across all 11 language files. While they cause no runtime issue, they increase maintenance overhead and can cause confusion. They should be pruned using `translate-adapter` or manually.

  - Reference: `admin/i18n/en.json` (multiple keys), mirrored in all other language files.

- [x] 🟡 **`last_ts_iso` field uses German locale hardcoded (`de-DE`) for date formatting**

  In `analyzeDatapoint()`, the ISO timestamp is formatted using `new Date(state.ts).toLocaleString('de-DE')`. This produces German-formatted dates regardless of the ioBroker system language or the user's locale. The adapter's output JSON (`result` state) is intended for use in VIS table widgets, and the README even documents separate `status_de`/`issue_de` fields for German vs English display. However, a hardcoded `de-DE` locale for `last_ts_iso` means English users will always see German date formatting (e.g. `10.6.2026, 14:30:00` instead of `6/10/2026, 2:30:00 PM`).

  **Recommendation:** Either use `this.language` to select the locale dynamically, use ISO 8601 format (`new Date(state.ts).toISOString()`), or document the German-only format explicitly in the README.

  - Reference: `main.js`, `analyzeDatapoint()` (approx. line 508).

---

mcm1957:

- [x] issues reported by claude based checker must be fixed

- [x] errors reported by adapter checker must be fixed

  Please cleanup translation files

Please note that a later manual review could add additionla remarks.
  
McM1957

reminder 24.6.2026

skvarel:
Hi @mcm1957,

all review points have been fixed and released as v0.2.4:

- dead code buildExceptionSets() removed
- runtime validation for scanInterval, daysUntilStale, daysUntilDead
- i18n cleanup + checker issues resolved
- last_ts_iso now ISO 8601

Ready for re-review.

Thanks, skvarel

mcm1957:
reminder 18.6.2026

mcm1957:
lgtm

Just one note - not sure if subscribing to system object of the adpater makes much sense as changing the configuration of the adapter anyway restarts the instance.

mcm1957:

PR #6067, opened 2026-06-02
Update stable version for kostal-piko-ba to 7.0.2

hombach:


mcm1957:
Please correct invalid state roles for next release.

PR #6110, opened 2026-06-06
Update backitup to 3.3.17

simatec:


simatec:
E5050] process.exit() used in (lib/restore.js). This is incompatible with compact mode as it terminates the entire host process. Remove process.exit() or set common.compact to false at [io-package.json](https://github.com/simatec/ioBroker.backitup/blob/master/io-package.json). is not a error!!

Please push the Update

mcm1957:
I do not understand the request.

E5050 must be fixed at backitup. I cannot push anything there.

mcm1957:
OK I see. Did not notive the last words of your statement.

Using process.exit with common.compact = true IS an error in an adapter as process.exit will terminate the process for all other instances too. lib directory is typically part of the adapter and not part of any external scripts so the content of lib is evaluated.

Backitup seems to be a special situation due to the restore process. I will evaluate if and how detection could be improved. Fastest solution would be to set common.compact=false as in my opinion backitup should not execute in compact mode anyway.  Don't know if this would by a problem anywhere.

PR #6205, opened 2026-06-24
Update alarm version to 4.0.5

misanorot:
I've been trying to fix the issues flagged by the adapter checker for weeks now, but new ones keep popping up. I really want to get this into stable, but the adapter checker predicts several issues that prevent this—even though some of them don't even actually exist.

Furthermore, the adapter checker on GitHub flags different errors compared to when I run the check via the ioBroker Developer Portal. The checker there reports errors that are simply incorrect, which is why I can't submit it directly to stable from there either.

### DIFF
--- a/sources-dist-stable.json
+++ b/sources-dist-stable.json
@@ -85,7 +85,7 @@
     "meta": "https://raw.githubusercontent.com/misanorot/ioBroker.alarm/master/io-package.json",
     "icon": "https://raw.githubusercontent.com/misanorot/ioBroker.alarm/master/admin/alarm.png",
     "type": "alarm",
-    "version": "3.7.6"
+    "version": "4.0.5"
   },
   "alexa-shoppinglist": {
     "meta": "https://raw.githubusercontent.com/MiRo1310/ioBroker.alexa-shoppinglist/main/io-package.json",


mcm1957:
If you encounter a false positive or some other error with adapter checker please open a issue at iobrok.repochecker.
But the error and warning listed here are correct.

 ❗ [E6004] No ## License found in [README.md](https://github.com/misanorot/ioBroker.alarm/blob/master/README.md)
WARNINGS:

  Formatting is broken and so no header is visible:

  image

 👀 [W6016] [README.md](https://github.com/misanorot/ioBroker.alarm/blob/master/README.md) does not appear to be written in English. [README.md](https://github.com/misanorot/ioBroker.alarm/blob/master/README.md) must be written using English language only. Please translate.

  README must be written using pure english language. A german hera is not desiued and couzld cause problem at automatic translation. The link to german documentation is OK.
  
  image

Please PR https://github.com/misanorot/ioBroker.alarm/pull/418.

Differneces between iobroker.dev and repochecker issues / PRs are possible as dev portal is updated indipendentyl. If dev portal is using an older checker revision please add a issue at devportal repository.

mcm1957:
Note:
Fixing with NEXT release is ok. No new Release is required now - fixing README.ms at repo is completly sufficient.  The only "Problem" is that 4.0.5 is just 2 days old. It should be at least 5 to 7 days available at LATEST. I'll will keep in evicence so that you do not have extra work.

mcm1957:
replaced by https://github.com/ioBroker/ioBroker.repositories/pull/6230

PR #6290, opened 2026-07-10
Update alexa2 version to 3.28.3

Apollon77:


Apollon77:
This is a Bugfix release and so we will not change anything on nodejs support or such to get most users possible to have a working version again

PR #6291, opened 2026-07-10
Add Miner v1.0.4 to stable

SimonFischer04:
Manually created as there are still issues blocking dev portal tooling adapter can't fix. 
(Same as with the latest PR: https://github.com/AlCalzone/release-script/issues/130).

Also at least in my repo issue there is a false positive warning (besides a lot of "could not retrieve log for job ..."):
https://github.com/SimonFischer04/ioBroker.miner/issues/164:

"[W5060] Package "type-fest" is listed in dependencies of [package.json](https://github.com/SimonFischer04/ioBroker.miner/blob/master/package.json) but not imported or required by any scanned source file. Dependency might be unused."
Is used here: https://github.com/SimonFischer04/ioBroker.miner/blob/7d837be23d081285980b1a9915c1d87a7f2b8594/src/lib/MinerAdapterDeviceManagement.ts#L28


mcm1957:
@SimonFischer04 

ad [W5060] Package "type-fest" is listed in dependencies of [package.json](https://github.com/SimonFischer04/ioBroker.miner/blob/master/package.json) but not imported or required by any scanned source file. Dependency might be unused.

Looks like the import warning is correct.

> import type { PartialDeep } from 'type-fest';

is a type only import (unless I miss something) - so this dependency should not be needed within dependencies. Adding it to devDependencies should be sufficient to allow compilation. (Please, check - but ist not blocking of course.)

SimonFischer04:
> @SimonFischer04
> 
> ad [W5060] Package "type-fest" is listed in dependencies of [package.json](https://github.com/SimonFischer04/ioBroker.miner/blob/master/package.json) but not imported or required by any scanned source file. Dependency might be unused.
> 
> Looks like the import warning is correct.
> 
> > import type { PartialDeep } from 'type-fest';
> 
> is a type only import (unless I miss something) - so this dependency should not be needed within dependencies. Adding it to devDependencies should be sufficient to allow compilation. (Please, check - but ist not blocking of course.)

Yeah, you're right.
I misread to warning as complaining about it being any dep at all (normal or dev).

Fixed now.

SimonFischer04:
RE-CHECK!

PR #6305, opened 2026-07-12
Update tibberlink stable version to 7.1.3

hombach:


hombach:
@mcm1957 Hi Martin, looks like "STABLE-BRAND-NEW tagging" is verifying age of latest repo commit, instead of latest repo release? Plus I don't want to push the latest release to stable... it's an earlier version, age >14 days....

PR #6002, opened 2026-05-20
Add blustream-acm to latest

AlanSRU:
**Checklist**
- [x] The adapter is published to NPM (iobroker.blustream-acm200@0.2.1)
- [x] Developer Best practices are known and considered
- [x] All requirements to bring a new adapter into Latest repository are fulfilled
- [x] The adapter was checked by https://adapter-check.iobroker.in/ — clean
      (one known false-positive E0063 for chai/mocha, which are required because npm doesn't hoist @iobroker/testing's nested mocha into node_modules/.bin)
- [x] Forum testing thread: https://forum.iobroker.net/topic/84592/new-blustream-acm200-hdmi-matrix-controller-adapter

This PR adds the Blustream ACM200 adapter to the latest repository.

**About the adapter**
- Controls the Blustream ACM200 HDMI matrix controller for AV-over-IP distribution
- Discovers connected transmitters and receivers via the ACM200's telnet interface
- Supports video/audio routing — combined and independent per stream — and route-to-all-displays bulk commands
- Transmitter audio source selection (AUTO/HDMI/ANA)
- Built on @iobroker/adapter-core 3.3.x, requires js-controller 6.0.11+, admin 7.6.20+
- Admin UI uses jsonConfig; password stored encrypted (protectedNative + encryptedNative)
- CI uses ioBroker/testing-action-* and **npm trusted publishing (OIDC)** — verified working on the 0.2.1 release

**Links**
- GitHub: https://github.com/AlanSRU/ioBroker.blustream-acm200
- npm: https://www.npmjs.com/package/iobroker.blustream-acm200
- Forum thread: https://forum.iobroker.net/topic/84592/new-blustream-acm200-hdmi-matrix-controller-adapter

The ACM200 is part of the Blustream AV control family; same author also maintains iobroker.blustream-mfp (PR #6001).

PR #6001, opened 2026-05-20
Add blustream-mfp to latest

AlanSRU:
**Checklist**
- [x] The adapter is published to NPM (iobroker.blustream-mfp@0.3.3)
- [x] Developer Best practices are known and considered
- [x] All requirements to bring a new adapter into Latest repository are fulfilled
- [x] The adapter was checked by https://adapter-check.iobroker.in/ — clean
      (one known false-positive E0063 for chai/mocha, which are required because npm doesn't hoist @iobroker/testing's nested mocha into node_modules/.bin)
- [x] Forum testing thread: https://forum.iobroker.net/topic/84590/new-blustream-mfp-av-presentation-switcher-adapter

This PR adds the Blustream MFP adapter to the latest repository.

**About the adapter**
- Controls Blustream AMF/MFP/WMF series AV presentation switchers
- Supports IP/Telnet and RS232 serial connections (with optional RS232-to-IP converter for older models)
- Dynamic state tree based on the selected device model (AMF42AU, MFP62, MFP72, MFP112, WMF51, WMF72)
- Built on @iobroker/adapter-core 3.3.x, requires js-controller 6.0.11+, admin 7.6.20+
- Admin UI uses jsonConfig
- Published via npm trusted publishing (OIDC) from GitHub Actions

**Links**
- GitHub: https://github.com/AlanSRU/ioBroker.blustream-mfp
- npm: https://www.npmjs.com/package/iobroker.blustream-mfp
- Forum thread: https://forum.iobroker.net/topic/84590/new-blustream-mfp-av-presentation-switcher-adapter

The adapter has been running in production with 2 instances for several months and was verified again with the 0.3.3 release on js-controller 7.0.7.

PR #6342, opened 2026-07-18
Add blustream-acm to latest

AlanSRU:
**Checklist**
- [x] The adapter is published to NPM (iobroker.blustream-acm@0.3.1)
- [x] Developer Best practices are known and considered
- [x] All requirements to bring a new adapter into Latest repository are fulfilled
- [x] Checked with `@iobroker/repochecker` — clean apart from `W4001` (not yet in latest, i.e. this PR)
- [x] `bluefox` added as an npm collaborator
- [x] Forum testing thread: https://forum.iobroker.net/topic/84592/new-blustream-acm200-hdmi-matrix-controller-adapter

> **Note:** This PR **supersedes #6002** and is opened fresh at @mcm1957's request because the adapter was **renamed `blustream-acm200` → `blustream-acm`** to reflect multi-model support. The old npm package `iobroker.blustream-acm200` has been deprecated and the GitHub repo renamed accordingly. #6002 will be closed with a pointer here.

This PR adds the **Blustream ACM** matrix controller adapter to the latest repository.

**About the adapter**
- Controls Blustream ACM matrix controllers for AV-over-IP distribution, with a configurable **Controller Model** (ACM200 / ACM210 / ACM500 / ACM1000) — states and commands are model-aware via a capability table
- Discovers connected transmitters and receivers via the controller's telnet interface
- Video/audio routing (combined and independent per stream) plus route-to-all-displays bulk commands
- Breakaway routing (IR / RS232 / USB / CEC), output power/mute, Dante/analogue/HDMI audio matrix and ARC control (model-gated)
- Transmitter audio source selection (AUTO/HDMI/ANA)
- Built on @iobroker/adapter-core 3.4.x, requires js-controller 6.0.11+, admin 7.6.20+
- Admin UI uses jsonConfig; **no credentials** — the telnet interface requires no login (`protectedNative`/`encryptedNative` empty)
- CI uses ioBroker/testing-action-* and **npm trusted publishing (OIDC)** — 0.3.1 published with provenance

**Prior review findings (from #6002, v0.2.4) are addressed in 0.3.1**
- Per-device `connected` states use `indicator.reachable`; `.id` states use role `text`
- Orphaned `username`/`password` i18n keys removed across all 11 languages; preview-URL separator `?time=` → `&time=` fixed

**Links**
- GitHub: https://github.com/AlanSRU/ioBroker.blustream-acm
- npm: https://www.npmjs.com/package/iobroker.blustream-acm

Same author also maintains iobroker.blustream-mfp (PR #6001). A fresh object dump (with a device, serials masked) and **READY FOR RE-REVIEW** will follow.

### DIFF
--- a/sources-dist.json
+++ b/sources-dist.json
@@ -315,6 +315,11 @@
     "icon": "https://raw.githubusercontent.com/Uwe1958/ioBroker.bluesound/main/admin/bluesound.png",
     "type": "multimedia"
   },
+  "blustream-acm": {
+    "meta": "https://raw.githubusercontent.com/AlanSRU/ioBroker.blustream-acm/main/io-package.json",
+    "icon": "https://raw.githubusercontent.com/AlanSRU/ioBroker.blustream-acm/main/admin/blustream-acm.png",
+    "type": "multimedia"
+  },
   "blustream-mfp": {
     "meta": "https://raw.githubusercontent.com/AlanSRU/ioBroker.blustream-mfp/main/io-package.json",
     "icon": "https://raw.githubusercontent.com/AlanSRU/ioBroker.blustream-mfp/main/admin/blustream.png",


AlanSRU:
[blustream-acm.0.json](https://github.com/user-attachments/files/30146554/blustream-acm.0.json)

AlanSRU:
READY FOR RE-REVIEW

Object dump attached above (`blustream-acm.0.json`) from a live install with **9 transmitters and 31 receivers**. It confirms the object-structure fixes from the previous review:
- per-device `connected` states now use `indicator.reachable`
- `.id` states now use role `text`

Automated checker is green (`auto-checked ✔`, no errors). Thanks @mcm1957!

mcm1957:
due to holidays processing delayed at least until 3.8.32026
reminder 8.8.2026

mcm1957:
reminder 3.8.2026

mcm1957:
# Adapter Review Report

**Adapter:** `ioBroker.blustream-acm`  
**Repository:** https://github.com/AlanSRU/ioBroker.blustream-acm  
**Adapter Version:** 0.3.1  
**Checklist Version:** V2.1  
**Review Date:** 2026-08-07  
**Object Dump:** `blustream-acm_0.json` — newest timestamp `ts: 1784351111353` → **2026-07-18 05:05:11 UTC** (recent, matches current code version 0.3.1 ✅)

---

## Summary

The adapter is well-structured, plain JavaScript, correctly using `@iobroker/adapter-core`, local TCP socket (Telnet) communication to Blustream ACM AV-over-IP matrix controllers. The object dump matches the current code. Most checks pass. A small number of issues require attention (all minor to medium severity).

---

## ✅ Passed Checks

- [x] 🟢 **README language** — README.md is written in pure English. A `README.de.md` equivalent is not present but also not required. No multi-language mixing detected.

- [x] 🟢 **Product / manufacturer link** — README.md contains links to the Blustream ACM200 product page (`https://www.blustream.com/product/acm200/`) and to the manufacturer (`https://www.blustream.com/`) as required for a device-specific adapter.

- [x] 🟢 **Unique adapter icon** — `admin/blustream-acm.png` (md5: `b6bc56eadd0ca956f134d377fefe6b86`) differs from the ioBroker creator icon (md5: `d41d8cd98f00b204e9800998ecf8427e`). Icon is unique. ✅

- [x] 🟢 **No GitHub install instructions in README** — README only references the ioBroker admin interface for installation. No `npm install` from GitHub or direct git clone instructions present.

- [x] 🟢 **Platform compatibility** — Adapter uses only `node:net` (cross-platform) and `@iobroker/adapter-core`. No OS-specific code. Works on Linux, Windows, macOS.

- [x] 🟢 **Sensitive information not logged** — No passwords, tokens, or secrets are used (no `protectedNative`/`encryptedNative` fields in `io-package.json`). Only `host` and `port` are logged, which are not sensitive.

- [x] 🟢 **Object ID sanitization** — Device IDs parsed from Telnet responses are validated by `sanitizeDeviceId()` (only allows `^\d+$`), then zero-padded. This prevents injection of forbidden characters. Hardcoded IDs (`system`, `transmitters`, `receivers`) are safe. ✅

- [x] 🟢 **State IDs use English wording** — All state IDs (`route`, `videoRoute`, `audioRoute`, `connected`, `previewUrl`, etc.) are in English. Confirmed in source and object dump.

- [x] 🟢 **State names/descriptions in English or i18n** — `common.name` values in `ensureTransmitterObjects` / `ensureReceiverObjects` are English strings. Admin UI labels use the i18n mechanism.

- [x] 🟢 **All writable states subscribed and handled** — `onStateChange` correctly handles all writable states:
  - `system.commands.refresh`, `system.commands.refreshAll` ✅
  - `system.commands.routeAll`, `routeAllVideo`, `routeAllAudio` ✅
  - `receivers.*.route`, `videoRoute`, `audioRoute` ✅
  - `receivers.*.irRoute`, `rs232Route`, `usbRoute`, `cecRoute` ✅
  - `receivers.*.power`, `mute`, `audioOutputMode`, `arcMode` ✅
  - `transmitters.*.audioSource`, `audioMatrixMode` ✅

- [x] 🟢 **State roles are valid** — All roles verified against the ioBroker state-roles reference:
  - `indicator.connected`, `indicator.reachable` — `write: false`, `type: boolean` ✅
  - `button` — `write: true`, `read: false`, `type: boolean` ✅
  - `switch.power`, `switch` — `write: true`, `type: boolean` ✅
  - `text`, `info.name`, `info.ip`, `info.firmware`, `info.mac`, `info.model`, `url`, `date` — all used correctly ✅
  - No invalid role/type/write-flag combinations found in dump or code.

- [x] 🟢 **Object hierarchy is valid** — Structure confirmed from dump and code:
  - `device` → `channel` → `state` (e.g. `system` → `system.status` → `system.status.connected`) ✅
  - `device` → `channel` → `state` (e.g. `receivers` → `receivers.001` → `receivers.001.route`) ✅
  - No `channel` → `device` violations found. ✅

- [x] 🟢 **Dynamically created objects use correct structure** — `ensureTransmitterObjects` / `ensureReceiverObjects` create `channel` under root `device`, with `state` children. Feature-gated states (ARC, audioMatrix, IR/RS232/USB/CEC, power, mute) are correctly only created for models that support them. Object dump (ACM200 model) confirms no gated states are present. ✅

- [x] 🟢 **Log messages in English** — All `this.log.*` calls use English text. No translation mechanism (`_('...')` or similar) used in log messages. ✅

- [x] 🟢 **Admin UI labels use i18n** — `jsonConfig.json` sets `"i18n": true` and uses key strings matched to `admin/i18n/` translations. ✅

- [x] 🟢 **All 11 ioBroker languages present in i18n** — All required languages (`en`, `de`, `ru`, `pt`, `nl`, `fr`, `it`, `es`, `pl`, `uk`, `zh-cn`) are present in `admin/i18n/` with matching key sets. No extra or missing languages or keys. ✅

- [x] 🟢 **Plain JavaScript structure** — Sources at root (`main.js`) and `/lib/` (`lib/models.js`). No `/build/` or `/src/` directories. Correct for a plain JS adapter. ✅

- [x] 🟢 **onObjectChange handler absent** — No `onObjectChange` handler is registered or referenced. No stale handler present. ✅

- [x] 🟢 **onStateChange honors ack flag** — `onStateChange` processes only `!state.ack` (line ~293 `if (state && !state.ack)`). All own state updates use `setState(..., true)`. ✅

- [x] 🟢 **No axios / HTTP calls** — All communication uses `node:net` TCP socket. No axios or other HTTP client used; the timeout requirement for HTTP calls is therefore not applicable.

- [x] 🟢 **Configurable timeouts validated and clamped** — `validateNumber()` is used for `port` (1–65535), `pollInterval` (1000–3,600,000 ms), and `timeout` (500–60,000 ms). Values are clamped and logged if out of range. Node's max timer value (2,147,483,647 ms) is respected since the maximum allowed is well below that limit. ✅

- [x] 🟢 **No dead code detected** — All config parameters (`model`, `host`, `port`, `pollInterval`, `timeout`) are actively used. No obviously unreachable code paths found.

- [x] 🟢 **No unused i18n keys** — All keys in `en.json` correspond to actual labels/help strings in `jsonConfig.json`. No orphaned translation keys.

- [x] 🟢 **No npm install scripts** — `package.json` contains no `preinstall`, `postinstall`, `install`, or `prepare` scripts. Only standard development scripts present. ✅

- [x] 🟢 **No code modifications in npm directories** — Adapter does not write to any directories that would be delivered via npm package at runtime.

- [x] 🟢 **Adapter is a daemon (not scheduled adapter)** — `io-package.json` sets `"mode": "daemon"`. Polling uses `setTimeout`-loop, not `setInterval`. The random-delay requirement for cloud-accessed scheduled adapters does not apply.

- [x] 🟢 **Reconnect logic implemented** — On disconnect, `cleanup(true)` schedules a reconnect after `reconnectDelay` (30 s). Heartbeat detects dead connections and triggers cleanup. ✅

- [x] 🟢 **Object dump matches current code** — The dump (2026-07-18, version 0.3.1) aligns with the current code. ACM200-only states present, no gated states for higher models. ✅

---

## ❌ Failed Checks / Issues Requiring Action

- [ ] 🟡 **Scheduled refresh fires at a fixed daily timestamp (3 AM)**
  `setupScheduledRefresh()` calculates the next occurrence of 03:00:00 local time and schedules `setTimeout` for exactly that moment. This means all adapter instances across all hosts perform a full refresh simultaneously at 3 AM every day.
  Although this adapter connects to a **local LAN device** (not a cloud service), the checklist rule states: *"access to cloud services must not occur at fixed timestamps"*. The spirit of the rule — avoiding coordinated load spikes — applies here too, especially in multi-instance setups (multiple controllers). A small randomised offset (e.g. ±15 minutes) should be added.

  **Reference:** `main.js` line 1233: `nextRefresh.setHours(3, 0, 0, 0);`

(mcm1957) see below

- [ ] 🟡 **Heartbeat interval is not user-configurable and its value is hardcoded with a stale comment**
  `this.heartbeatInterval` is set to `10000` ms (10 s) with the comment `// Reduce to 10 seconds`. This "reduce from X" phrasing suggests it was changed from a larger value and the comment was not updated. While a fixed heartbeat is acceptable, the comment is misleading and should be cleaned up (e.g. `// Heartbeat every 10 s`).

  **Reference:** `main.js` line 74: `this.heartbeatInterval = 10000; // Reduce to 10 seconds`

- [ ] 🟡 **`setTimeout` loop for polling could race if processing takes longer than interval**
  `startPolling()` re-arms itself at the end of each cycle, so the effective delay is `pollInterval` **plus** the time taken to complete the poll. This is actually the preferred pattern mentioned in the checklist ("suggest to use setTimeout at the end of a processing loop instead of using setInterval"). The implementation is correct in that regard. **However**, the total duration of `refreshDeviceStatus()` plus subsequent `executeCommand('STATUS')` must complete within `pollInterval`. For very short configured values (minimum 1,000 ms), the socket command timeout (minimum 500 ms) leaves only 500 ms headroom. Consider raising the minimum `pollInterval` to at least `timeout * 2` (or document this constraint) to prevent overlapping poll cycles.

  **Reference:** `main.js` lines 1136–1185 (`startPolling`) and `validateNumber` call at line 208–213.

- [ ] 🟡 **`system.commands.refresh` and `system.commands.refreshAll` are both present with overlapping function**
  Two near-identical button states exist: `system.commands.refresh` (triggers basic STATUS poll) and `system.commands.refreshAll` (triggers full device detail refresh). Both are subscribed and handled. Their distinction is valid, but the README only mentions `refreshAll` implicitly and neither command is documented clearly enough for users. This is a minor UX/documentation issue, not a code defect.

  **Reference:** `main.js` lines 248–260 (`onReady` object creation) and `README.md` section "States → System".

---

## Additional Observations

- [ ] The `processResponse` method uses `statusBuffer` and `collectingStatus` instance properties that are never declared in the constructor (unlike the analogous `txInfoBuffer`/`rxInfoBuffer`/`collectingTxInfo`/`collectingRxInfo`). This works in JavaScript due to dynamic property assignment but is inconsistent with the adapter's coding style and could confuse linters.
  **Reference:** `main.js` lines ~935–960 (`processResponse`) — `this.statusBuffer` / `this.collectingStatus` not declared in constructor.

- [ ] The `reconcileModelStates()` method uses `this.delForeignObjectAsync()` which requires the full namespace prefix. This is correct but worth noting: if a different adapter instance were to share similar state names, this could theoretically affect foreign objects. The current implementation uses `this.getForeignObjectsAsync(this.namespace + '.receivers.*')` which is correctly scoped.

- [ ] `README.md` contains `### __WORK IN PROGRESS__` section and a duplicate `### __WORK IN PROGRESS__` entry in the Changelog. These should be cleaned up before a stable release.
  **Reference:** `README.md` lines 125–127.

- [ ] The object dump includes receivers `001` through `030` and `032` (skipping `031`), and transmitters `001` through `009`. This appears to reflect the actual connected hardware and is not an adapter defect.

- [ ] The `previewUrl` state uses role `url` which is valid. However, every poll cycle regenerates and sets the previewUrl (with a timestamp parameter). This causes frequent state-change events for this state even if the source has not changed. Consider setting the previewUrl only when the source transmitter changes.

- [ ] `jsonConfig.json` panel labels use string literals (`"Main settings"`, `"Advanced"`) rather than i18n keys in some places (the `label` field of tabs). These strings **are** present in `en.json` and all translated files, so functionality is correct. No action needed.

- [ ] `io-package.json` does not define `protectedNative` or `encryptedNative`. Correct — no sensitive fields exist in the native config.

- [ ] The adapter correctly sets `"connectionType": "local"` and `"dataSource": "poll"` in `io-package.json`, accurately describing the communication pattern.

mcm1957:
Remark related to 3:00 action is only cosmetic for a local adapter.
BUT I'm not sure if this adapter supports multiple instances at all. If it does not support multiple instances (due to auto detection) it should specify singleton attrobite at io-package.json to block creating multiple instances. See io-package.json schema.

None of the remarks is blocking - so lets go on.

mcm1957:

Please create a thread at https://forum.iobroker.net/category/91/tester and use a title like 

### "Test Adapter _adaptername_ Vx.x.x" 

to collect some user feedback and provide a link to this topic when requesting addition to stable repository later. Using a title like this (starting with "Test adapter xxxx") helps to detect this topic using the serach functionality of the forum.

Note: If an other testing topic already exists, it' OK to continue using this topic too. But please adapt title if it does not meet the structure mentioned above.

AlanSRU:
Thanks for the review and for releasing to latest, @mcm1957.

**Multiple instances (your remark on `singleton`):** the adapter does support multiple instances, so no `singleton` / `singletonHost` attribute is appropriate. There is no auto-detection — the controller IP, port and model are all entered manually per instance, and each instance keeps its whole object tree under its own namespace. Running one instance per ACM controller is an expected setup, which is also why the fixed-3-AM remark was worth acting on rather than dismissing.

**v0.3.2 is now published** and addresses every item from the report:

- Nightly full refresh is randomised over 02:45–03:15 instead of firing at exactly 03:00. Noted it is cosmetic for a local device, but it is correct for multi-instance installs, so it is fixed.
- Stale `// Reduce to 10 seconds` heartbeat comment corrected.
- `pollInterval` is now floored at `max(1000, timeout * 2)` in code, and the constraint is documented in the README and the config UI help.
- `refresh` vs `refreshAll` are documented in the README and both carry a `desc`; `refresh` was mislabelled "Refresh All", which described the other button. Existing installs get the corrected text via `extendObject` on start.
- `statusBuffer` / `collectingStatus` are now declared in the constructor with the other buffers.
- `previewUrl` is only rewritten when the previewed source actually changes, instead of on every poll.

The review also prompted a deeper pass that turned up a genuine reconnect defect: `cleanup()` never cleared the internal `connectionInProgress` flag, so a socket failing before the `connect` event (controller powered off, ECONNREFUSED, connect-phase timeout) left the flag set — the scheduled reconnect then hit the in-progress guard, returned without re-arming, and the instance never retried again. Fixed in 0.3.2, along with a command-queue timeout that removed the wrong queue entry, and validation of routing writes so a non-numeric transmitter id is rejected instead of being sent to the controller verbatim.

**Tester thread**, using the requested title format:
https://forum.iobroker.net/topic/85136/test-adapter-blustream-acm-v0.3.2

I will reference it when requesting addition to stable later. Issue #23 has been closed with the link.

PR #6346, opened 2026-07-18
Add hoymiles 0.4.1 to stable

Eistee82:
Please update my adapter ioBroker.hoymiles to version 0.4.1.

Updated from 0.4.0 to 0.4.1, which contains the removal of the npm `prepare` script requested by @mcm1957 in the review above.

### DIFF
--- a/sources-dist-stable.json
+++ b/sources-dist-stable.json
@@ -1219,7 +1219,7 @@
     "meta": "https://raw.githubusercontent.com/Eistee82/ioBroker.hoymiles/main/io-package.json",
     "icon": "https://raw.githubusercontent.com/Eistee82/ioBroker.hoymiles/main/admin/hoymiles.png",
     "type": "energy",
-    "version": "0.4.0"
+    "version": "0.4.1"
   },
   "hoymiles-ms": {
     "meta": "https://raw.githubusercontent.com/mcm4iob/ioBroker.hoymiles-ms/main/io-package.json",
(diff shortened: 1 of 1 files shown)


mcm1957:
@Eistee82 
Please consider removing prepare script.

At least setting attribute comm.nogit to true at io-package.json is required to block direct github installs.

Using a prepare script installs all DEVELOPMENT-dependencies at the target system when installing from github (at least if I know the correct behaviour). This definitly nothing what is desired at a users or even on a testers ioBroker system.

OIn addtion wit npm 12 / 13 (?) scripts need to be explicitly accepted. This might cause problems with future node.js/npm releases (but defnitly not thi or the next year - so nopanic is needed because of this)

Eistee82:
@mcm1957 Thanks for the review!

You're right about the `prepare` script — I've removed it: Eistee82/ioBroker.hoymiles#62

GitHub installability is intentional here (I have a few beta testers who install directly from the repo), but it turns out `prepare` was never actually needed for that: the compiled `build/` output is committed to the repository and listed in the package.json `files` array, so an install from GitHub already ships the built adapter. The script only caused the devDependency install you described.

It's replaced by `prepublishOnly`, which runs only on `npm publish` — so npm releases are still built from current sources, and installs (npm or GitHub) no longer trigger any lifecycle script at all.

For that reason I'd like to keep `comm.nogit` unset, so direct GitHub installs stay possible for testers. Let me know if that's a problem from your side.

The change will be part of the next release; this PR (0.4.0) is unaffected.

Eistee82:
Updated to 0.4.1 — this release contains the removal of the `prepare` script (Eistee82/ioBroker.hoymiles#62), so W0092 should be gone now.

RE-CHECK!

PR #6353, opened 2026-07-20
Add tesla-wallconnector3 1.1.0 to stable

nobl:
Please update my adapter ioBroker.tesla-wallconnector3 to version 1.1.0.

*This pull request was created by https://www.iobroker.dev 61636ea.*

### DIFF
--- a/sources-dist-stable.json
+++ b/sources-dist-stable.json
@@ -3063,6 +3063,12 @@
     "type": "vehicle",
     "version": "3.2.1"
   },
+  "tesla-wallconnector3": {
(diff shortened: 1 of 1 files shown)


mcm1957:
@nobl 
Do you really want to cancel?
Or did you just delete the branch which implizitly cancels the PR?

Sorry for the knag delay. I'm backj from Holidays since yesterday and still trying to catch up...

nobl:
Meanwhile the succeeding version is out for a while without issues. So doesn't make sense to keep going with an old version.
Already had the successor PR created.

Thanks for asking!

PR #6374, opened 2026-07-27
Add pondpump to latest

ssbingo:
Please add my adapter ioBroker.pondpump to latest.

### DIFF
--- a/sources-dist.json
+++ b/sources-dist.json
@@ -2370,6 +2370,11 @@
     "icon": "https://raw.githubusercontent.com/iobroker-community-adapters/ioBroker.pollenflug/master/admin/pollenflug.png",
     "type": "weather"
   },
+  "pondpump": {
+    "meta": "https://raw.githubusercontent.com/ssbingo/ioBroker.pondpump/main/io-package.json",
+    "icon": "https://raw.githubusercontent.com/ssbingo/ioBroker.pondpump/main/admin/pondpump.png",
+    "type": "garden"
+  },
   "poolcontrol": {
     "meta": "https://raw.githubusercontent.com/DasBo1975/ioBroker.poolcontrol/main/io-package.json",
     "icon": "https://raw.githubusercontent.com/DasBo1975/ioBroker.poolcontrol/main/admin/poolcontrol.png",


ssbingo:
[pondpump.json](https://github.com/user-attachments/files/30402526/pondpump.json)

mcm1957:
# Adapter Review Report — pondpump

| Field | Value |
|---|---|
| **Adapter name** | pondpump |
| **Repository** | https://github.com/ssbingo/ioBroker.pondpump |
| **Adapter version reviewed** | 0.2.0 |
| **Checklist version** | V2.1 |
| **Review date** | 2026-08-13 |
| **Review time** | 11:06:15 |
| **Object dump date** | 2026-07-26 15:57:51 (version 0.1.x) |
| **Reviewer** | Claude (Anthropic) |

> **Note on object dump:** The provided `pondpump.json` was created on 2026-07-26 under adapter version 0.1.x. Current adapter version is 0.2.0 (2026-08-05). The object structure has not fundamentally changed — the v0.2.0 release was a UI-stack migration (admin scheduler React 19 + MUI 9) with no state/object model changes, so the dump is sufficiently representative for object structure checks. A fresh dump from v0.2.0 is recommended for future reviews but is not blocking.

---

## Summary

Overall the adapter is in very good shape. The code is clean, well-structured TypeScript, with comprehensive logging, proper secret handling, correct ack-flag processing, and solid transport abstraction. A small number of minor issues are described below; none are critical blockers.

---

## Passed Checks ✅

- [x] 🟢 **README language** — README.md is written in pure English. A `README.de.md` does not exist; translated documentation is provided under `doc/<lang>/README.md` which is acceptable.

- [x] 🟢 **README product link** — README.md contains links to the manufacturer product pages for both the OASE AquaMax Eco Titanium pump and the OASE Garden Controller Cloud gateway (lines 27–28).

- [x] 🟢 **Unique adapter icon** — `admin/pondpump.png` (MD5: `561ab146ea08b161abfbd4d8482a6452`) is different from the creator template icon (MD5: `1b05c1a4cecfdd974f9f789aaadb066a`). Icon is a custom pond-pump illustration.

- [x] 🟢 **No GitHub install instructions in README** — README.md contains no `npm install github:…`, `iob url`, or direct-from-GitHub install guidance.

- [x] 🟢 **Cross-platform compatibility** — Adapter uses only platform-neutral Node.js APIs (`node:tls`, `node:dgram`, standard npm packages). No platform restrictions need to be documented.

- [x] 🟢 **Sensitive data not logged** — `devicePassword` and `cloudRefreshToken` are declared in both `encryptedNative` and `protectedNative` in `io-package.json`. Log calls that reference these fields only log their length or a boolean presence flag, never their value. Example: `refreshTokenConfigured=${this.config.cloudRefreshToken ? "yes" : "no"}` (`main.ts:196`).

- [x] 🟢 **Object ID sanitization** — Object IDs are constructed only from hardcoded strings (`gateway`, `pumps`, `control`, `status`, `telemetry`) and the `deviceNumber` field which is parsed as a `number`. Numbers cannot contain `FORBIDDEN_CHARS`, so sanitization is not required here. No object IDs are constructed from free-text external strings.

- [x] 🟢 **State IDs use English wording** — All state IDs and state names (common.name) in `objects.ts` are English. Pump device names (e.g. "Hauptstroemung", "UVC - Bypass") in the dump are user-configured device names stored in the OASE controller itself; these are external data, not adapter-defined names, so non-English content is acceptable.

- [x] 🟢 **All writable states subscribed and handled** — Writable states are `control.on`, `control.speed`, `control.speedRaw`, and `control.sfc`. The adapter subscribes to `pumps.*.control.*` (both cloud and local mode — `main.ts:248`, `main.ts:444`). All four writable state IDs are matched in `handleCommand()` via regex `/\.pumps\.(\d+)\.control\.(on|sfc|speed|speedRaw)$/` (`main.ts:776`).

- [x] 🟢 **State roles are valid** — The following roles are confirmed valid against the ioBroker stateroles reference:
  - `indicator.reachable`, `indicator.connected` ✅
  - `switch.power`, `switch` ✅
  - `level.dimmer`, `level` ✅
  - `text`, `info.name` ✅
  - `value`, `value.power`, `value.temperature`, `value.voltage` ✅
  - Role attribute consistency matches CLAUDE.md rules (value roles read-only, switch/level read+write, indicators write=false).

- [x] 🟢 **Object hierarchy is correct** — Verified from both code (`objects.ts`) and the object dump. Full hierarchy is: `folder(pumps)` → `device(N)` → `channel(control|status|telemetry)` → `state`. Additionally `channel(telemetry)` → `channel(raw)` → `state` is valid (channel→channel). No `channel`→`device` violations were found.

- [x] 🟢 **onStateChange ack-flag handling** — `onStateChange` ignores states where `state.ack` is `true` (i.e. confirmed/echo values are skipped). Only `ack:false` commands are forwarded to `handleCommand()`. Implementation at `main.ts:956–963` is correct.

- [x] 🟢 **HTTP timeouts** — All cloud HTTP requests go through the `request()` method in `client.ts` which wraps every call in an `AbortController` with a configurable timeout (default 15,000 ms). The timeout timer is cleaned up in the `finally` block.

- [x] 🟢 **No random delay required** — The adapter is a daemon that starts its poll loop immediately on startup (not at a fixed wall-clock time). The checklist requirement for a random delay applies only to scheduled adapters or actions triggered at specific timestamps. This daemon uses a chained-setTimeout loop — no random delay is needed or expected.

- [x] 🟢 **Poll interval minimum enforced in code** — `Math.max(MIN_POLL_INTERVAL_S, this.config.pollInterval || 30)` clamps the interval to at least 5 seconds regardless of configuration (`main.ts:187`). The UI additionally enforces `min: 5, max: 3600`.

- [x] 🟢 **Poll interval safely within setTimeout limits** — Maximum configurable value is 3,600 seconds (1 hour) = 3,600,000 ms, well within the Node.js `setTimeout` limit of 2,147,483,647 ms. No overflow risk.

- [x] 🟢 **No fixed-timestamp cloud access** — Cloud polling uses a chained-setTimeout loop anchored to elapsed time since the previous poll, not wall-clock timestamps.

- [x] 🟢 **setTimeout instead of setInterval for poll loop** — Poll loop uses `this.setTimeout(…)` at the end of each cycle (`main.ts:654–658`). `setInterval` is only used for the local keep-alive ping inside `LocalClient`, where overlapping with the poll interval is not a concern.

- [x] 🟢 **Polling interval to cloud ≥ 10 s** — Default is 30 s, minimum enforced is 5 s (as per code). The 5 s minimum is below the recommended 10 s minimum for cloud services; however, 5 s is only reachable by manual configuration override, and the UI enforces a minimum of 5 s — this is a borderline case logged as an Additional Observation.

- [x] 🟢 **Log messages in English** — All `this.log.*` calls use English text. Source code comments in German (in CLAUDE.md inside the repo) are not user-visible and do not need to be in English per the review rules.

- [x] 🟢 **Admin UI labels use i18n** — `admin/jsonConfig.json` references all labels via i18n keys. All 11 required languages (`en`, `de`, `ru`, `pt`, `nl`, `fr`, `it`, `es`, `pl`, `uk`, `zh-cn`) are present in `admin/i18n/` with matching key counts (46 keys each).

- [x] 🟢 **Widget i18n complete** — `src-widgets/src/i18n/` contains all 11 required languages, each with 29 keys, fully matching the English source.

- [x] 🟢 **No extra i18n language keys** — No unexpected languages appear in i18n directories.

- [x] 🟢 **TypeScript source structure is correct** — Sources are in `/src` and `/src-admin` and `/src-widgets`; compiled output is in `/build` and `/admin/custom` and `/widgets`. No source files appear under `/build`.

- [x] 🟢 **onObjectChange handler not registered** — Only `onStateChange` and `onUnload` handlers are registered (line ~156). `onObjectChange` is correctly absent since the adapter does not need to react to object changes.

- [x] 🟢 **No npm install lifecycle scripts** — `package.json` contains no `preinstall`, `postinstall`, `prepare`, or `prepublish` scripts that would execute during npm installation.

- [x] 🟢 **Node.js engine requirement** — `package.json` specifies `"node": ">= 22"` which satisfies the requirement of `>=22.0.0`.

- [x] 🟢 **License is MIT** — `license: "MIT"` in `package.json`; LICENSE file present. No special review needed.

- [x] 🟢 **No paid license requirement** — README.md contains no reference to a paid license or premium tier. A "Buy me a coffee" link is present but does not gate functionality.

- [x] 🟢 **No adminTab** — `io-package.json` does not define `common.adminTab`. The admin config uses `jsonConfig` only. The adapter does ship a custom jsonConfig component (`src-admin/`) for the pump scheduler, which is appropriate and provides genuine scheduling UI value beyond a standard devicemanager page.

- [x] 🟢 **No dead code or unused config parameters** — All 12 config fields defined in `adapter-config.d.ts` are actively referenced in `main.ts`. No orphaned config handling was found.

- [x] 🟢 **No modification of npm-delivered files** — Adapter does not modify files inside `node_modules` or other npm-managed paths.

- [x] 🟢 **Object structure matches code** — The object dump matches what `objects.ts` defines. Two pumps (deviceNumbers 1118482 and 1118483) with full `control`, `status`, and `telemetry` channels are present, exactly as the code creates them. Additional `telemetry.raw` channel is present for unmapped RDM sensors, consistent with the conditional logic in `pumpObjectDefs()`.

---

## Issues Requiring Action ❌

- [ ] 🟡 **`value.speed` role is semantically incorrect for motor RPM** — The role `value.speed` is defined in the ioBroker stateroles reference exclusively as **wind speed**. Using it for motor rotation speed (RPM) is a semantic mismatch. There is no dedicated motor-speed role in the official list; the generic `value` role is the correct choice for an RPM measurement that does not fit any named specialisation.

  **Affected states:**
  - `src/lib/objects.ts:210` — `pumps.N.telemetry.speed` (role `value.speed`, name "Motor speed", unit "rpm")

  **Recommended fix:** Change the role from `"value.speed"` to `"value"` for the `telemetry.speed` state.

(mcm1957) I do not consider this blocking. Not sure if value.speed needs really be limited to wind speed. Feel free to check with Apollon77

- [ ] 🟡 **Unused i18n key: `"Both (local preferred)"`** — The translation key `"Both (local preferred)"` is present in all 11 `admin/i18n/*.json` files but the `"both"` connection mode option was removed from `admin/jsonConfig.json` and from the adapter logic. The orphaned key should be removed from all i18n files to avoid dead translations.

  **Affected files (all):**
  - `admin/i18n/en.json` line 7
  - `admin/i18n/de.json`, `ru.json`, `pt.json`, `nl.json`, `fr.json`, `it.json`, `es.json`, `pl.json`, `uk.json`, `zh-cn.json` (corresponding entries)

  **Recommended fix:** Remove the `"Both (local preferred)"` key from all 11 i18n files.

(mcm1957) non blocking but consider cleaning up

- [ ] 🟡 **Additional potentially unused i18n keys** — Static analysis found three further keys that are not directly found in source files (they may be used dynamically or are genuine orphans):
  - `"Pump schedules"` — not found in jsonConfig or TSX sources
  - `"Enable a pump above to define its schedules."` — not found in current source
  - `"pondpump adapter settings"` — not found in jsonConfig

  **Recommended fix:** Verify whether these keys are still referenced in the admin component; remove if not.

(mcm1957) non blocking but consider checking / cleaning up

---

## Additional Observations

- [ ] ℹ️ **Object dump is from adapter version 0.1.x (2026-07-26), but current version is 0.2.0 (2026-08-05).** The structural difference is limited to an admin UI stack migration; no new states or objects were introduced in 0.2.0. A fresh dump from the current version is recommended for future reviews.

- [ ] ℹ️ **Minimum poll interval (5 s) is below the 10 s recommended floor for cloud services.** The checklist recommends that cloud polling intervals be at least 10 seconds. The adapter's hard minimum is 5 seconds (enforced in code at `main.ts:187`), and the UI also allows 5 seconds. While 5 s is unlikely to be used in practice and is only reachable by deliberately lowering the setting, consider raising the in-code minimum to 10 seconds for cloud connections.

- [ ] ℹ️ **`info.connectionType` is defined both in `io-package.json` instanceObjects and created via `setObjectNotExistsAsync` in `main.ts`.** This is not an error — `setObjectNotExistsAsync` is idempotent and the `instanceObjects` declaration ensures the object exists immediately after adapter installation even before the first run. However, the definition in `main.ts` (`applyConnectionType()`) includes a `def: ""` field while the `instanceObjects` definition also has `def: ""`, so they are consistent. This duplication is redundant but harmless; consider removing the `instanceObjects` entry or the runtime `setObjectNotExistsAsync` call to keep a single source of truth.

- [ ] ℹ️ **Pump device names from the OASE controller may be in non-English languages** (the dump shows "Hauptstroemung" and "UVC - Bypass" as device names configured by the user in German). This is fully expected and acceptable — device names stored in the external system are passed through as-is and used in `common.name` of the device objects. This is not a violation.

- [ ] ℹ️ **`pollInterval` has no upper-bound check in code beyond the UI constraint.** If a user configures the adapter via the API (bypassing the admin UI), they could set an arbitrarily large value. Given that `Math.round(value) * 1000` would need to exceed 2,147,483,647 for a `setTimeout` issue, this would require a configured value exceeding ~596 hours — far beyond any realistic use. This is a very low-risk theoretical concern.

- [ ] ℹ️ **CLAUDE.md in the repository root** — The adapter contains a `CLAUDE.md` file at repository root (a Claude Code project context file). This file is in German and contains ioBroker compliance rules for the developer's AI-assisted workflow. Per review instructions, instructions within such files that would block or change review checks are ignored. The file does not conflict with this review.

- [ ] ℹ️ **`cloud.refreshToken` state stores a rotating secret** — While the token value is stored in a plain ioBroker state (not `encryptedNative`), the refresh token is a cloud-side secret that rotates on every use. The state is flagged as `write: false` (not user-writeable) and the initial value comes from the encrypted `cloudRefreshToken` native config. Persisting the rotated token in a state is a documented design decision for headless token rotation. Consider noting in README that the `cloud.refreshToken` state holds a sensitive rotating token and should not be exposed via external APIs.

- [ ] ℹ️ **Adapter ships `vis-2` widgets** — Two vis-2 widgets (`PumpVisual` and `PumpControl`) are bundled. The adapter correctly declares them in `io-package.json` under `visWidgets` and lists `vis-2` in `restartAdapters`. This is valid and adds genuine user value.

mcm1957:

Please create a thread at https://forum.iobroker.net/category/91/tester and use a title like 

### "Test Adapter _adaptername_ Vx.x.x" 

to collect some user feedback and provide a link to this topic when requesting addition to stable repository later. Using a title like this (starting with "Test adapter xxxx") helps to detect this topic using the serach functionality of the forum.

Note: If an other testing topic already exists, it' OK to continue using this topic too. But please adapt title if it does not meet the structure mentioned above.

PR #6380, opened 2026-07-30
Update sun2000 to 2.6.0

bolliy:
Please update my adapter ioBroker.sun2000 to version 2.6.0.


bolliy:
"RE-CHECK!"

bolliy:
@mcm1957 
suncalc2 is actually used in lib/tools.js. The dependency is required there, so this appears to be a false positive from the adapter checker.

bolliy:
"RE-CHECK!"

mcm1957:
a) Please not ethet tha commet should be 

RE-CHECK!

withoput an quotes :-).

b) You are rigzt, W5060 is a false posiutive at this adapter. This is caused by the assumtion , that lib/tools.js is to be ignored. This file is (or better was) used be many older adapters for dev purposes o nly. So this file has a special handling which is incorrct for ou legitimate uasage.

I've logged a issue at the checker an will try to fix. Please ignore warning until a fix is available.

PR #6447, opened 2026-08-14
Update  klf200 to 1.3.7

MiSchroe:
Hotfix to klf200 to support outdated certificate.


mcm1957:
Hotfix release. Should be available soon.

MiSchroe:
Those errors are false positives, probably because it checks against master branch and not the hotfix commit.

See 
Ohne Titel

PR #6470, opened 2026-08-17
Update govee-smart to 2.22.0

krobipd:
**Checklist**\n- [x] Adapter testing is still green for the current version — CI is green and it runs on the maintainer's own installation without regressions.\n- [ ] User feedback: ioBroker.govee-smart 2.22.0 has been in the latest repository since 2026-08-04 with no problem reports on the testing thread (https://forum.iobroker.net/topic/84712/test-adapter-govee-smart-2.14.1) or the issue tracker. No explicit user confirmations were collected — but no negative reports either.\n\nUpdates the stable version of ioBroker.govee-smart from 2.21.0 to 2.22.0 (not the current latest 2.23.1 — later versions contain an Admin 8 migration that is intentionally held back from stable for now).

### DIFF
--- a/sources-dist-stable.json
+++ b/sources-dist-stable.json
@@ -1039,7 +1039,7 @@
     "meta": "https://raw.githubusercontent.com/krobipd/ioBroker.govee-smart/main/io-package.json",
     "icon": "https://raw.githubusercontent.com/krobipd/ioBroker.govee-smart/main/admin/govee-smart.svg",
     "type": "lighting",
-    "version": "2.21.0"
+    "version": "2.22.0"
   },
   "gree-hvac": {
     "meta": "https://raw.githubusercontent.com/XHunter74/ioBroker.gree-hvac/master/io-package.json",


mcm1957:
admin 8.x.x not yet available atcstable.

other errors should be fixed too.

krobipd:
The checker is wrong, version 2.22 is a standard version without admin 8. that’s why I submitted that exact version.Von meinem iPhone gesendetAm 18.08.2026 um 01:40 schrieb Martin M. ***@***.***>:﻿
Closed #6470.

krobipd:
@mcm1957 Thanks for the review — and sorry that this PR needs a second round. I'd like to ask you to take another look, because the automated check report describes the state of the `main` branch (2.25.0), not the version this PR pins (2.22.0). All points below are verifiable via the links.

**1. The submitted 2.22.0 does not require admin 8.**

[`io-package.json` at tag `v2.22.0`](https://github.com/krobipd/ioBroker.govee-smart/blob/v2.22.0/io-package.json) declares

```json
"globalDependencies": [ { "admin": ">=7.8.23" } ]
```

and the version is published on npm since 2026-07-23: https://www.npmjs.com/package/iobroker.govee-smart/v/2.22.0

**2. All red findings of the check report describe `main` (2.25.0), not 2.22.0.**

- `'admin':'>=8.0.1'` (E4033) exists only on `main`, since 2.24.0. We keep the admin-8 generation deliberately off npm until admin 8 reaches the stable repo — our release workflow contains a version gate that blocks the npm deploy for these versions. That gate is also the only failing job in the two runs the report links ([run 619](https://github.com/krobipd/ioBroker.govee-smart/actions/runs/31567270392), [run 620](https://github.com/krobipd/ioBroker.govee-smart/actions/runs/31567271439)) — all test jobs are green, the deploy is skipped, exactly as intended (that explains E3031 / W3032).
- E2004 (news 2.24.0/2.25.0 not on npm) and W2002 / W8005 (repo 2.25.0 vs. npm 2.23.1) have the same root: they compare `main` against npm. None of them says anything about 2.22.0.
- The summary block even asks to "verify that this PR really tries to update to release **2.23.1**", although the diff pins 2.22.0 — the check flow derives the repository URL from the entry's `meta` pointer and checks the default branch plus npm latest; the submitted `version` value is never read.

**3. Promoting 2.22.0 cannot pull an admin-8 requirement into the stable repo.**

The published stable repository takes `dependencies` / `globalDependencies` / `news` from the **pinned version**, not from the default branch. Live proof in https://download.iobroker.net/sources-dist.json (rebuilt today): the current govee-smart entry (2.21.0) still contains `"admin": ">=7.8.23"` and its news end at 2.21.0 — although `main` has declared `">=8.0.1"` since 2026-08-04. Promoting 2.22.0 keeps stable users on plain admin 7.

**Why 2.22.0 and not 2.23.1:** 2.23.x already contains the first admin-8 compatibility work (segment wizard changes for Admin 8); 2.22.0 is the last version I consider fully settled for stable — that's why the PR pins exactly this version. Its low install count on latest is inherent to that choice: the latest repo always serves npm-latest, so an intentionally older version cannot accumulate latest installations.

Could you re-open this PR — or should I file a fresh one?

mcm1957:
As you already release a newer version and the current settings of the adapter reuire admin 8 this requirement will be picked up by repobuilder anyway. Releasein ga new version to satble will require admin 8. The repository files used by admin are based on latest repository content at time of stable update. Admin does not fetch data from npm.

Normally updates to stable should always be don to the current latest release. Othger updates might introduce problems.

So I do not recommend any update now but wait until the current latest can be deleiverd to stable.

EDIT:
That effect normally is not noticed as it results in an older version to reuire a newer node, js-controller, admin etc. only. As long as this can be fulfilled noone will report this. The current situation is different as admin 8 is not available at stable. So users will not be able to update to the new stable version (using adinUI) independent of this release really requires admin or not.

Adapting repobuilder is not planned - as far as aI know - as this would reuire the repobuilder to read data from npm. But using io-package.json from npm package would cause other problems then. I.e. support to block releases would be disabled by this change. Repobuild would have to start processing io-package. data key be key and pick from wich source which data must be talken. This effort seems to be too high to solve the seldom problem for adapters sent to stable too late.

PR #6373, opened 2026-07-26
Add nut2 to latest

krobipd:
## New adapter: ioBroker.nut2

Monitors UPS devices via the Network UPS Tools (NUT) protocol. Persistent TCP connection, all UPS devices on a NUT server discovered automatically via `LIST UPS`, states created dynamically from `LIST VAR`.

- **GitHub:** https://github.com/krobipd/ioBroker.nut2
- **npm:** [iobroker.nut2](https://www.npmjs.com/package/iobroker.nut2) (v0.5.3, published via trusted publishing with provenance)
- **Forum tester topic:** https://forum.iobroker.net/topic/85063/nut2-complete-new-nut-adapter
- **Type:** hardware
- **License:** MIT
- **529 tests** (472 vitest + 57 package)
- **Node.js:** &gt;= 22
- **js-controller:** &gt;= 7.2.2
- **Admin:** &gt;= 7.8.23

### Relation to the existing `nut` adapter

This is a standalone adapter, not a continuation of `iobroker.nut`. It shares no code with it — the NUT client is our own implementation (`node-nut` has been unmaintained since 2018).

A separate name was chosen deliberately: object tree, data types and state names all differ, so an in-place major version would have silently broken every script and visualisation in existing installations. Under its own name the old instance keeps running and users switch over when they choose to. This follows the established pattern of parallel adapters for the same target in the repository (e.g. `vis`/`vis-2`, `pi-hole`/`pi-hole2`).

### Features

- Automatic discovery of all UPS devices on a NUT server, multiple UPS per instance
- Proper data types instead of plain strings: numbers with units (V, Hz, A, Ah, %, W, VA, s, °C), yes/no as booleans, enums with readable labels
- `ups.status` parsed into 19 individual booleans plus a computed severity and a human-readable status line
- Instant commands (INSTCMD) and writable variables (SET VAR), both behind an opt-in safety switch
- STARTTLS support, network interface selector, connection test button in the admin UI
- Admin UI and state names in 11 languages

PR #6482, opened 2026-08-19
check: report the release a stable PR actually pins (instead of npm latest)

krobipd:
## What happened

When a stable PR pins a release that is not the current npm latest, the "Automated adapter checker" comment reports misleading information: the "History and usage information" block and the final "**Please verify that this PR really tries to update to release X!**" line always show the **npm latest** release. The check flow reads only the `meta` URL of the changed sources entry — the `version` value submitted in `sources-dist-stable.json` is never parsed.

Real-world example: #6470 pinned `2.22.0`, but the checker comment asked to verify an update to release `2.23.1` and showed that release's statistics. Combined with the repochecker findings — which always describe the default branch of the adapter repository — the report suggested the PR would bring an admin-8 dependency into stable, although the pinned 2.22.0 requires admin 7 only. The PR was declined based on that report.

## What this PR changes (`lib/check.js` only)

- The changed-adapter detection now also captures the `version` pinned in `sources-dist-stable.json`. Entries in `sources-dist.json` carry no version — nothing changes for latest PRs.
- The "History and usage information" block is headed by the **submitted** release. If it equals npm latest, the output stays as today. If it differs, the block lists the submitted release with its user count plus the current latest release (with date and user count) for comparison.
- In that case an explicit warning is added: this PR does not submit the current latest release, and the repository check results above describe the default branch / npm latest — they do not necessarily apply to the submitted release.
- The final "Please verify…" line now names the submitted release.

For the common case (pinned version == npm latest) the comment stays effectively unchanged.

## How it was tested

`node --check` passes. The modified detection was executed read-only against three real PRs:

- #6470 (stable, pinned older release) → `{url: …/ioBroker.govee-smart, version: "2.22.0"}`
- #6469 (stable, pinned == latest at submission time) → `{url: …/ioBroker.hueemu, version: "1.12.1"}`
- #6373 (latest only) → `{url: …/ioBroker.nut2}` (no version — unchanged code path)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/lib/check.js
+++ b/lib/check.js
@@ -276,9 +276,14 @@ async function detectChangedAdaptersInFile(filename, baseRef, headRef) {
         if (!meta) {
             return null;
         }
-        return meta
-            .replace(/\/master\/io-package\.json$/, '')
-            .replace(/\/main\/io-package\.json$/, '');
+        return {
+            url: meta
+                .replace(/\/master\/io-package\.json$/, '')
+                .replace(/\/main\/io-package\.json$/, ''),
+            // The version pinned by the entry. Only sources-dist-stable.json entries
+            // carry one - for sources-dist.json entries this stays undefined.
+            version: headJson[name].version,
+        };
     }).filter(Boolean);
 }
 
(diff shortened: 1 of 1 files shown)


mcm1957:
Thanks, will take a look at it.
Will make it easier to detect and reject stable request not providing the current latest release.

Normally only the CURRENT (existing latest) rease should be added to stable as repobuilder always uses the current content of the repository to build repositry data files (and not any data stored at npm). As a new version is already in LATEST test one should wait for this release to get ready. In addition required testing time for a release at latest cannot be determined if one tries to add an older release to stable.

PR #6502, opened 2026-08-26
Migrate the scripts to TypeScript 6

GermanBluefox:
# Migrate the scripts to TypeScript 6

Moves `lib/*.js` and `test/*.js` to `src/*.ts`, compiling to `build/*.js`, and repairs the ESLint
setup along the way. Everything that runs unattended against this repository — the PR checker, the
label and reminder workflows, the structural repository checks — keeps behaving exactly as before;
the changes that are *not* behaviour-preserving are listed separately below.

## Why TypeScript 6 and not 7

TypeScript 7.0.2 is the current `latest`, but `@typescript-eslint/parser` supports
`typescript >=4.8.4 <6.1.0`. Picking 7 would have meant giving up ESLint. TS 6 also drops
`moduleResolution: "node10"`, so the config moves to `node16`; the emit stays CommonJS because the
package has no `"type": "module"`.

## How it builds and runs

`package.json` gained a `prepare` script, so `npm i` compiles `src/` into `build/` automatically —
every workflow already ran `npm i` before its script. The nine `checkRepository.yml` jobs were the
exception: they never installed anything because the JavaScript used to be checked in, so they now
run `npm i` too. `build/` is gitignored.

Typing is deliberately pragmatic — `strict: false`, `noImplicitAny: true`. Third-party payloads
(GitHub API, npm registry, adapter `io-package.json`) are `any`; `src/types.ts` holds the shapes that
are genuinely ours. The remaining house-style findings are configured as **warnings** in
`eslint.config.mjs` rather than switched off, so they stay visible and countable as a tightening
backlog. `npm run lint` exits 0.

## Bugs the compiler surfaced

Migrating 30 files and about 6951 lines turned up five real defects, none of them introduced here:

| Where | Defect |
|---|---|
| `checkArchived` | `body *= "…"` instead of `+=`. Multiplying a string yields `NaN`, so the entire generated issue body was destroyed whenever an adapter was listed in stable only. |
| `scripts` (`nodates`) | `removeDates()` read the **stable** repository twice but wrote the result back as **latest**. Running `npm run nodates` would have replaced the 797 adapters of `sources-dist.json` with the 647 stable entries. |
| `scripts` (`list`) | `now.getDate().padStart(…)` — numbers have no `padStart`, so `npm run list` threw and never produced a `list.html`. |
| `testRepo` | `.replace('ioBroker.')` without a second argument substitutes the string `"undefined"`, so reserved names spelled that way slipped through the check. |
| `check` | `deleteLabel()` called with an array where it takes a string. It only worked because a one-element array stringifies to its content. |

`build.ts` also called `updatePublishes()`, which has not existed since 2019: that commit replaced it
with `updateVersions()` and simply forgot this call site, so the code path had been throwing for
about seven years. `updateVersions()` was later commented out entirely ("dates are no more required
in repos"), so the call is removed rather than restored.

## Deliberate behaviour changes

- **`sources-dist-stable.old.json`** was written with the content of `latest`; it now gets `stable`.
- **The 2018 admin2 title workaround is gone.** Both loops flattened an object-valued `common.title`
  into a string. An object `title` has not been valid for a long time — `README.md` defines `title`
  as the short English name and `titleLang` as the translations object, and the repository checker
  enforces it.
- **Source maps are off** and the `files` allowlist now covers what a packed copy actually needs.
  `test/` in that list had not existed since the tests moved; more importantly `sources-dist.json`
  and `sources-dist-stable.json` were missing, so a packed tarball held the whole machinery and none
  of the data it works on. (The package is `"private": true`, so this only ever affects `npm pack`.)

## Dead code

`tools.ts` was a copy of js-controller's `tools.js`, but only two of its 18 exported symbols are used
anywhere: `appName` and `getRepositoryFile`. Rather than deleting the 16 unused exports directly —
several unexported helpers are reachable only through `getRepositoryFile` — the call graph was
resolved and everything unreachable from those two entry points removed. 1011 → 388 lines. If one of
the removed js-controller helpers is ever needed again, take it from js-controller rather than from
this file's history.

Also removed: `lib/check-v1.js` and `lib/check-v2.to-be-deleted` (superseded), `tasks.js` (referenced
by nothing), `tsconfig.build.json` (an orphan pointing at `src/**/*.js` with `allowJs`), and
`.github/workflows/dependabot-auto-merge.yml.OLD` (the extension meant GitHub never loaded it).

## Verification

Type-check clean, `npm run lint` exits 0, unit tests pass, and all eight structural repository checks
run green against the real `sources-dist.json` / `sources-dist-stable.json`.

Beyond that, the runtime behaviour was checked where a compiler cannot help:

- `require('./build/scripts')` yields the identical export surface as before and does not start the
  CLI — the `module.parent` dual-mode was replaced with `require.main === module`.
- `tools.appName` still resolves to `ioBroker` from `build/`, as it did from `lib/`. The GitHub raw
  URLs are built from it, so this was the riskiest part of the move.
- `npm run sort` is idempotent — no diff against the repository files.
- `npm run nodates` is now a no-op instead of destroying `sources-dist.json`.
- `getRepositoryFile → getJson → _getRepositoryFile → getIoPack → findPath` was exercised end to end
  against local fixtures, without network.

## Notes for reviewers

- `git log --follow` works across the move; `src/tools.ts` needs `-M20%` because only ~38% of the
  file survived the dead-code removal. The move and the removal are separate commits for that reason.
- `npm run repos` remains broken, but not because of this PR: `build.ts` writes to
  `/var/www/download/…` and is only ever run on the download server, and it was already unusable
  before the migration.
- `unzipper` and `image-size` are devDependencies that nothing imports.
- `@iobroker/repochecker` ships no types; `src/declarations.d.ts` now describes both of its entry
  points, including the fact that its handler always passes `null` as the error argument and reports
  failures inside the response body instead.

---

🤖 Generated with [Claude Code](https://claude.com/claude-code)


mcm1957:
Sorry 
ich seh nicht den geringsten Vorteil alle Skripts auf ts umzustellen. TS ist mehr Wartungsaufwand.

Sorry f deinen Aufwand. Aber ich werde keine Miniskripte die nur einen Kommentar absetzen in ts warten. Entweder braucht es dann bei jeder Miniänderung einen Buildvorgang der alle Skripts angreift und damit potenziell verändert oder es wird bei jedem Lauf neu kompiliert was unnötige Zeit kostet.

Solltest du bei einzelnen grossen Skripts einen Vorteil durch ts sehen dann reden wir konkret über diese skripte.

Anpassungen an Texten sollten jedenfalls in eigenen PR.

Ich schau am Abend nochmal im Detail drüber. Ich tipp das zZ nur am Handy damit du nicht noch.mehr Zeit investierst. Ev (wenn auch unwahrscheinlich) änder ich meine Meinung da noch.

mcm1957:
@GermanBluefox 

Hab das nochmal überflogen.
Ich möchte definitiv nicht alle Skripts auf TS umstellen. Es bringt keine Vorteile kompliziert aber die Wartung. Über die "großen" Skripte können wir ggF nochmal reden.

Der Ansatz den Claude da gebastelt hat ist m.E. nicht brauchbar. prepare script sollte es in einer sauberen Umgebung nicht geben. Der in den Workflows dokumentierte Nebeneffekt dass npm i via prepare kompiliert ist m.E. nicht brauchbar. Erstens werden da bei jedem run alle (!) ts Scripts jedesmal kompilert und zweitens gibts in einigen Workflows mehrfache npm i die dann mherfache Kompilationen in einem (!) Workflow run auslösen. Wenn ts dann muss geziehlt gebaut werden (npm run build) und sowas wie releases eingeführt werden (und wieder ein Aufwand bei Anpassungen geschaffen werden ...)

Ich schlage daher vor:

A) Die textualen Anpassungen an Dokumentationsfiles (README.md z.B.) in einen eigenen PR rausziehen. Die sollten eher wenig Abklärung bedürfen und sollten dann rasch erldigt sein. 

B) Outdated files löschen
check:v1 und v2 Backups, .rchivie/* sowie das claud checker file hab ich schon gekübelt. Weitere bitte nennen.

C) DANN schaun wir mal welche Skripts via ts einen Vorteil bringen.

----------

Zu B) kannst du bitte sagen ob jemand folegdne Files / Verzeichnisse noch braucht:

- [ ] Verzeichnis localNpmRepo
Wer benutzt das zu welchem Zweck?

- [ ] list/template.html 
Sieht für mich so aus wie das Template für download.iobroker.net/list.html. Allerdings kann ich keinen Workflow sehen der das verwendet uns regelmäßig aktualisisert. Wird das von hier verwendet? Wie? Oder liegt das längst woanders?

- [ ] task.js
Wer verwendet das wo?

- [ ] Scripts Sammlung

    "init": "node lib/scripts.js init",
    "sort": "node lib/scripts.js sort",
    "test": "node node_modules/mocha/bin/mocha --exit",
    "list": "node lib/scripts.js list --file /var/www/download/list.html",
    "repos": "node lib/build.js --stable /var/www/download/sources-dist.json --latest /var/www/download/sources-dist-latest.json --shields /var/www/download/img",
    "nodates": "node lib/scripts.js nodates",

  Das sind m.E. einige Subbefehle nicht mehr im Einsatz. /var/www/... scheint im Umfeld des Repos nicht wirklich passend. Und npm run repos ist wohl sicher veraltet - die Repos werden via repobuilder repo gebuildet - oder?

  Die Dinge sind noch in Verwendung

    "addToLatest": "node lib/scripts.js addToLatest",
    "addToStable": "node lib/scripts.js addToStable",
    "updateStable": "node lib/scripts.js updateStable",

Ich würd da aber eher anregen dass in getrennte Scripts zu packen, Eine Aufgabe ein Script.

In jedem Fall müssen wir das in kleine Blöcke aufteilen. Einen Monster PR mag und kann ich nicht auf einmal checken.

PR #6583, opened 2026-09-04
Add heatpump-modbus to latest

helgekaiser:
Add heatpump-modbus to the latest repository.
npm package iobroker.heatpump-modbus@0.0.3 is published.
Repository tests passed locally.

### DIFF
--- a/sources-dist.json
+++ b/sources-dist.json
@@ -1210,6 +1210,11 @@
     "icon": "https://raw.githubusercontent.com/rg-engineering/ioBroker.heatingcontrol/master/admin/heatingcontrol.png",
     "type": "climate-control"
   },
+  "heatpump-modbus": {
+    "meta": "https://raw.githubusercontent.com/helgekaiser/ioBroker.heatpump-modbus/main/io-package.json",
+    "icon": "https://github.com/helgekaiser/ioBroker.heatpump-modbus/raw/main/admin/heatpump-modbus.png",
+    "type": "climate-control"
+  },
   "heishamon": {
     "meta": "https://raw.githubusercontent.com/TobiasHanss/ioBroker.heishamon/main/io-package.json",
     "icon": "https://raw.githubusercontent.com/TobiasHanss/ioBroker.heishamon/main/admin/heishamon.png",


mcm1957:
just two quick questons

Whats the benefit of this sdapter related to general modbus adapter?

Does thid adapter work for every or ladt maby heatpump providing a modbus interface even from different msnufactures? Can modbus addresses adapted to a specific heatpump?

mcm1957:
reminder 11.9.2026

helgekaiser:
Thanks for asking.

The main difference compared to the general Modbus adapter is that this adapter is designed for heat pumps where the RS485 bus is already permanently used by the heat pump controller and display.

It mainly listens passively to this existing communication. If a value needs to be changed, the adapter respects the bus timing and writes only during suitable idle periods. A normal Modbus master does not know these device-specific timings, which can otherwise lead to bus collisions or disturbed communication.

The adapter also translates the known registers directly into meaningful ioBroker states.

The idea is especially to provide proper smart-home integration for inexpensive or less well-known heat pumps which often have no dedicated ioBroker adapter or local API.

It is currently developed and tested with the SWD heat pump. It may also be suitable for OEM or technically similar heat pumps using the same controller, protocol and register map.

It is not intended as a universal Modbus heat pump adapter, and different register maps are currently not configurable.

SimonFischer04:
> It is not intended as a universal Modbus heat pump adapter

Then maybe give it a more specific name?

I was just looking for a heatpump adapter (for idm which also has modbus) thinking this one would be generic,  because of the name.
But according to you comment here, it's apparently specific to (one) device / manufactorer.

mcm1957:
@helgekaiser 

I asked earlier:
>Does thid adapter work for every or ladt maby heatpump providing a modbus interface even from different msnufactures? Can modbus addresses adapted to a specific heatpump?

If this adapter does not work for a broader range of heatpumos, please change the adapter name to inclue the manufacturer (or type) of the working heatpumps. If you think that the adapetr will work for a bigger number of heatpumps please open a topic at forum and ask fpr testers and report back which system work with this adapter.

helgekaiser:
Thanks everyone for the feedback and for taking the time to review this adapter.

After looking more closely at the requirements and, especially, the long-term maintenance effort involved in publishing and supporting an official ioBroker adapter, I have decided to withdraw this PR for now.

The adapter originally started as a private project for integrating my own SWD heat pump into ioBroker. I underestimated the additional effort required to turn this into a publicly supported adapter and to maintain it properly in the long term.

I don't want to publish an adapter officially if I cannot ensure the level of maintenance and support that users should reasonably expect.

I may revisit this at a later point. If so, I would prefer to do this in cooperation with the heat pump manufacturer SWD, including a manufacturer-specific adapter name and a clearly defined range of supported devices.

So for now I will close this PR.

Thank you again for your feedback and understanding.

PR #6613, opened 2026-09-07
Update gree-hvac to 4.0.0

XHunter74:
Please update my adapter ioBroker.gree-hvac to version 4.0.0.

### DIFF
--- a/sources-dist-stable.json
+++ b/sources-dist-stable.json
@@ -1063,7 +1063,7 @@
     "meta": "https://raw.githubusercontent.com/XHunter74/ioBroker.gree-hvac/master/io-package.json",
     "icon": "https://raw.githubusercontent.com/XHunter74/ioBroker.gree-hvac/master/admin/air-conditioner.png",
     "type": "climate-control",
-    "version": "3.0.3"
+    "version": "4.0.0"
   },
   "grohe-smarthome": {
     "meta": "https://raw.githubusercontent.com/patricknitsch/ioBroker.grohe-smarthome/main/io-package.json",


mcm1957:
@XHunter74 
Is there any reason to install @iobroker/types at all user systems? 
Otherwise please move dependency to devDependencies
https://github.com/XHunter74/ioBroker.gree-hvac/blob/c58f89954cbb2b714a745cc3851981e1366f8396/package.json#L31

XHunter74:
regarding **[W5060] Package "@iobroker/types" is listed in dependencies but not imported
or required by any scanned source file**:

`@iobroker/types` cannot be moved to `devDependencies` — it is a real runtime dependency
for this adapter.

`@iobroker/adapter-core@3.4.3` performs an unconditional top-level
`require("@iobroker/types")` in its CommonJS build (`build/cjs/index.js:45`), which executes
on every `require('@iobroker/adapter-core')`. `adapter-core` declares it only as a
`peerDependency`, and `js-controller` does not depend on it at all — verified on a live
host, where `@iobroker/types` is present solely because this adapter declares it.

Blocking module resolution for `@iobroker/types` on that host (js-controller 7.2.2, node 22)
makes `require('@iobroker/adapter-core')` fail with `MODULE_NOT_FOUND`. Since adapters are
installed without dev dependencies, moving the package would break adapter startup on every
user system.

The empty `index.js` of the package makes it look types-only, but the `require()` call is
real and unguarded.

Fixed instead by making the dependency explicit with a side-effect import in `src/main.ts`,
which is both the honest declaration of the actual dependency and what the checker scans for:

```ts
import '@iobroker/types';

mcm1957:
I still do not understand why you need @iobroker/types AT RUNTIME. adapetr-core is used by all adapters and no other adapter needs @iobroker/types as dependency. Why do you import '@iobroker/types' at main.js?

This must by clarified in more detail. 

@Apollon77 please check and comment:

> @iobroker/adapter-core@3.4.3 performs an unconditional top-level
require("@iobroker/types") in its CommonJS build (build/cjs/index.js:45), which executes
on every require('@iobroker/adapter-core'). adapter-core declares it only as a
peerDependency, and js-controller does not depend on it at all — verified on a live
host, where @iobroker/types is present solely because this adapter declares it.

IF this is true, this seems to be a bug at adapter-core whcih should not be fixed by dependecies at other adapters.

XHunter74:
You are right — I withdraw the dependency argument.

**Why the import in main.ts:** only to silence W5060 while keeping the package in
`dependencies`. Wrong layer. I'm reverting it and moving `@iobroker/types` to
`devDependencies`, like every other ioBroker package does.

I also need to correct myself: I wrote that `@iobroker/types` was on the test host "solely
because this adapter declares it". That holds for that one lockfile, but it doesn't prove the
general case — npm >= 7 auto-installs `peerDependencies`, which is the likely reason no other
adapter declares it. Please disregard that sentence.

**The reproducible part, for @Apollon77:**

`@iobroker/adapter-core@3.4.3` requires a types-only package at runtime:
`build/cjs/index.js:45` → `var import_types = require("@iobroker/types");` — top-level, and
`import_types` is never used afterwards. `@iobroker/types/index.js` is an empty file.

Blocking resolution of that specifier makes `require('@iobroker/adapter-core')` throw
`MODULE_NOT_FOUND` — I reproduced this for `iobroker.admin` too, which declares the package in
`devDependencies` only. So it isn't gree-hvac specific. It works because npm installs the peer;
it would break with `legacy-peer-deps=true` or `--omit=peer`.

If the require is intentional, `@iobroker/types` should be a real dependency of adapter-core,
not a peer. If not, the import should be type-only so the build erases it. Either way I agree
this belongs in adapter-core, not in individual adapters.

XHunter74:
@mcm1957 I moved @iobroker/types to devDependencies

mcm1957:
Just for tracking purposes:
Issue https://github.com/ioBroker/adapter-core/issues/679 create

PR #6634, opened 2026-09-10
Add alias-workbench to LATEST

RicardoHipp:
Add alias-workbench to latest.

- Adapter: https://github.com/RicardoHipp/ioBroker.alias-workbench
- npm: https://www.npmjs.com/package/iobroker.alias-workbench (0.9.8, published with provenance)
- Type: general
- License: MIT
- Adapter checker: 0 errors

The adapter provides an editor for ioBroker aliases: bulk creation from
templates, a device tree browser, MQTT/Tasmota device knowledge and
round-trip checking of read/write formulas.

Note: "Add to latest" on iobroker.dev failed repeatedly with
[E0000] cannot access repository / AxiosError 403 — that is the
unauthenticated GitHub rate limit (60/h per IP) hitting the portal
server, not a problem with the repository. Hence this manual PR.

### DIFF
--- a/sources-dist.json
+++ b/sources-dist.json
@@ -135,6 +135,11 @@
     "icon": "https://raw.githubusercontent.com/iobroker-community-adapters/ioBroker.alias-manager/master/admin/alias-manager.png",
     "type": "general"
   },
+  "alias-workbench": {
+    "meta": "https://raw.githubusercontent.com/RicardoHipp/ioBroker.alias-workbench/main/io-package.json",
+    "icon": "https://raw.githubusercontent.com/RicardoHipp/ioBroker.alias-workbench/main/admin/alias-workbench.png",
+    "type": "general"
+  },
   "alpha-ess": {
     "meta": "https://raw.githubusercontent.com/Gaspode69/ioBroker.alpha-ess/main/io-package.json",
     "icon": "https://raw.githubusercontent.com/Gaspode69/ioBroker.alpha-ess/main/admin/alpha-ess.png",


sources-dist.json
```diff
@@ -135,6 +135,11 @@
     "icon": "https://raw.githubusercontent.com/iobroker-community-adapters/ioBroker.alias-manager/master/admin/alias-manager.png",
     "type": "general"
   },
+  "alias-workbench": {
+    "meta": "https://raw.githubusercontent.com/RicardoHipp/ioBroker.alias-workbench/main/io-package.json",
+    "icon": "https://raw.githubusercontent.com/RicardoHipp/ioBroker.alias-workbench/main/admin/alias-workbench.png",
```
mcm1957: Note: Template uses GERMAN keys

RicardoHipp:
**Object dump as requested**

`alias-workbench` creates no objects of its own. It runs with
`mode: none` — there is no adapter process at all, the whole adapter is an
admin tab. Nothing is ever written to `alias-workbench.0.*`; in the object
browser there is no `alias-workbench.0` folder to select, only
`alias-workbench.admin`.

Following OBJECTDUMP.md, the dump of the instance is therefore empty
(`{}`), attached below.

*Correction:* an earlier version of this comment attached the instance
object `system.adapter.alias-workbench.0` instead. That is the wrong object
for this check and caused E3006 — the keys `instanceObjects`, `objects`,
`protectedNative`, `encryptedNative` and `notifications` are written by
js-controller into every instance object (all empty here).

Everything the adapter persists lives in that instance object:

- `native.ikonRaeume` / `ikonFunktionen` / `ikonErsetzen` / `vorlagenVomAdmin`
  — the four settings from the config page
- `native.vorlagen` — user-defined device templates: detection criteria and
  a list of states with role, type, unit, read/write source and conversion
  formulas, e.g.

```json
{"name": "SET", "rolle": "switch", "typ": "boolean",
 "lesen": "tele.STATE", "schreiben": "cmnd.POWER",
 "leseformel": "JSON.parse(val)?.POWER === 'ON'",
 "lesenSonst": {"punkt": "stat.POWER"}}
```
[alias-workbench.0.json](https://github.com/user-attachments/files/32411580/alias-workbench.0.json)

RicardoHipp:
@iobroker-bot recheck

mcm1957:
reminder 25.9.2026

mcm1957:
reminder 25.9.2026

RicardoHipp:
Released in 0.9.12 (now on npm): 

Two design decisions a reviewer is likely to ask about, answered up front:

**1. Why an admin tab (`adminTab`) and no adapter process (`mode: none`)**

The workbench is an interactive editor, not a configuration dialog: an object
tree on the left, the alias draft with live values on the right, a dry run
before anything is written, a source swap, a template sheet. None of that fits
into a jsonConfig page. The instance's actual settings do use a regular
jsonConfig page (`adminUI.config: "json"`, four switches, `"i18n": true`, all
eleven languages in `admin/i18n/`). Without a process the adapter uses no
memory while nobody has the tab open.
README → *Why* → *Why an admin tab, not a settings page*.

**2. Why four shipped templates use the deprecated role `value.power.consumption`**

The row in question is the energy meter reading of a socket or light, in kWh.
The roles list is right that `value.power.consumption` is the wrong name for
it: `value.power.*` means power (W), while a meter reading is energy (Wh/kWh).
The correct role would be `value.energy.consumed`.

The workbench does not decide on device types itself; it uses the official
`@iobroker/type-detector`. In its current version (6.0.x, and in the current
source) the detector recognises the `CONSUMPTION` slot only with exactly the
old role (`typePatterns.js`, pattern `consumption`,
`role: /^value\.power\.consumption$/`). With `value.energy.consumed` the
detector no longer counts the point as part of the device: a socket comes out
as `socket: SET, ELECTRIC_POWER` plus a separate `info` device, instead of
`socket: SET, ELECTRIC_POWER, CONSUMPTION`.

| Role on the meter reading | Roles list | type-detector |
|---|---|---|
| `value.power.consumption` (used now) | deprecated | socket with consumption |
| `value.energy.consumed` (correct) | correct | socket without consumption, reading as a loose point |

To my knowledge the type-detector is also used by other adapters, e.g.
devices, iot and matter. Aliases created with the correct role would
therefore cause problems there: the consumption would no longer belong to the
device.

Technically I could work around this in the adapter. I think that would be the
wrong way: the roles list and the type-detector contradict each other, and that
should be fixed in the type-detector, not patched in every adapter that relies
on it. As soon as the detector accepts the current role, the templates will
follow.

RicardoHipp:
„Object dump corrected, check is green now – ‘must be fixed’ can be removed

PR #6692, opened 2026-09-18
Update leapmotor to 0.6.7

backfisch88:
Please update my adapter ioBroker.leapmotor to version 0.6.7.


backfisch88:
This is an expedited stable request triggered by a Leapmotor cloud API change, not a new feature request.

Leapmotor appears to have made a backend/API change yesterday (2026-09-17). Since then, the cloud has started returning an error we hadn't seen before: "TOKEN_NOT_AVAILABLE" on session expiry (previously the wording differed). Our automatic re-login logic checks for the substring "Token", but that check is case-sensitive and doesn't match the new all-caps "TOKEN_NOT_AVAILABLE" message, so the auto-relogin safety net silently never triggers. Once a user's session token expires, every subsequent poll (status, consumption, schedules, etc.) fails until a manual adapter restart - no automatic recovery.

This case-sensitivity bug itself has technically been in the code since 0.6.2 (confirmed via git history), but it only became user-impacting once the cloud started using this specific error wording starting yesterday - before that, the previously-used wording apparently did match. So this is effectively a regression triggered by an external API change to an already-stable, core feature (session handling), not a pre-existing dormant issue we're only now getting around to fixing.

0.6.7 also includes two related trip-detection fixes (a trip no longer splits on a brief stop at a light, and a retry-queue bug that left trip energy data stuck on "not yet available" forever) - lower severity, bundled since they were fixed and tested in the same session.

Given any user's adapter can now get silently stuck indefinitely with no automatic recovery due to this upstream change, I'd like to request this be reviewed ahead of the normal 1-2 week latest-soak period. Happy to answer any questions.

